PR #296, opened 2021-12-16
Add check that non-material filled cells are fissile for adding tallies

### LINKED ISSUE #74: Add warning for non-material cells without fissile material
For `OpenMCCellAverageProblem`, we currently check that any cell filled with a material is fissionable before trying to add tallies to it. For cells filled with universes or lattices, we can extend this check to check that the cell contains fissile material by recursing down through all the various fills.

aprilnovak:
With the recent addition of @pshriwise's to OpenMC that lets us get all the material cells within a parent cell, we can now actually see whether any of those contained material cells contain fissile material, for the sake of printing warnings.

Closes #74

### DIFF
--- a/include/base/OpenMCCellAverageProblem.h
+++ b/include/base/OpenMCCellAverageProblem.h
@@ -769,6 +769,9 @@ protected:
    */
   std::vector<coupling::CouplingFields> _elem_phase {};
 
+  /// Maximum number of contained cells in the coupling
+  int _max_n_contained;
+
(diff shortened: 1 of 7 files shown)


aprilnovak:
Actually, this doesn't work with the current logic - there's an interdependence between `cacheContainedCells` and `storeTallyCells` that can't be resolved without calling `cacheContainedCells` in full-"verbosity" mode where all contained cells are explicitly identified, which is super expensive.

PR #312, opened 2022-01-11
Add option to import OpenMC properties

pshriwise:
This adds the ability to import properties before starting OpenMC transport in Cardinal to reproduce an OpenMC transport step in a coupled simulation if needed.


pshriwise:
Looks like there are some conflicts after the merge of #310. But other than that this looks good to me!

PR #319, opened 2022-01-25
Initiate NekRSSeparateDomainProblem

ahuxford:
First iteration of separate-domain coupling between NekRS and a 1D thermal hydraulic code such as SAM, THM, or RELAP-7.


src/base/NekRSSeparateDomainProblem.C
```diff
@@ -0,0 +1,343 @@
...
+    _NekRS_inlet_boundary(isParamValid("NekRS_inlet_boundary") ? &getParam<std::vector<int>>("NekRS_inlet_boundary") : nullptr)
+{
+
+  if (!_toNekRS && !_fromNekRS)
+    mooseError("This problem type needs atleast one of toNekRS_interface \n", 
+    "or fromNekRS_interface to be set to true\n");
+
+  if (_fromNekRS)
+    {
+    if (_fromNekRS_boundary->size() != 1)
```
aprilnovak, line 78: Hypothetical situation here - what if I have just one _physical_ inlet in my NekRS model, but for some reason it's split over two sidesets? As in, I have a pipe and the upper hemisphere entrance is boundary 1 and the lower hemisphere entrance is boundary 2. I think your algorithm should still work, correct? 

Is this error message here protecting against some fundamental issue, or just that we might need to adjust a function somewhere that takes a vector instead of a single value?

src/actions/NekInitAction.C
```diff
@@ -102,7 +102,7 @@ NekInitAction::act()
   }
 
   // setup actions only needed if coupling with MOOSE
-  if (_type == "NekRSProblem")
+  if (_type == "NekRSProblem" || _type == "NekRSSeparateDomainProblem")
```
aprilnovak, line 105: What should happen if the NekRS case doesn't have temperature at all? As in, it only solves for the flow? Is that an error, or do you want to allow it?

src/base/NekInterface.C
```diff
@@ -553,6 +553,56 @@ void flux(const int elem_id, const int order, double * flux_face)
   }
 }
 
+void velocity(const int elem_id, const int order, const double* velocity_SAM)
```
aprilnovak, line 556: The `velocity` and `temperature` functions can be a bit simpler here, since we're not doing any interpolation - you can remove `order` as an input parameter and delete a few unused local variables in the functions themselves (`end_1d` for instance).

include/base/NekInterface.h
```diff
@@ -301,6 +301,22 @@ void volumeSolution(const int order, const bool needs_interpolation, const field
  */
  void flux(const int elem_id, const int order, double * flux_face);
 
+/**
+ * Send velocity from SAM to the nekRS mesh
+ * @param[in] elem_id global element ID
+ * @param[in] order enumeration of the surface mesh order (0 = first, 1 = second, etc.)
+ * @param[in] velocity_SAM is SAM boundary velocity
+ */
+ void velocity(const int elem_id, const int order, const double* velocity_SAM);
```
aprilnovak, line 310: If the temperature/velocity we're sending in are just a single double value, we don't need to use a pointer here. You can change this to `const double velocity`). Let's also remove the explicit reference to SAM here. There might be a few other places I missed as well.

src/base/NekRSSeparateDomainProblem.C
```diff
@@ -0,0 +1,343 @@
...
+
+InputParameters
+NekRSSeparateDomainProblem::validParams()
+{
+  InputParameters params = NekRSProblemBase::validParams();
+  params.addParam<bool>("minimize_transfers_in", false, "Whether to only synchronize nekRS "
+    "for the direction TO_EXTERNAL_APP on multiapp synchronization steps");
+  params.addParam<bool>("minimize_transfers_out", false, "Whether to only synchronize nekRS "
+    "for the direction FROM_EXTERNAL_APP on multiapp synchronization steps");
+  params.addParam<bool>("moving_mesh", false, "Moving mesh");
```
aprilnovak, line 43: If setting `moving_mesh = true` will never do anything in this class, we can just remove it as an input parameter altogether.

src/base/NekRSSeparateDomainProblem.C
```diff
@@ -0,0 +1,343 @@
...
+    _toNekRS_temperature(getParam<bool>("toNekRS_temperature")),
+    _fromNekRS(getParam<bool>("fromNekRS_interface")),
+    _fromNekRS_temperature(getParam<bool>("fromNekRS_temperature")),
+    _ExternalApp_filename(getParam<std::vector<FileName>>("ExternalApp_filename")),
+    _ExternalApp_type(getParam<MooseEnum>("ExternalApp_type")),
+    _fromNekRS_boundary(isParamValid("fromNekRS_boundary") ? &getParam<std::vector<int>>("fromNekRS_boundary") : nullptr),
+    _NekRS_inlet_boundary(isParamValid("NekRS_inlet_boundary") ? &getParam<std::vector<int>>("NekRS_inlet_boundary") : nullptr)
+{
+
+  if (!_toNekRS && !_fromNekRS)
```
aprilnovak, line 72: Good error check! It's minor, but something I like to do when I write error messages is to put the names of input parameters in single quotes so that they pop out a bit more to the user. i.e. write `'toNekRS_interface'` instead of `toNekRS_interface`. Then it's really clear it's an input file parameter.

src/base/NekRSSeparateDomainProblem.C
```diff
@@ -0,0 +1,343 @@
...
+    "for the direction FROM_EXTERNAL_APP on multiapp synchronization steps");
+  params.addParam<bool>("moving_mesh", false, "Moving mesh");
+  params.addParam<bool>("toNekRS_interface", false, "External app -> NekRS interface present?");
+  params.addParam<bool>("toNekRS_temperature", false, "External app -> NekRS temperature transfer?");
+  params.addParam<bool>("fromNekRS_interface", false, "NekRS -> external app interface present?");
+  params.addParam<bool>("fromNekRS_temperature", false, "NekRS -> external app temperature transfer?");
+  params.addRequiredParam<std::vector<FileName>>("ExternalApp_filename", "External app input file name");
+  MooseEnum app_type("SamApp THMApp");
+  params.addRequiredParam<MooseEnum>("ExternalApp_type", app_type, "External app type");
+  params.addParam<std::vector<int>>("fromNekRS_boundary", "Boundary ID through which nekRS will be coupled to external app");
```
aprilnovak, line 51: What if the user enters a boundary ID that doesn't exist? Check out the `nekrs::mesh::validBoundaryIDs` function, and we can add an error check that the boundaries here are valid.

src/base/NekRSSeparateDomainProblem.C
ahuxford: `_fromNekRS_boundary` is only used to calculate a global pressure drop over the NekRS component, to use as a boundary condition in the system code (SAM, etc.). So the variable is only used in creating an average pressure postProcessor on the inlet boundary specified. The error ensures that there's only one inlet boundary used in the postprocessor. But your hypothetical situation would throw that error even if we didn't want it to.

Maybe we should actually keep this within the user's cardinal input file instead of automating the PostProcessor creation. It'll just require the user to generate a "pressure drop PostProcessor" by themselves for the nekRS component. That might be best, similar to removing all the automated MultiApp and transfers we have right now

src/actions/NekInitAction.C
ahuxford: I want to allow that because I want temperature to be an optional coupling, where as flow is always coupled. If we do have temperature coupling but nekrs doesn't have temperature, than we do want an error. So I will add that error for sure.

include/base/NekInterface.h
ahuxford: My fault I'll doublecheck everywhere for SAM

src/actions/NekInitAction.C
aprilnovak: The reason I ask is because within this if-statement at line 105, Cardinal is going to throw an error if the NekRS problem _doesn't_ have temperature (because we need temperature to exist for `NekRSProblem`). So if we're okay with temperature not existing for `NekRSSeparateDomainProblem`, we should split up this if statement into two parts (one that only applies for `NekRSProblem` that errors if temperautre doesn't exist, and a second that initializes scratch for us - which we want for both problems)

src/base/NekRSSeparateDomainProblem.C
ahuxford: There's a few things I still left in the updated version of this file because I wasn't sure how they're used, like `minimize_transfers_in`. But along these lines, I tried removing `moving_mesh` from the problem type but I couldn't compile the problem without it. To be honest I didn't check into the reason behind this very much.

src/base/NekInterface.C
ahuxford: These functions are now located in `src/base/NekRSSeparateDomain.C` in the updated files. But I'm not sure which variables there aren't necessary for example specifically in `velocity` or in `sendBoundaryVelocityToNek`.

PR #330, opened 2022-02-11
Adding NEKRS_HOME to submit script for bebop

pshriwise:
Noticed this was missing in the firs submit script example for bebop.


pshriwise:
Oops, nevermind. I see that example was for running OpenMC only. Doesn't require this variable to be defined.

PR #334, opened 2022-02-22
Adds error for moving mesh problem when CreateDisplacedProblemAction is not triggered

anshchaube:
To properly simulate moving mesh problems, Moose requires the NekRS subapp to have
```
displacements = '<string with three displacement variables>'
```
in the `[Mesh]` block to trigger `CreateDisplacedProblemAction`. This PR adds the error handling code necessary to ensure moving mesh problems are run with the conditions necessary to initialise objects associated with said action (e.g. `_displaced_problem`, `_displaced_mesh` etc.).
EDIT (2022-03-02): specifying `use_displaced_mesh=true` seems unnecessary as of today.


src/base/NekRSProblem.C
```diff
@@ -63,6 +63,10 @@ NekRSProblem::NekRSProblem(const InputParameters &params) : NekRSProblemBase(par
 
     if (!_nek_mesh->getMesh().is_replicated())
       mooseError("Distributed mesh features are not yet implemented for moving mesh cases!");
+
+    if (!(_nek_mesh->has_displacements() && _nek_mesh->use_displaced_mesh()))
+      mooseError("Moving mesh problems require displacements and "
```
aprilnovak, line 68: It would be good to have a test that covers this to make sure we have correct usage

anshchaube: Thanks! Tests were added in the latest commit `c1d9c6a`.

test/tests/nek_errors/deformation/nek2.i
```diff
@@ -0,0 +1,105 @@
...
+  [TimeStepper]
+    type = NekTimeStepper
+  []
+  [Quadrature]
+    type = GAUSS_LOBATTO
+    order = SECOND
+  []
+[]
+
+[Postprocessors]
```
aprilnovak, line 51: For these two tests, let's delete all the postprocessors, source terms, and other extraneous stuff in these input files. If we need to change any syntax in the future, reducing things not actually used by the test will help.

anshchaube: Thanks, I think I have removed everything. Tests are still working fine.

test/tests/nek_errors/deformation/box-test1.i
```diff
@@ -0,0 +1,149 @@
...
+  use_displaced_mesh = true
+  displacements = 'disp_x_o disp_y_o disp_z_o'
+[]
+
+[Mesh]
+  type = FileMesh
+  file = box.msh
+  parallel_type = replicated
+  order = SECOND
+  displacements = 'disp_x_o disp_y_o disp_z_o'
```
aprilnovak, line 11: So, from my understanding, everything put into `[GlobalParams]` essentially gets copied into all the other blocks. So I don't think we need the separate `[GlobalParams]` block?

anshchaube: I'll test it and remove it from the `[Mesh]` block. I think you asked a question about this in the MOOSE slack a few months ago, and someone responded that `displacements` and `use_displaced_mesh` in `[GlobalParams]` don't always get applied to other blocks. Is that still true? If so, maybe we should avoid `[GlobalParams]` altogether from now on.

aprilnovak: Ohhh I totally forgot about that. I think you're right - if we _can_ avoid `[GlobalParams]`, then let's do that. Can you test it out and see what the behavior is to remind us?

test/tests/nek_errors/deformation/box-test1.i
```diff
@@ -50,6 +45,7 @@
   [heat]
     type = HeatConduction
     variable = temp
+    use_displaced_mesh = true
```
aprilnovak, line 48: Ooh, if it's going to be this much repetition then I will backpedal and say let's go with the `[GlobalParams]` - I thought we'd just need `use_displaced_mesh = true` in the `[Mesh]` block. 

Sorry for the whiplash!

test/tests/nek_errors/deformation/box-test1.i
anshchaube: I think it works either way for this case, but I have had experiences in the past where it triggers MOOSE errors when using the Tensor Mechanics module.

test/tests/nek_errors/deformation/box-test1.i
anshchaube: Yeah, it gets a bit tedious. This will take some time to test (turning it off for each of these objects, seeing where behaviour diverges from expectations). So far we've just been using GlobalParams at the top as a redundant safety net, but it does create issues when used with the Tensor Mechanics module and its dependencies.

No worries, I'll undo the change.

PR #337, opened 2022-02-26
Allow initial source to be taken from previous Picard iteration. 

### LINKED ISSUE #336: Allow re-use of the fission source for Picard iterations
For all but the first Picard iteration, it would improve convergence of the initial fission source if we re-use the converged source from the previous iteration (then, you aren't starting "from scratch," but instead from converged source from the previous iteration, which presumably is a better approximation of the true fission source than whatever initial source you set before doing any inactive batches).

aprilnovak:
Closes #336

### DIFF
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -75,7 +75,7 @@ public:
   std::string printMaterial(const int32_t & index) const;
 
   /**
-   * Write the source bank to HDF5 for postprocessing or using in subsequent solves
+   * Write the source bank to HDF5 for postprocessing or for use in subsequent solves
    * @param[in] filename file name
    */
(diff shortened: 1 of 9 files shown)


include/base/OpenMCProblemBase.h
```diff
@@ -74,6 +74,18 @@ class OpenMCProblemBase : public ExternalProblem
    */
   std::string printMaterial(const int32_t & index) const;
 
+  /**
+   * Write the source bank to HDF5 for postprocessing or using in subsequent solves
```
pshriwise, line 78:
```suggestion
   * Write the source bank to HDF5 for postprocessing or for use in subsequent solves
```

pshriwise (approved):
Super small suggestion but looks great otherwise! Nice feature @aprilnovak.

PR #354, opened 2022-04-02
Fix error in dP calculation, initialize coupling_type booleans

ahuxford:
Quick fixes to NekRSSeparateDomainProblem

### DIFF
--- a/src/base/NekRSSeparateDomainProblem.C
+++ b/src/base/NekRSSeparateDomainProblem.C
@@ -52,6 +52,11 @@ NekRSSeparateDomainProblem::NekRSSeparateDomainProblem(const InputParameters & p
     _outlet_boundary(getParam<std::vector<int>>("outlet_boundary")),
     _inlet_boundary(getParam<std::vector<int>>("inlet_boundary"))
 {
+
+  // check coupling_type provided
+  _inlet_coupling = false;
+  _outlet_coupling = false;
+
   for (const auto & c : _coupling_type)
   {
     if (c == "inlet")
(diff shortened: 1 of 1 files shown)


src/base/NekRSSeparateDomainProblem.C
```diff
@@ -52,6 +52,11 @@ NekRSSeparateDomainProblem::NekRSSeparateDomainProblem(const InputParameters & p
     _outlet_boundary(getParam<std::vector<int>>("outlet_boundary")),
     _inlet_boundary(getParam<std::vector<int>>("inlet_boundary"))
 {
+
+  // check coupling_type provided
+  _inlet_coupling = false;
```
aprilnovak, line 57: Are you going to be using these variables in a future PR? If not, then I'd suggest removing them so that the code is a bit clearer.

ahuxford: I was making sure we had input models for the 3 coupling options (NekRS b/w two 1-D domains, NekRS at exit of 1-D domain, NekRS at inlet of 1-D domain), and I found that for  **NekRS at inlet of 1-D domain**, the coupling was hanging unless I initialized these booleans. These booleans are overwritten in the for loop written in lines 60-66.

aprilnovak: Oh I see, thanks for explaining. For future reference, typical C++ coding style is to set any "unset values" for variables in the member initialization list of a class, like this:

```
Class::Class():
  _variable(0.0),
  _thing(0.0)
{
  if (condition)
  {
    // set non-zero/non-false values for stuff
    _thing = 4.0;
 }
}
```

But what you have is also fine here, since we're not enforcing style very tightly.

PR #357, opened 2022-04-06
Initial documentation for NekRSSeparateDomainProblem

ahuxford:
I went with a layout that made sense to me and have figures and examples to show the usage. I plan to add example inputs and more error checking within the next week so that the documentation makes more sense for users.


ahuxford:
Also I wasn't sure about naming convection for images in `cardinal/doc/content/media`. So I appended the images with the problem name.

aprilnovak:
Looks very nice! Great work! I'll go ahead and merge after the tests pass

aprilnovak:
By the way, you can name the figures in `media` whatever name you like - when you include a figure in documentation, it will just look for a figure with the same name.

PR #363, opened 2022-04-13
Reduce parallel test requirements.

aprilnovak:
These tests don't need so many resources, so we can let them run on more systems by reducing these requirements.

### DIFF
--- a/tutorials/pebbles/tests
+++ b/tutorials/pebbles/tests
@@ -3,7 +3,7 @@
     type = RunApp
     input = solid.i
     cli_args = "openmc:Problem/particles=1000 openmc:Problem/batches=10 openmc:Problem/inactive_batches=5"
-    min_parallel = 8
+    min_parallel = 4
     requirement = "The system shall be able to run the pebble tutorial."
     required_objects = 'OpenMCCellAverageProblem'
   []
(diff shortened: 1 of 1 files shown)


lewisgross1296:
One thing I noticed about the failed tests I was getting (that you changed the required `min_parallel` from 8 to 4) was that I supposedly have 8 cores according to `lscpu`, but the output of my failed test said otherwise:
```
tutorials/pebbles.solid_um: There are not enough slots available in the system to satisfy the 8
tutorials/pebbles.solid_um: slots that were requested by the application
```
It gave the suggestions:
```
tutorials/pebbles.solid_um: In all the above cases, if you want Open MPI to default to the number
tutorials/pebbles.solid_um: of hardware threads instead of the number of processor cores, use the
tutorials/pebbles.solid_um: --use-hwthread-cpus option.
tutorials/pebbles.solid_um: 
tutorials/pebbles.solid_um: Alternatively, you can use the --oversubscribe option to ignore the
tutorials/pebbles.solid_um: number of available slots when deciding the number of processes to
tutorials/pebbles.solid_um: launch.
```
I can look into trying these by modifying the test scripts locally in my Docker image. I think for performance reasons, oversubscribing is probably not the way, but I'd be curious to see what happens with `--use-hwthread-cpus`. I don't think this is a problem caused by Docker, but I need to verify information about if I have a Hostfile in my image and what lives there.

PR #364, opened 2022-04-14
Add scalar transfers for NekRSSeparateDomainProblem

ahuxford:
This PR is dependent on [PR 358 here](https://github.com/neams-th-coe/cardinal/pull/358), but I'd like to get some early feedback if possible. 

The changes allow `NekRSSeparateDomainProblem` to couple scalars between NekRS and a 1D T/H code, for up to a maximum of 5 total coupled fields now (velocity, temperature, scalar01, scalar02 and scalar03).

There aren't any error checks in this PR, but I'm working on a PR for adding error checking into `NekRSSeparateDomainProblem`. That will come soon.


src/base/NekRSSeparateDomainProblem.C
```diff
@@ -224,6 +253,40 @@ NekRSSeparateDomainProblem::sendBoundaryTemperatureToNek()
...
+  }
+
+  solution.localize(*_serialized_solution);
+
+  auto & mesh = _nek_mesh->getMesh();
+
+  // check which scalar PP to pass to NekRS
+  const PostprocessorValue * scalarValue = nullptr;
+
+  if(scalarId==1)
```
aprilnovak, line 275: With code like this, we run the risk of future updates that add scalars, but we forget to extend this part. What I'd do instead here is change the `sendBoundaryScalarToNek` function to take the `scalarValue` as an input, then call 

```
sendBoundaryScalarToNek(1, _toNekRS_scalar01);
```

src/base/NekRSSeparateDomainProblem.C
```diff
@@ -336,4 +432,33 @@ NekRSSeparateDomainProblem::temperature(const int elem_id, const double temperat
...
+  const auto & bc = _nek_mesh->boundaryCoupling();
+
+  // We can only write into the nekRS scratch space if that face is "owned" by the current process
+  if (nekrs::commRank() == bc.processor_id(elem_id))
+  {
+    nrs_t * nrs = (nrs_t *)nekrs::nrsPtr();
+    mesh_t * mesh = nekrs::entireMesh();
+
+    int scalarFieldOffset = nekrs::scalarFieldOffset();
+    int scalarWrkOffset = (scalarId + 1)*scalarFieldOffset; // offset by 1, first "scalar" is temperature
```
aprilnovak, line 447: I think this should be indexed like `scalarId * scalarFieldOffset`, right? Because we now have `scalarID >= 1`.

test/tests/nek_errors/invalid_field/nek_uo.i
```diff
@@ -19,10 +19,10 @@
 [UserObjects]
   [subchannel_binning]
     type = HexagonalSubchannelBin
-    bundle_pitch = 0.02583914354890463
```
aprilnovak, line 22: Why were these changed? We should avoid changing any test specifications unless we absolutely need to.

ahuxford: Now that the temperature check was moved to within `solutionPointer`, running that test with the original `HexagonalSubchannelBins` inputs throws an error saying the bins set are not valid, which is an error reached before getting to `solutionPointer` to throw the temperature error. 

Since this test is for checking an invalid temperature field, I changed the `HexagonalSubchannelBins` inputs to these new values so that the code can run enough to reach the `solutionPointer` and throw the temperature error we want to see. Otherwise, this test would throw an error related to `HexagonalSubchannelBins`

src/base/NekRSSeparateDomainProblem.C
ahuxford: Yes good point, I'll change this

test/tests/nek_errors/invalid_field/nek_uo.i
aprilnovak: Oh, gotcha - I didn't realize that invalid values were here in the first place. Thanks for the fix!

src/base/NekRSSeparateDomainProblem.C
ahuxford: My thought process is that for this coupling class, we'd keep the following portions of `usrwrk` always reserved for these following coupled values. 

`usrwrk[id]` reserved for Velocity
`usrwrk[id + scalarFieldOffset]` reserved for Temperature, even if temperature isn't solved for
`usrwrk[id + 2*scalarFieldOffset]` reserved for Scalar01, even if it isn't solved for
`usrwrk[id + 3*scalarFieldOffset]` reserved for Scalar02, even if it isn't solved for
`usrwrk[id + 4*scalarFieldOffset]` reserved for Scalar03, even if it isn't solved for

So `scalarWrkOffset = (scalarId + 1)*scalarFieldOffset` for the scalars because `usrwrk[id + scalarFieldOffset]` is reserved for temperature no matter what.

aprilnovak: I see your point - if in the future we wanted to couple your NekRS domain to a solid heat conduction model, then we'd still need the 0th and 1st slots in `usrwrk` to hold heat flux and temperature, respectively. 

We'll just need to be sure this is well documented so people don't get confused with the indexing.

ahuxford:
I addressed the `sendBoundaryScalarToNek` feedback and added the scalar transfers into the problem's documentation.
This PR got a little messy because I thought I accidentally updated contrib directories. I can make a new clean version if that'd help.

doc/content/source/problems/NekRSSeparateDomainProblem.md
```diff
@@ -1,33 +1,51 @@
...
 
 !alert warning
 This class currently only supports dimensional solutions coming from NekRS. Nondimensional support is in progress.
 
-### Velocity and temperature coupling
+### Velocity, temperature, and scalar coupling
 
-[inlet_outlet] shows the coupling of velocity and temperature for `coupling_type = inlet outlet`, where the 1D T/H code is coupled to NekRS's inlet and outlet boundaries.
+[inlet_outlet] shows the coupling of velocity (V), temperature (T), and scalar01 (S01)
+ for `coupling_type = inlet outlet`, where the 1D T/H code is coupled to NekRS's inlet and outlet boundaries.
```
aprilnovak, line 19: I think technically the syntax needs to have single quotes around the `MultiMooseEnum` if more than one is specified, i.e. `coupling_type = 'inlet outlet'`. I didn't catch this before, but let's update it now (there may be other places in the documentation where this happens too).

aprilnovak, line 19: I think technically the syntax needs to have single quotes around the `MultiMooseEnum` if more than one is specified, i.e. `coupling_type = 'inlet outlet'`. I didn't catch this before, but let's update it now (there may be other places in the documentation where this happens too).

doc/content/source/problems/NekRSSeparateDomainProblem.md
```diff
@@ -37,67 +55,86 @@ Pressure information is transferred between NekRS and the 1D T/H code using a gl
...
 | outlet_T | &nbsp; | $\checkmark$, if temperature solved | $\checkmark$, if temperature solved |
+| inlet_S01 | $\checkmark$, if scalar01 in `coupled_scalars` | &nbsp; | $\checkmark$, if scalar01 in `coupled_scalars` |
+| outlet_S01 | &nbsp; | $\checkmark$, if scalar01 in `coupled_scalars` | $\checkmark$, if scalar01 in `coupled_scalars` |
+
 
 
-Example MultiApp transfers are shown below for `coupling_type = 'inlet outlet'` with transfers between the NekRS MainApp and the `1D_TH_code` SubApp. These transfers include all possible postProcessors created.
+Example MultiApp transfers for velocity, temperature and scalar01 coupling are shown below for
+ `coupling_type = 'inlet outlet'` and `coupled_scalars = 'scalar01'`. These transfers assume NekRS
+ is acting as the MainApp and the `1D_TH_code` as the SubApp.
```
aprilnovak, line 77: I would add another sentence making it clear that Nek doesn't *need* to be the main app, but just that the documentation shows this example.

aprilnovak, line 77: I would add another sentence making it clear that Nek doesn't *need* to be the main app, but just that the documentation shows this example.

src/base/NekRSSeparateDomainProblem.C
```diff
@@ -297,34 +269,15 @@ NekRSSeparateDomainProblem::sendBoundaryScalar02ToNek()
...
-  {
-    _serialized_solution->init(_aux->sys().n_dofs(), false, SERIAL);
-    _first = false;
-  }
-
-  solution.localize(*_serialized_solution);
-
-  auto & mesh = _nek_mesh->getMesh();
+//  // check which scalar PP to pass to NekRS
+//  const PostprocessorValue * scalarValue = nullptr;
```
aprilnovak, line 273: Does this commented-out code need to be here?

aprilnovak, line 273: Does this commented-out code need to be here?

aprilnovak:
I like the combination of the various functions for scalars into a single shared function! These are all my comments, thanks for the contribution!

aprilnovak:
I like the combination of the various functions for scalars into a single shared function! These are all my comments, thanks for the contribution!

ahuxford:
Thanks I'll address this comments asap

src/base/NekRSSeparateDomainProblem.C
ahuxford: Thanks for pointing this out, I missed it.

PR #387, opened 2022-05-25
Allow power to be provided by a postprocessor

### LINKED ISSUE #384: Change `power` to a postprocessor
The power for `OpenMCCellAverageProblem` right now can only take a `Real` value, but for transients we may want to change the power as a function of time. If we make `power` a `Postprocessor`, we can retain the current behavior while also allowing another application to control the power.

fyi @luizaldeia

aprilnovak:
For time-dependent simulations, the power used to normalize the OpenMC tallies should vary with time. This PR allows a postprocessor to be provided for `power`. No syntax changes are needed if you still want to set power with a Real value, like `power = 100.0`. 

Closes #384

### DIFF
--- a/test/tests/neutronics/heat_source/from_postprocessor.i
+++ b/test/tests/neutronics/heat_source/from_postprocessor.i
@@ -1,0 +1,96 @@
+[Mesh]
+  [sphere]
+    type = FileMeshGenerator
+    file = ../meshes/sphere.e
+  []
+  [solid]
+    type = CombinerGenerator
+    inputs = sphere
+    positions = '0 0 0
+                 0 0 4
+                 0 0 8'
+  []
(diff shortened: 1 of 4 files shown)


test/tests/neutronics/heat_source/from_postprocessor.i
```diff
@@ -0,0 +1,96 @@
...
+[Postprocessors]
+  [p]
+    type = Receiver
+    default = 100.0
+  []
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  power = p
```
aprilnovak, line 57: @luizaldeia, once this is merged you will be able to provide a postprocessor name (in this example, this is `p`) for power.

PR #388, opened 2022-05-25
Allow start time to be specified on Executioner.

aprilnovak:
This PR now respects whatever choice is set on `Executioner/start_time` for the NekRS wrapping. Previously, we would only ever respect `Executioner/start_time = 0.0`, which would set the NekRS start time to 0.0 even if it's beginning from a restart file at a different time value.

I don't think there is a clean path to *setting* `Executioner/start_time` based on what time value is actually in a NekRS restart file, though. Once MOOSE reads the `Executioner/start_time` from the input file, it uses that value in a bunch of other objects and if we want to change it after-the-fact, we'd need to re-call a bunch of things (like, initializing time steppers, etc.). This might be possible but probably would miss something.


ahuxford (approved):
This looks good to me.

aprilnovak:
Thanks for the speedy review!

PR #389, opened 2022-05-27
Do not make a directory in LIBMESH_DIR for openmc compile

### LINKED ISSUE #390: Do not make a `$LIBMESH_DIR/include/contrib` directory
This `mkdir` is unnecessary:

https://github.com/neams-th-coe/cardinal/blob/b31a3b0418a640df0afae9ffceabbff1b0b2be4d/config/openmc.mk#L3

loganharbour:
Closes #390


aprilnovak:
Nice! Always in favor of when things can be simplified.

PR #399, opened 2022-06-15
start.md cleanup

lewisgross1296:
removed $ from beginning of commands to be copied, so user can copy/paste without having to delete the $

### DIFF
--- a/doc/content/start.md
+++ b/doc/content/start.md
@@ -14,8 +14,8 @@ To get access to Cardinal, simply clone the repository and
 are located in the root directory of the Cardinal repository).
 
 ```
-$ git clone https://github.com/neams-th-coe/cardinal.git
-$ cd cardinal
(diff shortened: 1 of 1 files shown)


lewisgross1296:
@aprilnovak this is ready for review

lewisgross1296:
Another thought just came to me as I was attempting to `run_tests` on Ulam. The machine has both python2 and python3 and was getting confused. Even when I followed the instructions and added `export PYTHONPATH=/home/lgross/cardinal/contrib/moost/python` to my `.bashrc`, `./run_tests -j8` failed. Maybe we should explicitly tell users to run the tests with the command `python3 run_tests -j8` since python2 doesn't work anyways, as opposed to have the instructions be`./run_tests -j8`. I think the user would just experience a "no command: python3" if they only have python2 installed on their machine, which should be a good enough clue to use a higher version, imo.

aprilnovak (approved):

PR #400, opened 2022-06-15
Spell out verification steps

aprilnovak:
Inspired by discussion with @lewisgross1296, this PR adds a few steps the user can try in order to verify that all the basic parts of Cardinal have been installed correctly.

Also moved the common test suite errors to a separate page to try and slim down the getting started page to just the essentials.


lewisgross1296:
Maybe would be good to just eliminate the extra `$` here and close PR#399?

PR #405, opened 2022-06-24
Control allocation of  NekRS problem scratch spaces

aprilnovak:
As Cardinal cases get more complicated, it is helpful to print for the user a table showing exactly what Cardinal is writing into each slot of `nrs->usrwrk`. This PR adds a table print, and also provides user control over the total number of scratch spaces to allocate. For the vast majority of `NekRSProblem` and `NekRSSeparateDomainProblem` cases, you only need 1-3 scratch spaces reserved for Cardinal, so we can cut down on memory usage by allowing the user to use less. 

New documentation for `NekRSProblem` is here: https://cardinal.cels.anl.gov/source/problems/NekRSProblem.html (see `Transfers to NekRS` section).

New documentation for `NekRSSeparateDomainProblem` is here: https://cardinal.cels.anl.gov/source/problems/NekRSSeparateDomainProblem.html (see `Using transferred data...` section)


doc/content/source/problems/NekRSSeparateDomainProblem.md
```diff
@@ -128,56 +130,114 @@ Example MultiApp transfers for velocity, temperature and scalar01 coupling are s
...
+
+For example, if your case couples NekRS to MOOSE with `coupling_type = inlet`
+and `coupled_scalars = scalar02`, but has a NekRS case without a temperature solve,
+the slices normally dedicated to storing inlet temperature and scalar01 are still
+allocated (because we keep the order the same in `nrs->usrwrk`), but just won't hold
+any meaningful information.
+A table similar to the following would print out at the start of your simulation.
+You could use slices 4 onwards for custom purposes.
+
+!listing id=l11 caption=Table printed at start of Cardinal simulation that describes available scratch space for a case that couples NekRS to MOOSE via an inilet and with the second passive scalar, but without the first passive scalar or a temperature solve. A total of 7 slots are allocated by setting `n_usrwrk_slots` to 7
```
ahuxford, line 167:
```suggestion
!listing id=l11 caption=Table printed at start of Cardinal simulation that describes available scratch space for a case that couples NekRS to MOOSE via an inlet and with the second passive scalar, but without the first passive scalar or a temperature solve. A total of 7 slots are allocated by setting `n_usrwrk_slots` to 7
```
Just a typo in the caption

doc/content/source/problems/NekRSSeparateDomainProblem.md
```diff
@@ -128,56 +130,114 @@ Example MultiApp transfers for velocity, temperature and scalar01 coupling are s
...
+- Slot 2: `inlet_S01` postprocessor value
+- Slot 3: `inlet_S02` postprocessor value
+- Slot 4: `inlet_S03` postprocessor value
 
 This allows the user the freedom to choose what type of profile they want to implement at the inlet boundary, i.e. flat or fully-developed. Below are a few example implementations.
 
 Example for a flat velocity(x) profile:
 
-```
+!listing language=cpp
```
ahuxford, line 206: Thanks for adding the language.

ahuxford:
I think everything is clear.

src/base/NekRSProblem.C
```diff
@@ -62,8 +62,43 @@ NekRSProblem::NekRSProblem(const InputParameters & params)
...
+  //   3: y_displacement   (if _moving_mesh is true)
+  //   4: z_displacement   (if _movign_mesh is true)
+  //
+  // The most we will do is skip allocating terms at the end of this ordering if we
+  // don't need them. We never change the ordering of "earlier" terms.
+  std::vector<std::string> indices =
+    {"flux", "heat_source", "x_displacement", "y_displacement", "z_displacement"};
+
+  // progressively erase terms from the back if we don't need them
+  if (!_moving_mesh)
```
anshchaube, line 81: Do we want to support moving_mesh without CHT? E.g. for tensor mechanics + ALE solver in nek. In that case, this needs to be split up into separate if conditions.

anshchaube, line 81: Do we want to support moving_mesh without CHT? E.g. for tensor mechanics + ALE solver in nek. In that case, this needs to be split up into separate if conditions.

doc/content/source/problems/NekRSProblem.md
```diff
@@ -140,22 +145,77 @@ the last time it was sent to NekRS). A means by which to reduce some of these
...
+The specific data transfers going into NekRS are determined based on how the
 `NekRSMesh` was constructed, i.e. whether boundary and/or volume coupling is used.
 
+Data is "sent" into NekRS by writing into the `nrs->usrwrk` scratch space array,
+which NekRS makes available within the boundary condition functions in the `.oudf` file
+(on device, this array is technically called the `nrs->o_usrwrk` array).
+[usrwrk_nrsp] shows the assignment of "slots" in the `nrs->usrwrk` scratch space
+array with quantities written by Cardinal. Because different quantities are written into
+Cardinal depending on the problem setup, if a particular slice is not needed for a
+case, it will just hold zero values. That is, the **order** of the various quantities
```
anshchaube: Say I want to pass `flux heat source y_displacement`:
1. Won't we be wasting memory zeros allocated for `x_displacement` in our efforts to keep the order constant for developers?
2. As a user, I would expect y_displacement to be stored at offset #3 (i.e. 2), but that could just be me.
3. We already expect the user to keep their `usrwrk` slices and Cardinal's `usrwrk` slices separate. I don't think it's too bad to pass on the responsibility to keep offsets straight to the user, as long as we're clear in the documentation.
4. For the convenience of developers, we could use an enum to make sure that 
  a) "y_displacement" is a valid input
  b) map y_displacement to a valid offset, that is incremented each time a valid input is detected. 
  That way developers have to call on one enum, which is stored/modified in a single place (say `NekInterface.C`)

aprilnovak: I agree, this part might be confusing. I don't think there's a super clear winner on our two options (keeping the order the same + wasting memory, or having the order be problem-dependent + not wasting memory). Consider the case where a user is solving a moving mesh problem with a volumetric heat source (like the cases we've been working on). If the user wants to momentarily turn off the heat source part, then they'll need to modify the indexing in the `.oudf` file, which opens up the door to making mistakes if they forget to do this. 

I think my long-term solution will be to figure out how to get `libUDF` to recognize Cardinal source code. Then we could have a function in `NekInterface.C` like `usrwrk_index(field::nekWriteEnum & field)` (like you have on your local  `meshmvmt` branch) that can get called directly in the `.oudf` file so that the user doesn't ever mess up the indexing. Then, we can absolutely go the route of not wasting any memory at all. What do you think?

anshchaube: I agree. I know I'll make that same mistake very soon, but as long as the documentation is clear about this, it's a good compromise for now, and I think we're right to lean towards wanting to save memory. Hopefully, it won't be too hard for the user to debug - just based on all these terms likely being different orders of magnitude and having separate physical effects that will become apparent relatively easily ("I turned off my heat source, but the temperature is still going up...").

anshchaube: At some point, maybe I should try to create a segfault based around a potential user error in this scenario (reading outside the bounds of currently allocated `usrwrk`), and write a debugging guide based off of that (once we have the HYPRE bug figured out) for the Cardinal documentation.

aprilnovak: Yeah, the only change that this PR makes is to allow the users to "cut off" allocating members in `usrwrk` that they definitely don't need. In the future if we want to exclusively allocate only what the user needs, I'd prefer to do that as a separate PR, since it'll require some additional testing. At the moment, pretty much everyone using `NekRSProblem` is only using the conjugate heat transfer capabilties, and only need a single slot in `nrs->usrwrk` saved for flux. 

When the mesh solver is in, let's definitely revisit this!

PR #418, opened 2022-07-09
Write usrwrk to field files.

### LINKED ISSUE #416: Add helper function to write quantities in `nsr->usrwrk`
When coupling Nek to MOOSE, it can be useful to view the data that gets sent into NekRS via the `nrs->usrwrk` scratch array. We should add a helper utility to Cardinal (perhaps a new type of `Output`) to write field files.

inspired by @anshchaube's work

aprilnovak:
This allows the user to write entries in the `nrs->usrwrk` array to field files via Cardinal. This can be very useful for viewing the data sent from MOOSE into NekRS. The user specifies which slots of the usrwrk array they want to write, and then how they want the files to be named:

```
[Problem]
  type = NekRSProblem # or any other NekRS problem

  usrwrk_output = '0 3' # output the 0th and 3rd slots
  usrwrk_output_prefix = 'flx vel' # will name the files with these prefixes
[]
```

Closes #416


anshchaube (approved):

anshchaube:
Looks good, thanks!

PR #425, opened 2022-07-15
Replace flux interpolation with generic boundary interpolation.

aprilnovak:
This replaces the flux-specific usrwrk writing function with a generic function that applies to any boundary quantity. This will be useful when adding the mesh elasticity solver implementation.


anshchaube:
Looks good, thanks!

PR #426, opened 2022-07-15
Move `_moving_mesh` from NekRSProblem to NekRSMesh

aprilnovak:
It makes more sense for the moving mesh information to be stored on `NekRSMesh`, as opposed to `NekRSProblem`. I cherry-picked a commit from @anshchaube to set up this PR to reduce the amount of code we need to review later when the mesh elasticity solver is merged.

### DIFF
--- a/include/base/NekRSProblem.h
+++ b/include/base/NekRSProblem.h
@@ -106,7 +106,7 @@ public:
    */
   virtual double minInterpolatedTemperature() const;
 
-  virtual bool movingMesh() const override { return _moving_mesh; }
+  virtual const bool hasMovingNekMesh() const override { return _nek_mesh->movingMesh(); }
 
 protected:
   virtual void addTemperatureVariable() override { return; }
@@ -119,9 +119,6 @@ protected:
   void flux(const int elem_id, double * flux_face);
 
   std::unique_ptr<NumericVector<Number>> _serialized_solution;
-
-  /// Whether the problem is a moving mesh problem i.e. with on-the-fly mesh deformation enabled
-  const bool & _moving_mesh;
 
   /// Whether a heat source will be applied to NekRS from MOOSE
   const bool & _has_heat_source;
--- a/include/base/NekRSProblemBase.h
+++ b/include/base/NekRSProblemBase.h
@@ -81,10 +81,10 @@ public:
   virtual bool nondimensional() const { return _nondimensional; }
 
   /**
-   * Whether the mesh is moving
-   * @return whether the mesh is moving
-   */
-  virtual bool movingMesh() const { return false; }
+  * Whether the mesh is moving
+  * @return whether the mesh is moving
+  */
+  virtual const bool hasMovingNekMesh() const { return false; }
 
   /**
    * Whether data should be synchronized in to nekRS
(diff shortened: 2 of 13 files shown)


include/base/NekRSProblem.h
```diff
@@ -106,7 +106,7 @@ class NekRSProblem : public NekRSProblemBase
    */
   virtual double minInterpolatedTemperature() const;
 
-  virtual bool movingMesh() const override { return _moving_mesh; }
+  virtual const bool hasMovingNekMesh() const override { return _nek_mesh->movingMesh(); }
```
aprilnovak, line 109: Is there a reason why this implementation of `hasMovingNekMesh()` couldn't be placed into the `NekRSProblemBase`, and shared by all derived classes?

include/base/NekRSStandaloneProblem.h
```diff
@@ -31,5 +31,5 @@ class NekRSStandaloneProblem : public NekRSProblemBase
 
   static InputParameters validParams();
 
-  virtual bool movingMesh() const override;
+  virtual bool movingMesh() const;
```
aprilnovak, line 34: I think this line can be deleted altogether - there is no longer a need to define `movingMesh()` on a problem-type class.

include/base/NekRSProblem.h
anshchaube: I actually wasn't sure if it needed to be in the base class, but I'm happy to change that. I think the previous version of this in `NekRSProblemBase` just returned `false`??

include/base/NekRSProblemBase.h
```diff
@@ -81,10 +81,10 @@ class NekRSProblemBase : public ExternalProblem
...
 
   /**
-   * Whether the mesh is moving
-   * @return whether the mesh is moving
-   */
-  virtual bool movingMesh() const { return false; }
+  * Whether the mesh is moving
+  * @return whether the mesh is moving
+  */
+  virtual const bool hasMovingNekMesh() const { return false; }
```
anshchaube, line 87: This is what `hasMovingNekMesh()` looks like in the base class right now

include/base/NekRSProblem.h
aprilnovak: Oh, you're right. Ok I think I recall why we had `moving_mesh` be on `NekRSProblem`, since it only made sense for that class. With these changes, we should technically be throwing a warning if someone sets `moving_mesh` on `NekRSMesh` but pairs it with `NekRSStandaloneProblem`/`NekRSSeparateDomainProblem` (because it's unused).

Which do you prefer?

- Keep `moving_mesh` on `NekRSProblem`, so there's no ambiguity about when it is used, but some things related to the mesh aren't located (most logically) on a Cardinal class that deals with meshes
- Move `moving_mesh` to `NekRSMesh`, and add extra warning checks that no one specifies `moving_mesh = true` paired with one of the NekRS-wrapper classes that don't make sense for it to be specified.

anshchaube: I can add the checks to `NekRSMesh`

aprilnovak: Great, thanks! Yes let's add checks to NekRSMesh that throw a warning to the user if they set `moving_mesh` for anything except `NekRSProblem`.

aprilnovak:
Closing because this is addressed in #497

PR #428, opened 2022-07-19
Nek plotting scripts

aprilnovak:
This adds a script to plot the NekRS console output as a function of timestep; this can be useful for making decisions on what time step to use, preconditioners, etc.


anshchaube (approved):
Looks good, thanks!

PR #435, opened 2022-07-26
Allowing OpenMCCellAverageProblem.C to use a libMesh pointer or mesh file during a simulation

lewisgross1296:
This PR corresponds to [an OpenMC PR](https://github.com/openmc-dev/openmc/pull/2137) that added a constructor to OpenMC to create a mesh form a libMesh pointer. A Cardinal simulation can access the pointer of a libMesh mesh and modify it appropriately during simulations, as opposed to creating/reading a mesh input file. This will help facilitate adaptive mesh refinement simulations, as well as core thermal expansion simulations.

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -1746,18 +1746,15 @@ OpenMCCellAverageProblem::initializeTallies()
       {
         // create a local mesh from a libMesh mesh
         local_mesh = std::make_unique<openmc::LibMesh>(_mesh.getMesh(),_scaling);
-        // by setting the ID to -1, OpenMC will automatically detect the next available ID
-        local_mesh->set_id(-1);
-        local_mesh->output_ = false;
       }
       else
       {
         // create a local mesh from the file name
         local_mesh = std::make_unique<openmc::LibMesh>(_mesh_template_filename, _scaling);
-        // by setting the ID to -1, OpenMC will automatically detect the next available ID
-        local_mesh->set_id(-1);
-        local_mesh->output_ = false;
       }
+      // by setting the ID to -1, OpenMC will automatically detect the next available ID
+      local_mesh->set_id(-1);
+      local_mesh->output_ = false;
 
       int32_t mesh_index = openmc::model::meshes.size();
 
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -372,7 +372,6 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
     }
     case tally::mesh:
     {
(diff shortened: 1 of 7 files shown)


src/base/OpenMCCellAverageProblem.C
```diff
@@ -1708,16 +1708,28 @@ OpenMCCellAverageProblem::initializeTallies()
...
-      auto mesh = std::make_unique<openmc::LibMesh>(_mesh_template_filename, _scaling);
-      mesh->set_id(-1);
-      mesh->output_ = false;
+      std::unique_ptr<openmc::LibMesh> local_mesh; // local mesh that will be created from a libMesh pointer or a mesh file
+      if(_mesh_template_filename.empty())
+      {
+        // create a local mesh from a libMesh mesh
+        local_mesh = std::make_unique<openmc::LibMesh>(_mesh.getMesh(),_scaling);
+        // by setting the ID to -1, OpenMC will automatically detect the next available ID
+        local_mesh->set_id(-1);
```
aprilnovak, line 1717: We can move the `local_mesh->set_id` and `local_mesh->output_` lines outside of the if-else statement since they're identical between the two different mesh approaches.

aprilnovak:
Why did `one_mesh_out.e` get re-added? By using that output file from the previous test, we do not want to change it at all, because now that it's been changed there's no way to verify that the previous test is also unchanged. Can you please do a force push where you do _not_ commit any changes to `gold/one_mesh_out.e`?

test/tests/neutronics/mesh_tally/tests
```diff
@@ -12,6 +12,20 @@
...
   []
+  [one_mesh_no_input_file]
+    type = Exodiff
+    input = one_mesh_no_input_file.i
+    exodiff = 'one_mesh_out.e'
+    cli_args = "Outputs/file_base=one_mesh_out"
+    # This test has very few particles, and OpenMC will error if there aren't any particles
+    # on a particular process
+    max_parallel = 32
+    requirement = "The heat source shall be tallied on an unstructured mesh and normalized against "
```
aprilnovak, line 23: Please update this `requirement` line and remove the TODO.

lewisgross1296: Updated

src/base/OpenMCCellAverageProblem.C
lewisgross1296: Moved these outside

lewisgross1296:
Addressed the the two comments. I will remove the WIP once the [OpenMC PR](https://github.com/openmc-dev/openmc/pull/2137) is merged

aprilnovak:
Thank you for the contribution @lewisgross1296 ! I only made some minor changes to documentation and added a few extra checks to warn the user if they don't have the right combination of input parameters.

PR #442, opened 2022-07-28
Fixed source mode

aprilnovak:
This PR adds a fixed source mode to Cardinal (this PR is rebased off of #439, and will be updated once that is merged). The only major difference is how the tallies are normalized. Instead of preserving a user-specified power level, the tally results are converted to power by:

tally (eV / particle) * source_strength (particles / s) * Joule / eV

### DIFF
--- a/include/base/OpenMCCellAverageProblem.h
+++ b/include/base/OpenMCCellAverageProblem.h
@@ -374,7 +374,7 @@ protected:
    * for k-eigenvalue simulations), we are basically relaxing the distribution of the heat
    * source, but then multiplying it by the _current_ mean tally magnitude.
    *
-   * There will be very small errors in these approximations unless power/the source strength
+   * There will be very small errors in these approximations unless the power/source strength
    * change dramatically with iteration. But because relaxation is itself a numerical approximation,
    * this is still inconsequential at the end of the day as long as your problem has converged
    * the relaxed heat source to the raw (unrelaxed) tally.
--- a/include/base/OpenMCCellAverageProblem.h
+++ b/include/base/OpenMCCellAverageProblem.h
@@ -361,7 +361,7 @@ protected:
 
   /**
    * Relax the heat source and normalize it so that it has units of power fraction (i.e. an
-   * integral of unity, where that "integral" is over the entire OpenMC domain if you set
+   * integral of unity, where that "integral" is over the entire OpenMC domain) if you set
    * 'normalize_by_global_tally = true', but only over the Cardinal-created tallies if you
    * instead set 'normalize_by_global_tally = false'.
    *
--- a/include/base/OpenMCCellAverageProblem.h
+++ b/include/base/OpenMCCellAverageProblem.h
@@ -472,8 +472,18 @@ protected:
   void getFissionTallyFromOpenMC(const unsigned int & var_num);
 
   /**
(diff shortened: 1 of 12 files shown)


include/base/OpenMCCellAverageProblem.h
```diff
@@ -359,6 +359,26 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
...
+   * NOTE: This function relaxes the power _distribution_, and not the actual magnitude of the
+   * power. That is, we relax the power distribution and then multiply it by the power
+   * (for k-eigenvalue) or source strength (for fixed source) of _the current step_ before
+   * applying it to MOOSE. If the magnitude of the power is constant in time, there is zero
+   * error in this. But for fixed source simulations where the actual magnitude of the tally
+   * can vary based on simulation (b/c we don't renormalize it in the sense that we do
+   * for k-eigenvalue simulations), we are basically relaxing the distribution of the heat
+   * source, but then multiplying it by the _current_ mean tally magnitude.
+   *
+   * There will be very small errors in these approximations unless power/the source strength
```
pshriwise, line 377:
```suggestion
   * There will be very small errors in these approximations unless the power/source strength
```

include/base/OpenMCCellAverageProblem.h
```diff
@@ -659,7 +693,7 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
    * miss some of that power when sending to MOOSE. So, in this case, it is better to
    * normalize against the local tally itself so that the correct power is preserved.
    */
-  const bool & _normalize_by_global;
+  const bool _normalize_by_global;
```
pshriwise, line 696: Why the changs from the `const` references?

include/base/OpenMCCellAverageProblem.h
```diff
@@ -359,6 +359,26 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
    */
   void storeElementPhase();
 
+  /**
+   * Relax the heat source and normalize it so that it has units of power fraction (i.e. an
+   * integral of unity, where that "integral" is over the entire OpenMC domain if you set
+   * 'normalize_by_global_tally = true', but only over the Cardinal-created tallies if you
+   * instead set 'normalize_by_global_tally = false'.
```
pshriwise, line 366:
```suggestion
  /**
   * Relax the heat source and normalize it so that it has units of power fraction (i.e. an
   * integral of unity, where that "integral" is over the entire OpenMC domain) if you set
   * 'normalize_by_global_tally = true', but only over the Cardinal-created tallies if you
   * instead set 'normalize_by_global_tally = false'.
```

pshriwise (changes requested):
A couple of small corrections to the docstrings, but otherwise this looks great! It'll be nice to have this capability for shielding calcs 👍🏻

include/base/OpenMCCellAverageProblem.h
aprilnovak: This parameter is now set with a `value ? a : b` inside the member initialization list, as opposed to from the `getParam` (which returns a const ref)

pshriwise: Ah I see. Thanks!

PR #448, opened 2022-08-05
Correct understanding about conda vs. script petsc/libmesh

aprilnovak:
Thank you @milljm for correcting my understanding of switching b/w conda vs. scripts petsc/libmesh!


milljm:
FYI, this is getting some TLC in our help pages as well:

https://github.com/idaholab/moose/pull/21769

PR #456, opened 2022-08-16
Skip mesh tally checks when tallying on MooseMesh.

aprilnovak:
When tallying directly on the `[Mesh]`, we don't need to perform any consistency checks between OpenMC's tally mesh and the [Mesh]. This will save some overhead.


aprilnovak:
On second thought I will close this, because this error check previously illuminated the issue with multiplying the actual OpenMC tally mesh when we had the `scaling` parameter turned on.

PR #460, opened 2022-08-18
Add flux tally score

aprilnovak:
This adds a flux tally score, which is necessary to complete some Monte Carlo-TH benchmarks.


aprilnovak:
@pshriwise did you have any comments on this PR? I'd like to merge soon-ish so that @AyaHegazy22 can use this for some fixed source problems.

Thanks 😄

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2163,10 +2168,23 @@ OpenMCCellAverageProblem::checkZeroTally(const Real & power_fraction,
...
 OpenMCCellAverageProblem::tallyMultiplier() const
 {
-  if (_run_mode == openmc::RunMode::EIGENVALUE)
-    return *_power;
+  Real m;
+
+  if (_tally_score == "flux")
+  {
+    // Flux tally has units of particle - cm / source particle; we also only get here for fixed source
+    m = *_source_strength * _local_mean_tally / _scaling;
```
pshriwise, line 2176: I'm wondering, how do we want to indicate the units of this tally to the user? Flux is sort of a unique case in comparison to all the other scores supported here.

Also, are we ever transferring this to a MOOSE mesh? If so, we'd want to perform a volume normalization of the tally based on the MOOSE mesh elements the value is being mapped to.

pshriwise (changes requested):
Really nice flexibility upgrade!! A thought and small comment from me here.

src/base/OpenMCCellAverageProblem.C
aprilnovak: The output of the tally is written to the MOOSE mesh. The units of the tally are particles / length^2 / time, where the length unit is consistent with the MOOSE mesh. 

In other words, if the [Mesh] has units of meters, the flux that is written to the output file has units of particles/m^2/s.

The normalization happens elsewhere in the file, where we divide by length^3.

pshriwise: Right, the heating values are volume-normalized there too. Duh. Thanks for verifying!

pshriwise (approved):

src/base/OpenMCCellAverageProblem.C
aprilnovak: Thanks for the review!

PR #465, opened 2022-09-16
Allow mesh generator to ignore faces that are perpendicular to the axis

aprilnovak:
When moving sidesets to circular boundaries, the `Hex20Generator` currently does not permit a single element to have more than one face on the circular sideset. This can be overly restrictive, especially if that element has some faces on the moving sideset that are actually perpendicular to the specified `axis` and therefore should not move at all. This adds some extra logic to skip those faces perpendicular to the `axis`, giving more flexibility to this mesh generator.


aprilnovak:
Closing because it'd be more robust to simply have the user set up different sidesets for the top/bottom faces to be ignored.

PR #467, opened 2022-09-22
Update NekRS version

aprilnovak:
This updates the NekRS version in Cardinal. Mostly just syntax changes in the input files and lots of changes to the website documentation to make things more user-friendly.

In terms of the test suite, there's only three tests that had non-negligible differences:

- `cht/pebble` [CHT]: relative differences in temperature, flux < 5.5e-3. I re-ran this case for several difference solve tolerances (for pressure, velocity, and temperature solves) and found that our initial choice of a pretty loose tolerance for temperature _does_ have a significant effect on the test output, within the same magnitude as we're seeing now. I don't consider this test to be problematic.
- `boundary_and_volume/prism` [CHT plus volumetric heat source]: relative differences in solid/fluid temperatures increased from 5.5e-6 to 5.5e-3. Still acceptably small. The flux value (integrated) was totally the wrong magnitude, though. Questionable.
- `deformation/simple-cube` [volumetric heat source plus mesh movement]: relative differences in temperature increased from 5.5e-6 to 1e-3. Questionable.

In terms of test coverage, only the `deformation/simple-cube` does not have duplicate test coverage elsewhere (i.e. there are lots of other tests for CHT, and lots of other tests for volumetric heat sources, which did not fail). This is worth looking into with more depth, but a 1e-3 difference in temperature is still fairly small. At the end of the day, we can't control small changes in NekRS related to tolerances, changes to defaults, etc. that might be possibly influencing these solves.


scripts/job_nek5k
```diff
@@ -0,0 +1,23 @@
...
+#SBATCH --ntasks-per-node=40
+#SBATCH --time=00:20:00
+#SBATCH --output=pincell.log
+#SBATCH -p compute
+
+# Revise for your repository and cross section data locations
+export OPENMC_CROSS_SECTIONS=$HOME/cross_sections/endfb71_hdf5/cross_sections.xml
+export CARDINAL_DIR=$HOME/cardinal
+export NEKRS_HOME=$CARDINAL_DIR/install
+
```
anshchaube, line 18:
```suggestion
# need to point to a newer CMake version
export PATH=/shared/cmake-3.24.2/bin:$PATH
```

anshchaube (approved):
Looks good, thanks! I added a small suggestion to ensure consistency between the HPC documentation and the `job_nek5k` script.

scripts/job_nek5k
aprilnovak: This isn't actually necessary - the CMake version is only needed during the compile stage, not running.

PR #493, opened 2022-10-25
Update faq.md for running tests without conda.

ahuxford:
When trying to run tests without the conda environment, libmesh's default installation is `cardinal/contrib/moose/libmesh/installed`, not `cardinal/contrib/moose/libmesh`. 

Using the LIBMESH_DIR in FAQ results in same libmesh-config error message as without the env variable declared.

Correcting LIBMESH_DIR to the '.../installed' directory removes the error and allows for the tests to build and run.

- Tested on Nek5k cluster


aprilnovak:
Thanks @ahuxford!

PR #494, opened 2022-10-25
Adds errors and tests for NekRSSeparateDomainProblem

ahuxford:
These are tests for the NekRSSeparateDomainProblem class
1. Error checking for coupled boundaries
2. Unit tests for temperature transfers
3. Unit tests for velocity transfers

- I will add scalar error checking and unit tests next


aprilnovak:
Thank you for this contribution, @ahuxford! I only skimmed through the content so far, but it looks very thorough and well-done. I'll take a closer look within the next week so that we can merge this soon.

PR #497, opened 2022-10-27
Adds mesh movement using boundary velocities

anshchaube:
This PR adds features to Cardinal that allow the utilization of NekRS's `elasticity` mesh solver, using MOOSE boundary mesh mirrors to calculate mesh velocities necessary for the ALE formulation used by NekRS.


aprilnovak:
This is ready for one last rebase now - latest NekRS version has been merged.

include/base/CardinalEnums.h
```diff
@@ -91,7 +91,10 @@ enum NekWriteEnum
   heat_source,
   x_displacement,
   y_displacement,
-  z_displacement
+  z_displacement,
+  mesh_velocity_x,
+  mesh_velocity_y,
+  mesh_velocity_z
```
aprilnovak, line 97: I see in NekInterface that the names we use to index into the `usrwrk` were renamed, like `displacement_x` because `mesh_velocity_x`. What is the meaning of `x/y/z_displacement` now? Should we be deleting them from `NekWriteEnum`?

anshchaube: I am using `mesh_velocity_x/y/z` for mesh velocities because `displacement_x/y/z` have been used elsewhere for variables associated with volume-based displacements that are never written to `usrwrk` - we add them directly to `mesh->x/y/z`, then copy to device. I don't really see `NekWriteEnum::displacement_x/y/z` being used anywhere, so I think we can delete them from `NekWriteEnum`.

src/base/NekRSProblem.C
```diff
@@ -488,6 +568,11 @@ NekRSProblem::checkInitialFluxValues(const Real & nek_flux, const Real & moose_f
 void
 NekRSProblem::sendVolumeDeformationToNek()
 {
+  double disp_scale = 1.0;
+
+  if (_nondimensional)
```
aprilnovak, line 573: If `nondimensional = false`, the `nekrs::solution::referenceLength()` is set to a default value of 1, so we can use simpler logic here and just divide lines 619-621 straight away by `nekrs::solution::referenceLength()`.

aprilnovak: It's also available as `_L_ref`

src/base/NekRSProblem.C
```diff
@@ -828,4 +919,45 @@ NekRSProblem::addExternalVariables()
...
+      for (int i=0; i <len * len; i++)
+      {
+        tmp_meshvel[i] = ( _displacement_z[i]-_nek_mesh->prev_disp_z().at((e*len*len) + i))
+                           /dt/_U_ref;
+      }
+      writeBoundarySolution(e, field::mesh_velocity_z,tmp_meshvel);
+      _nek_mesh->updateDisplacement(e, _displacement_z, field::z_displacement);
+      break;
+    default:
+      throw std::runtime_error("Unhandled NekWriteEnum in NekRSProblem::interpolateBoundaryDisplacement!\n");
```
aprilnovak, line 959: Please use `mooseError` - it basically wraps this type of command, but is meant to work with the rest of MOOSE's console system.

src/mesh/NekRSMesh.C
```diff
@@ -36,6 +36,7 @@ NekRSMesh::validParams()
   params.addParam<std::vector<int>>("boundary",
                                     "Boundary ID(s) through which nekRS will be coupled to MOOSE");
   params.addParam<bool>("volume", false, "Whether the nekRS volume will be coupled to MOOSE");
+  params.addParam<bool>("moving_mesh", false, "Whether we have a moving mesh problem or not");
```
aprilnovak, line 39: Could you please jog my memory on this - what was the motivation for having this `moving_mesh` parameter be a user-set quantity, if we could technically infer if the mesh is moving from the `.par` file settings? 

Is there ever a situation where we would have mesh deformation turned ON in the `.par` file, but have `moving_mesh = false` (or vice versa)? Just wondering if we can simply some of this logic by accessing information "at the root."

test/tests/deformation/mesh-velocity-areas/nek.i
```diff
@@ -0,0 +1,64 @@
...
+  [nekbdry_ar]
+    type = NekSideIntegral
+    field = unity
+    boundary = '2'
+    use_displaced_mesh = true
+  []
+[]
+
+[Outputs]
+  exodus = false
```
aprilnovak, line 59: I think you can just delete this line, and it will not output exodus.

test/tests/deformation/mesh-velocity-areas/nek.i
```diff
@@ -0,0 +1,64 @@
...
+    field = unity
+    boundary = '2'
+    use_displaced_mesh = true
+    execute_on = INITIAL
+  []
+  [nekbdry_ar]
+    type = NekSideIntegral
+    field = unity
+    boundary = '2'
+    use_displaced_mesh = true
```
aprilnovak, line 54: Do you think we should perhaps set `use_displaced_mesh` automatically, based on the moving mesh settings? I'm sort of on the fence - it would be convenient for the user, but then the user would need to remember to do `use_displaced_mesh = true` for any postprocessors that we're not defining in Cardinal (i.e. any postprocessor in the MOOSE framework itself).

test/tests/deformation/mesh-velocity-areas/pipe.i
```diff
@@ -0,0 +1,200 @@
...
+
+[Kernels]
+  [heat]
+    type = HeatConduction
+    variable = temp
+  []
+  [heat_source]
+    type = CoupledForce
+    variable = temp
+    v = 0.0
```
aprilnovak, line 27: Why not delete this kernel altogether? If the heat source is zero, the residual is zero?

test/tests/deformation/mesh-velocity-areas/pipe.i
```diff
@@ -0,0 +1,200 @@
...
+  []
+[]
+
+[Materials]
+  [thermal]
+    type = HeatConductionMaterial
+    specific_heat = 1.0
+    thermal_conductivity = 1.0
+  []
+  [ddiff]
```
aprilnovak, line 37: Is this material used anywhere?

test/tests/deformation/mesh-velocity-areas/pipe.i
```diff
@@ -0,0 +1,200 @@
...
+  []
+  [heat_source]
+    type = CoupledForce
+    variable = temp
+    v = 0.0
+  []
+[]
+
+[Materials]
+  [thermal]
```
aprilnovak, line 32: I think you can delete this material, if you change `HeatConduction` to `Diffusion`. Because I don't think anywhere in this input file are you needing a thermal conductivity in order to compute a flux (at least I don't see it anywhere).

test/tests/deformation/mesh-velocity-areas/tests
```diff
@@ -0,0 +1,17 @@
+[Tests]
+  [mesh_velocity_areas]
+    type = CSVDiff
+    input = pipe.i
+    csvdiff = 'pipe_out_nek0.csv pipe_out.csv'
+    min_parallel = 4
+    abs_zero = 1e-6
+    rel_err = 5e-4
```
aprilnovak, line 8: If this is a new test, why make the relative error kind of loose? I'd advocate for removing line 8 so that we have a stronger test, unless there's a reason otherwise?

src/mesh/NekRSMesh.C
anshchaube: I don't think we ever spent too much time thinking about this. We did this in summer '21, when nekRS didn't have the elasticity solver and we(mostly me) were still figuring things out.

I can't think of a single scenario where we won't be able to infer the desired setting from the `par` file.

test/tests/deformation/mesh-velocity-nondimensional/pipe.i
```diff
@@ -0,0 +1,200 @@
...
+  []
+[]
+
+[Materials]
+  [thermal]
+    type = HeatConductionMaterial
+    specific_heat = 1.0
+    thermal_conductivity = 1.0
+  []
+  [ddiff]
```
aprilnovak, line 37: Same comments for this input file about the materials & zero heat source kernel

test/tests/deformation/mesh-velocity-nondimensional/pipe.oudf
```diff
@@ -0,0 +1,31 @@
+// Boundary conditions
+void velocityDirichletConditions(bcData *bc)
+{
+  const dfloat pi = M_PI;
```
aprilnovak, line 4: Some of these parameters aren't used here

test/tests/nek_errors/deformation/err3.par
```diff
@@ -0,0 +1,29 @@
...
+stopAt = numSteps
+numSteps = 100
+dt = 1e-3
+timeStepper = tombo2
+writeInterval = 100
+
+[MESH]
+boundaryTypeMap = fixedValue,fixedValue,fixedValue,fixedValue,zeroValue,zeroValue
+solver = elasticity
+file = "nekbox.re2"
```
aprilnovak, line 12: whoa, can you do this to set specifically the name of the mesh file to use?

src/base/NekRSProblem.C
```diff
@@ -828,4 +919,45 @@ NekRSProblem::addExternalVariables()
...
+
+  switch (field)
+  {
+    case field::mesh_velocity_x:
+      for (int i=0; i <len * len; i++)
+      {
+        tmp_meshvel[i] = ( _displacement_x[i]-_nek_mesh->prev_disp_x().at((e*len*len) + i))
+                           /dt/_U_ref;
+      }
+      writeBoundarySolution(e, field::mesh_velocity_x,tmp_meshvel);
```
aprilnovak, line 937: I recently changed `writeBoundarySolution`, because I was encountering some slow performance for a big case. 

Before, `writeBoundarySolution` would first check if the MPI rank matched the elements we're trying to write. But this can be slow, if you go through all the effort to figure out `tmp_meshvel` only to not even write because the MPI rank isn't a match. So I moved the rank check outside of `writeBoundarySolution`.

To be consistent, you'll need to put 

```
        // We can only write into the nekRS scratch space if that face is "owned" by the current process
        if (nekrs::commRank() != _nek_mesh->boundaryCoupling().processor_id(e))
          continue;
```

at the start of this function and any others that are calling `writeBoundarySolution`. Same thing goes for `writeVolumeSolution`.

include/base/CardinalEnums.h
anshchaube: Actually, we are using `field::x/y/z_displacement` with `writeVolumeSolution` for volumetric mesh displacement. That said, there's no reason for these two to be different, except for clarity around variable names (displacement vs mesh-velocity)

aprilnovak: Ok I see. This is fine like you have it.

src/mesh/NekRSMesh.C
```diff
@@ -98,16 +98,28 @@ NekRSMesh::NekRSMesh(const InputParameters & parameters)
...
-    int offset = k * _nek_internal_mesh->Np;
-    for (int v = 0; v < _nek_internal_mesh->Np; ++v)
-    {
-      _initial_x.push_back(_nek_internal_mesh->x[offset + v]);
-      _initial_y.push_back(_nek_internal_mesh->y[offset + v]);
-      _initial_z.push_back(_nek_internal_mesh->z[offset + v]);
-    }
-  }
+
+//  if (_moving_mesh && _volume)
```
aprilnovak, line 102: Can you please explain why this code writing `_initial_x/y/z` was removed? If this isn't used, I'd prefer to not version comments.

anshchaube: This is still in-progress, I will let you know when this is ready for review. Sorry about the commit, but hopefully a final squash will make the commit history cleaner.

aprilnovak: Oh gotcha, sorry for getting ahead of the curve a bit. Let me know when you'd like a re-review.

include/base/NekRSProblem.h
```diff
@@ -196,19 +208,22 @@ class NekRSProblem : public NekRSProblemBase
...
+  double * _mesh_velocity_elem = nullptr;
+
   /// temperature transfer variable written to be nekRS
   unsigned int _temp_var;
 
   /// flux transfer variable read from by nekRS
   unsigned int _avg_flux_var;
 
-  /// x-displacment transfer variable read from for moving mesh problems
+  /// x-displacment transfer variable read from main app for moving mesh problems
```
aprilnovak, line 220: I don't think we need to have "main app" in this Doxygen comment - technically, the displacement could come from  a sub-app. This is super minor, but just want the comments to be as general as possible

src/base/NekRSProblem.C
```diff
@@ -694,20 +851,23 @@ NekRSProblem::syncSolutions(ExternalProblem::Direction direction)
       if (_volume && _has_heat_source)
         sendVolumeHeatSourceToNek();
 
-      // copy the boundary heat flux and/or volume heat source in the scratch space to device
-      nekrs::copyScratchToDevice(_minimum_scratch_size_for_coupling);
 
-      if (_moving_mesh)
+      if (nekrs::hasUserMeshSolver() && _volume)
```
aprilnovak, line 855: Is this boolean here actually equivalent to `if (nekrs::hasUserMeshSolver())`?

src/base/NekRSProblem.C
```diff
@@ -694,20 +851,23 @@ NekRSProblem::syncSolutions(ExternalProblem::Direction direction)
...
-
-        // no boundary-based mesh movement available in nekRS yet
+        sendVolumeDeformationToNek();
+      }
+      else if (nekrs::hasElasticitySolver())
+      {
+        sendBoundaryDeformationToNek();
       }
 
+      // copy the boundary heat flux and/or volume heat source in the scratch space to device
```
aprilnovak, line 864: Should this comment technically be updated to something like "copy the boundary heat flux, volume heat source, and/or volume mesh displacements in the scratch space to device"?

src/base/NekRSProblem.C
```diff
@@ -694,20 +851,23 @@ NekRSProblem::syncSolutions(ExternalProblem::Direction direction)
...
+      nekrs::copyScratchToDevice(_minimum_scratch_size_for_coupling);
+      if (nekrs::hasMovingMesh())
```
aprilnovak, line 866: Should this line actually be `if (nekrs::hasUserMeshSolver())`?

aprilnovak:
Is it possible for us to set `displacements = 'disp_x disp_y disp_z'` from `NekRSMesh` ourselves based on the settings the user has in the .par file? This could reduce the effort for the user, and could eliminate error checks related to `[Mesh]/displacements` being consists with the `.par` settings.

src/actions/NekInitAction.C
```diff
@@ -156,6 +155,11 @@ NekInitAction::act()
 
   _n_cases++;
 
+  if (nekrs::hasMovingMesh() && _type != "NekRSProblem")
```
aprilnovak, line 158: I think this check is a bit narrower than we need - technically, `NekRSStandaloneProblem` should allow any valid NekRS input file to run within Cardinal (it doesn't have any data transfers to/from). So I think we want this to swap out `_type != "NekRSProblem"` with `_type == "NekRSSeparateDomainProblem"`

aprilnovak: After this change is made, I'd suggest adding a `NekRSStandaloneProblem` test that just runs the `mv_cyl` example

src/base/NekRSProblem.C
```diff
@@ -218,6 +223,23 @@ NekRSProblem::initialSetup()
       }
   }
 
+  if (boundary && nekrs::hasMovingMesh() && nekrs::hasElasticitySolver())
```
aprilnovak, line 226: I think that `nekrs::hasMovingMesh()` can be removed here, because `hasMovingMesh()` will be true as long as either `solver = user` or `solver = elasticity`

include/base/NekRSProblem.h
anshchaube: Thanks. How about
`/// x-displacment transfer variable` or `/// x-displacment transfer variable read for moving mesh problems`?

aprilnovak: I like the second one, but either is fine

src/base/NekRSProblem.C
anshchaube: We also want to ensure a volume mesh mirror exists to transfer displacements for all elements, including the interior ones. If the user has `user` mesh solver but just a boundary mesh mirror, we want to stop that and trigger an error (`nek_errors/deformation/missing_volume_for_user_solver`).
The `elasticity` solver works with both `_volume` and `_boundary` so there's no such check for `sendBoundaryDeformationToNek()`

src/base/NekRSProblem.C
anshchaube: Done, thanks!

include/base/NekRSProblem.h
anshchaube: Done, thanks!

src/base/NekRSProblem.C
anshchaube:
```
      if (nekrs::hasMovingMesh())
      {
      // update geometric factors and the mesh on the device
        nekrs::copyDeformationToDevice();
      }
```
We decided to do this every time regardless of whether we are using the `user` mesh solver (with a volume mirror) or the `elasticity` solver (works with either volume or boundary mirrors). This is because it seemed to us that the nekRS team was making changes to their mesh solvers across different versions and sometimes this was necessary with one solver, and sometimes with another. So for consistency, we decided to leave this in for all moving mesh solvers since this update is relatively cheap and ensures that future versions of Cardinal don't break due to changes in nekRS.

src/base/NekRSProblem.C
anshchaube: Done, thanks!

src/base/NekRSProblem.C
aprilnovak: Got it, thanks for the detailed explanation. Would you please add a comment into the code to this effect, in case future developers have the same question?

src/actions/NekInitAction.C
anshchaube: Done, thanks! Waiting for tests to pass...

src/base/NekRSProblem.C
anshchaube: Done, thanks!

src/base/NekRSProblem.C
aprilnovak: Sorry, I wasn't super precise. What I mean is that, earlier we throw an error if `_volume` is false, but the mesh solver is `user`. So by the time we get to this line, this two-term if-statement is redundant, right? Or am I missing some logic?

PR #511, opened 2022-11-14
enable NekRS conjugate heat transer

ahuxford:
This PR allows Cardinal to run NekRS models that include conjugate heat transfer, i.e. a fluid mesh and a solid mesh.
- Users can now choose for Nek postprocessors to act on the entire mesh (fluid+solid domains) or just the flow mesh (fluid domain only). Previously it was the entire mesh by default, but this doesn't make sense for all uses, especially Reynolds Number, mass flow rate, etc.
- I chose flowMesh as the parameter, but maybe there's a more-clear name for this parameter, to choose which mesh to act on.

1. I modified the test in `nek_standalone/conj_ht` to reflect these changes with a gold csv file.
2. All values appear to be correct, easy analytical checks are the differences in Boundary 3's Surface Area and Total Volume between mesh used.
3. I didn't add any MOOSE errors though, I'm not sure what would be helpful for that.
4. To not break implementations in coupling Problem Classes, I made duplicate functions for now in `NekInterface` that choose which mesh to act on given in input boolean. 
5. I know I will need to update `NekRSSeparateDomainProblem` to reflect these changes, so I expect other Classes will need the same.


src/postprocessors/NekPostprocessor.C
```diff
@@ -24,11 +24,15 @@ InputParameters
 NekPostprocessor::validParams()
 {
   InputParameters params = GeneralPostprocessor::validParams();
+
+  params.addParam<bool>("flowMesh", false, "Only compute postprocessor for flow mesh, not solid mesh");
```
aprilnovak, line 28: What do you think about an enumeration instead? We might want to integrate over both the fluid & solid mesh for porous applications. Something like this:

```
domain = 'fluid solid '
domain = 'fluid'
domain = 'solid'
```

ahuxford: Yes that can work and add more capability. I know we can do `domain = 'fluid solid '` and `domain = 'fluid'`, but I'm not confident on how to do `domain = 'solid'`. 

For  `domain = 'fluid'` we can just pull the flowMesh().
- This is same as having the parameter `flowMesh=True` right now in what I implemented. No problem here.

For  `domain = 'fluid solid'` we can just pull the temperatureMesh() (entireMesh()=temperatureMesh() for this case)
- This is same as having the parameter `flowMesh=False` right now in what I implemented. No problem here.

But for `domain = 'solid'` we currently don't have a mesh in Cardinal that points to the solid mesh only, and I'm not sure how to do this. Could you help me with this? Unless we can just do `solidMesh() = temperatureMesh() - fluidMesh()`, that seems too easy.

ahuxford: You're right about the element ordering, and yes I will follow your suggestions.

For `domain = 'fluid solid'` with `field=velocity_x` , it doesn't seg fault because the solid mesh still has velocity(x,y,z) variables they are all just set to zero. So there's no seg faulting.

aprilnovak: Good to know, thanks! 

I think we will also want to avoid having 2 versions of every function - it can become more effort to maintain if we ever need to change the behavior of those functions. I'd suggest just deleting the previous versions of the functions and keeping the new versions with the extra parameter. Perhaps having this extra parameter be a [default argument](https://en.cppreference.com/w/cpp/language/default_arguments) can help with not needing to touch other parts of the code base.

FYI you will need to rebase this branch off the latest devel, I merged some other things earlier this morning.

ahuxford: Okay thanks!

ahuxford:
I updated following your suggestions @aprilnovak. Let me know what you think now.

src/postprocessors/NekPostprocessor.C
```diff
@@ -41,6 +47,10 @@ NekPostprocessor::NekPostprocessor(const InputParameters & parameters)
                "options: 'NekRSProblem', 'NekRSSeparateDomainProblem', 'NekRSStandaloneProblem'");
   }
 
+  // Cannot currently postProcess over only the NekRS solid mesh, must include fluid mesh
+  if (*_pp_mesh.begin() == "solid")
```
aprilnovak, line 51: What if the user does `mesh = 'solid fluid'`? We should allow any ordering

src/postprocessors/NekSidePostprocessor.C
```diff
@@ -30,7 +30,8 @@ NekSidePostprocessor::validParams()
 }
 
 NekSidePostprocessor::NekSidePostprocessor(const InputParameters & parameters)
-  : NekPostprocessor(parameters), _boundary(getParam<std::vector<int>>("boundary"))
+  : NekPostprocessor(parameters),
+    _boundary(getParam<std::vector<int>>("boundary"))
```
aprilnovak, line 34: Why was this line needed?

src/base/NekInterface.C
```diff
@@ -633,9 +633,10 @@ copyDeformationToDevice()
 }
 
 double
-sideMaxValue(const std::vector<int> & boundary_id, const field::NekFieldEnum & field)
+sideMaxValue(const std::vector<int> & boundary_id, const field::NekFieldEnum & field,
+             const MultiMooseEnum & pp_mesh)
 {
-  mesh_t * mesh = entireMesh();
+  mesh_t * mesh = (pp_mesh.size()>1) ? entireMesh() : flowMesh();
```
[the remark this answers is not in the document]
aprilnovak: Maybe we can do the MooseEnum options as `fluid, solid, all`  if that's easier to deal with. I'll leave that up to you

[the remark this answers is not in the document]
ahuxford: My bad, I thought a MultiMooseEnum wouldn't allow `fluid fluid` as an input. And I like having inputs as `fluid solid all` a lot, I'll change to that.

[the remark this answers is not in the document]
aprilnovak: Honestly I thought it also would not allow it as well, but I thought I'd test it and `fluid fluid` gave 2x the fluid volume. I like the `fluid solid all` option better as well, now that I think about it. You'll want to use just a regular `MooseEnum` so that we're sure the user can only put in one of them at a time. Otherwise hopefully it's straightforward to swap.

src/postprocessors/NekMassFluxWeightedSideAverage.C
```diff
@@ -34,6 +34,9 @@ NekMassFluxWeightedSideAverage::validParams()
 NekMassFluxWeightedSideAverage::NekMassFluxWeightedSideAverage(const InputParameters & parameters)
   : NekMassFluxWeightedSideIntegral(parameters)
 {
+  // make sure mesh is only fluid
+  if ( std::find(_pp_mesh.begin(), _pp_mesh.end(), "solid") != _pp_mesh.end() )
```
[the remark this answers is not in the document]
ahuxford: There's nothing wrong in terms of seg faults or anything, I just didn't think it made sense for mass flux. But as I think more, I think for a porous media application users would mayble find this useful actually.

I'll change this to follow the other functions.

src/postprocessors/NekSidePostprocessor.C
ahuxford: Sorry nothing should have been changed here, I'll fix this.

ahuxford:
@aprilnovak I updated things to follow your suggestions. It looks like I messed up the history though when trying to squash my nonconsecutive commits into one. It now shows my authoring some commits that you did, is this normal?

This PR now allows users to input `mesh = 'fluid' 'solid' or 'all'`. There's now an error along with a RunException test when users tries to use `mesh='solid'` because that capability is not available yet.

aprilnovak:
Hmm, yes I see that as well. Can you try rebasing onto `devel` and then doing a force push?

aprilnovak:
I think the `getMesh` function should actually look like this:

```
mesh_t *
getMesh(const MooseEnum & pp_mesh)
{
  if (pp_mesh == "fluid")
    return flowMesh();
  else if (pp_mesh == "all")
    return entireMesh();
  else
    mooseError("Cardinal cannot operate on this mesh region of ", pp_mesh);
}
```

(I swapped the call to `temperatureMesh()` for `entireMesh()`). What do you think? I don't think there's technically anything wrong with the user putting `mesh = 'all'` if the problem is fluid-only (in that case, `mesh = all` should do the same thing as `mesh = fluid`).

ahuxford:
>

aprilnovak:
Also, it looks like in all the postprocessor-related functions, `entireMesh` was replaced by `getMesh` (which defaults with `fluid` as the region). Should the default instead be `all` to match that previous behavior?

ahuxford:
Sure to be consistent I'll change that as well.

src/postprocessors/NekPostprocessor.C
```diff
@@ -41,6 +48,12 @@ NekPostprocessor::NekPostprocessor(const InputParameters & parameters)
                "options: 'NekRSProblem', 'NekRSSeparateDomainProblem', 'NekRSStandaloneProblem'");
   }
 
+  if (_pp_mesh=="solid")
+    mooseError("Cardinal cannot operate soley on the NekRS solid mesh, but this capability will\n"
+               "be added in the future. Please use 'fluid' or 'all' until then.");
+//    mooseError("Cardinal cannot operate soley on the NekRS solid mesh, but this capability will\n"
```
aprilnovak, line 54: Would you please remove these two commented lines?

aprilnovak:
Would you also please reduce the number of time steps in the new `conj_ht` test? I don't think we need to run all 1000 time steps for the sake of testing, and running that many steps will slow down our overall test suite.

test/tests/nek_standalone/conj_ht/tests
```diff
@@ -1,10 +1,23 @@
 [Tests]
-  [test]
+  [nek_example]
+    type = CSVDiff
+    input = nek.i
+    csvdiff = nek_out.csv
+    min_parallel = 2
+    mesh_mode = 'replicated'
```
aprilnovak, line 7: Ah, sorry just noticed this as well - why is the `mesh_mode` here set to replicated? I think this input should also work in distributed mesh cases.

You can run the test suite in distributed mode with

```
./run_tests --parallel 4 --distributed-mesh
```

where 4 is the number of MPI ranks per test.

ahuxford: I think I just copied this from another test and didn't change this, thanks for pointing it out.

ahuxford:
I had to remove a couple unnecessary changes. I think this PR is all clean and ready now.

src/postprocessors/NekPostprocessor.C
```diff
@@ -24,11 +24,18 @@ InputParameters
 NekPostprocessor::validParams()
 {
   InputParameters params = GeneralPostprocessor::validParams();
+
+  MooseEnum mesh("fluid solid all", "fluid");
```
aprilnovak, line 28: I think this default should also be `all` in order to match the previous behavior.

ahuxford:
Okay I had some typos to fix as well. Should be good now.

src/postprocessors/NekMassFluxWeightedSideAverage.C
```diff
@@ -39,7 +39,7 @@ NekMassFluxWeightedSideAverage::NekMassFluxWeightedSideAverage(const InputParame
 Real
 NekMassFluxWeightedSideAverage::getValue()
 {
-  return NekMassFluxWeightedSideIntegral::getValue() / nekrs::massFlowrate(_boundary);
+  return NekMassFluxWeightedSideIntegral::getValue() / nekrs::massFlowrate(_boundary, _pp_mesh);
```
aprilnovak, line 42: function 2

src/postprocessors/NekMassFluxWeightedSideIntegral.C
```diff
@@ -42,7 +42,7 @@ NekMassFluxWeightedSideIntegral::NekMassFluxWeightedSideIntegral(const InputPara
 Real
 NekMassFluxWeightedSideIntegral::getValue()
 {
-  return nekrs::sideMassFluxWeightedIntegral(_boundary, _field);
+  return nekrs::sideMassFluxWeightedIntegral(_boundary, _field, _pp_mesh);
```
aprilnovak, line 45: function 3

PR #531, opened 2022-12-08
Modify NekRSSeparateDomainProblem to work correctly with NekRS CHT models

ahuxford:
This modifies two aspects of the NekRSSeparateDomainProblem class, based on updates from #511.
This class should only act on the Nek fluid mesh, so changes were made to reflect this while allowing for flexibility for future modifications.

1. All Nek-based postProcessors automatically created now act on Nek fluid mesh only.
2. All data transfers to Nek's usrwrk array now require an input mesh (similar to Nek-based postProcessors). However, for now the class should only act on the fluid mesh so a default mesh is specified and always used.


ahuxford:
Actually I think it'd be better to have this input mesh as a input parameter to the problem, where a user can chose which mesh to transfer info on. But I'll set the default to the fluid mesh and have errors when non-fluid mesh is input, because for this class we only want the fluid mesh used.

ahuxford:
This class now takes in a mesh parameter that defaults to the Nek fluid mesh and throws an error if any other input is provided. I also added a test for this mesh parameter as well.

Please let me know what you think when you have the chance @aprilnovak

aprilnovak:
Thanks for this contribution! This all looks good to me - is this ready for merge?

aprilnovak:
Thank you for renaming the test directory in order to share files - this is helpful for reducing out JIT time for these quick error tests.

ahuxford:
> Thanks for this contribution! This all looks good to me - is this ready for merge?

Yes it's ready to merge.

PR #532, opened 2022-12-13
Add support for volume postprocessors over Nek solid mesh

aprilnovak:
This adds support for the volume-based postprocessors over the NekRS solid mesh. A summary of major changes:

- To simplify the switch statements, I fetch the actual value of the enum (an int, as opposed to the char values of `"solid"`, etc.)
- Add support for integrals, averages, and extrema over the solid portions of the NekRS mesh
- I add a test for volume-based postprocessors for the `conj_ht` example, by comparing to equivalent MOOSE postprocessors


test/tests/cht/multi_cht/conj_ht.usr
```diff
@@ -0,0 +1,73 @@
...
+
+      include 'SIZE'
+      include 'TOTAL'
+      include 'NEKUSE'
+
+      ux   = 4.0*y*(1. - y)
+      uy   = 0.0
+      uz   = 0.0
+      temp = 0.0
+
```
ahuxford, line 34: If we're setting the initial conditions from `UDF_Setup` of the `udf` file, we don't need/want to use the `useric` in the `.usr` file. I think the `udf` is just overwriting the `useric`'s initial field conditions, so `useric` in unneeded for this test.

test/tests/cht/multi_cht/nek_pp.i
```diff
@@ -0,0 +1,268 @@
...
+[]
+
+[Problem]
+  type = NekRSStandaloneProblem
+  casename = 'conj_ht'
+  output = 'temperature pressure'
+[]
+
+# This AuxVariable and AuxKernel is only here to get the postprocessors
+# to evaluate correctly. This can be deleted after MOOSE issue #17534 is fixed.
```
ahuxford, line 13: This is a side note not related to this PR, but does this MOOSE issue affect `NekRSSeparateDomainProblem` specifically for SAM-NekRS or THM-NekRS coupling?

ahuxford:
Those are my only comments. Looks good to me, I didn't find any problems.

test/tests/cht/multi_cht/nek_pp.i
aprilnovak: This MOOSE issue affects any class that derives from `ExternalProblem` (all the NekRS classes are affected, as well as the OpenMC ones).

test/tests/cht/multi_cht/conj_ht.usr
aprilnovak: Yes, good point!

aprilnovak:
Closing because this has been revived in a separate PR at #700

PR #542, opened 2022-12-22
Refactor data mapping from MOOSE to Nek

aprilnovak:
Now that we have quite a few different data mappings from MOOSE into NekRS (heat flux, volumetric heat source, x/y/z displacements), we can reduce some duplicate code by creating helper functions that establish the mapping from MOOSE nodal points to NekRS's GLL points.


aprilnovak:
tag @anshchaube for awareness, since this touches quite a bit of the displacement code

PR #550, opened 2023-01-06
Moose fix 17534

aprilnovak:
Remove unnecessary dummy auxkernels from Cardinal's tests after adding @lindsayad's fix. 

The re-golds are fixes to account for the (now correct) output of variables, postprocessors, and userobjects in time. In a few test files, I forgot to put the dummy auxkernel, so our gold files were artificially offset by 1 time step.


lindsayad:
Nice!

PR #559, opened 2023-01-17
Fix for deformation/mesh-velocity and deformation/nek_standalone tests

### LINKED ISSUE #552: Unskip standalone deformation cases
I somehow missed it in #497, but all the tests added for running deformation cases with `NekRSStandaloneProblem` are actually skipped (so they're not being tested). If I enable those tests, they all fail. We need to add these tests to our test suite and properly ensure they are doing what we're intending for them.

The tests I'm referring to are here: `test/tests/deformation/nek_standalone/tests`

anshchaube:
This pull request

- fixes the mesh-velocity test suite by removing flawed tests that had `Mesh/volume = true` but no boundary of interest specified, which resulted in no mesh movement in the sub-app,
- makes the size of variables used in mesh velocity calculations compatible with recent changes from #542 , removing the segfaults occurring in said calculations,
- adds an error to prevent use of the elasticity solver without specifying a boundary of interest,
- closes #552 by specifying the boundary of interest in every case, and removing postprocessors that do not exist on boundary mesh mirrors, and
- makes minor improvements to `NekRSProblem::calculateMeshVelocity` for clarity/readability.


src/base/NekRSProblem.C
```diff
@@ -247,6 +246,14 @@ NekRSProblem::initialSetup()
                  " in the [MESH] block.");
   }
 
+  if (!boundary && nekrs::hasElasticitySolver())
+    mooseError("Your nekRS .par file has 'solver = elasticity' in the [MESH] block. This solver uses\n"
+               "displacement values at a boundary of interest to calcualte the mesh velocity. This\n"
+               "mesh velocity is applied within nekRS on the same boundary to solve for fluid flow\n"
+               "in a moving mesh domain using the elasticity solver. Please indicate which boundary\n"
+               "is moving in your Cardinal sub-app input file's [Mesh] block, regardless of whether\n"
```
aprilnovak, line 254: I think `boundary` should be the union of all boundaries which either have a heat flux BC from MOOSE, a displacement from MOOSE, or both. We just need `boundary` to set up some internal data structures for those transfers. Can you rephrase this to reflect that all the entries in `boundary` don't _need_ to be moving? Or if there is something I'm missing here, let me know.

src/base/NekRSProblem.C
```diff
@@ -785,37 +792,36 @@ NekRSProblem::addExternalVariables()
 void
 NekRSProblem::calculateMeshVelocity(int e, const field::NekWriteEnum & field)
 {
-  int len = _nek_mesh->nekPolynomialOrder() + 1;
+  int len = _volume? _n_vertices_per_volume : _n_vertices_per_surface;
   double dt = _timestepper->getCurrentDT();
 
+  double * displacement = NULL, *prev_disp = NULL;
```
aprilnovak, line 798: `NULL` should be `nullptr` according to newer C++ standards (`NULL` is discouraged)

src/base/NekRSProblem.C
```diff
@@ -785,37 +792,36 @@ NekRSProblem::addExternalVariables()
...
-      _nek_mesh->updateDisplacement(e, _displacement_z, field::z_displacement);
+      displacement = _displacement_z;
+      prev_disp = _nek_mesh->prev_disp_z().data();
+      disp_field = field::z_displacement;
       break;
     default:
       mooseError("Unhandled NekWriteEnum in NekRSProblem::calculateMeshVelocity!\n");
   }
+
+  for (int i=0; i <len; i++)
```
aprilnovak, line 822: Nice! I like this reduction of code.

test/tests/deformation/nek_standalone/nek.i
```diff
@@ -85,7 +85,7 @@
     value_type = min
   []
   [min_p_output]
-    type = NodalExtremeValue
+    type = ElementExtremeValue
```
aprilnovak, line 88: Why this change?

anshchaube: This seems to work with both boundary and volume mirrors (we had a separate discussion about this on Slack). Please let me know if you would like me to change it to something else.

src/base/NekRSProblem.C
anshchaube: Thanks for pointing that out. Yes, it should be a union of all those boundaries. How is this?
```
    mooseError("Your nekRS .par file has 'solver = elasticity' in the [MESH] block. This solver uses\n"
               "displacement values at a boundary of interest to calcualte the mesh velocity. This\n"
               "mesh velocity is applied within nekRS on the same boundary to solve for fluid flow\n"
               "in a moving mesh domain using the elasticity solver. If you intend to use the elasticity\n"
               "solver, please indicate at least one boundary of interest in your Cardinal sub-app\n"
               "input file's [Mesh] block, regardless of whether you are using a volume mirror or not.");
```

include/mesh/NekRSMesh.h
```diff
@@ -319,7 +319,7 @@ class NekRSMesh : public MooseMesh
    * @param[in] e the element to which src belongs
    * @param[in] field the displacement field we are updating
    */
-  void updateDisplacement (const int e, const double *src, const field::NekWriteEnum field);
```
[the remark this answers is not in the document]
anshchaube: I erroneously thought removing it was necessary - it wasn't. Thanks, I will make it `const` again.

src/base/NekRSProblem.C
aprilnovak: looks great, I'd perhaps put `boundary` within single quotes in the second-to-last line, like `'boundary'`, so that it's clear exactly what the user should do to address this error.

aprilnovak:
@anshchaube would you please resolve the merge conflicts and rebase this branch off of `devel`?

test/tests/deformation/nek_standalone/nek.i
aprilnovak: NodalExtremeValue should also work with both boundary and volume mirrors (the only difference is whether the extreme value is searched over the DOFs themselves (the node values) vs. by evaluating the shape functions (the elemental value)).

anshchaube: That's correct, and I have changed it back to `NodalExtremeValue`. Thanks!

src/base/NekRSProblem.C
anshchaube: This change has been made, thanks!

PR #576, opened 2023-02-08
Add option for AMGX build.

aprilnovak:
Adds a Makefile variable to easily turn on the AMGX in Cardinal.


jma6166 (changes requested):
In the makefile if we are using amgx we should enable cuda since it is only applicable on nvidia gpus. Also, we should enable amgx even with denable flag.

aprilnovak:
Can you clarify what you mean by this?

> Also, we should enable amgx even with denable flag.

PR #578, opened 2023-02-08
Use consistent indices data structure for mapping to scratch space. 

aprilnovak:
`NekRSProblem` sets values in an `indices` data structure to make the data writing in the user scratch space a bit more human-readable. This will replace `4 * scalarFieldOffset()` by `indices.boundary_scalar02` (or whatever that slot in the scratch space physically represents).

This PR just makes `NekRSSeparateDomainProblem` consistent with `NekRSProblem`'s approach. This is a necessary prerequisite to integrating NekRS with the MOOSE stochastic tools module, because we're going to have additional scratch space slots and it'll be easier to design that system once everything in Cardinal is self-consistent.

### DIFF
--- a/include/base/NekInterface.h
+++ b/include/base/NekInterface.h
@@ -695,18 +695,41 @@ void storeBoundaryCoupling(const std::vector<int> & boundary_id, int & N);
 
 } // end namespace mesh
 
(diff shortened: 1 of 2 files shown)


ahuxford (approved):
Looks good to me.

PR #582, opened 2023-02-09
Detecting Convergence of the Fission Source via Shannon Entropy

### LINKED ISSUE #577: Quantitative detection of convergence of Shannon Entropy desired
Currently `cardinal/scripts/inative_study.py` will run a simulation to measure the Shannon entropy of the fission source and eigenvalue. The purpose is to ensure the fission source is properly converged by the number of inactive batches. This is important so that when active batches begin, the tallies are free of contamination from any initial guess of the fission source that may be lingering. 

This script is quite useful for visualizing the fission source and eigenvalue's entropy. However, as is, the user has to visually inspect the plots to decide a number of inactive batches. If the user picks a conservative guess, likely their simulation will be fine, but it would be useful to add a way to quantify if the source is converged or not. According to [Brown (2006)](https://www.oecd-nea.org/science/wpncs/sccsa/documents/brown-physor-2006.pdf)
```
MCNP5 will then report the first cycle found (active or inactive) where Hsrc falls with 
one standard deviation of its average for the last half of cycles, along with a recommendation 
that at least that many cycles should be discarded (inactive).
```

I will add extra functionality to this script to detect stationarity a few ways and report to the user at which inactive batch number the source is considered converged according to the above logic.

lewisgross1296:
This PR closes #577. It provides the user three options for detecting convergence using Shannon entropy. All methods assess the standard deviation and mean for some set of entropy values. If the current batch's entropy value is within one standard deviation of the mean (of the specified number of batches), it will report this batch number and recommend using at least this many inactive cycles. 

I tested this locally for a case with multiple layers for each detection type, and it went smoothly. If desired, I can provide the example, but I don't think those files need to be pushed. The script is setup to work assuming a user provides the appropriate arguments and input files, so other developers can test that this works locally using their own Cardinal simulation if desired.

### DIFF
--- a/doc/content/tutorials/convergence.md
+++ b/doc/content/tutorials/convergence.md
@@ -62,6 +62,7 @@ To determine an appropriate number of inactive batches, you can use the
   of cell divisions to create
 - A Cardinal input file that runs your OpenMC model given some imposed
   temperature and density distribution
+- Criteria used to assess the convergence of the fission source
 
 For example, to run an inactive cycle study for the model developed
 for [the gas compact tutorial](https://cardinal.cels.anl.gov/tutorials/gas_compact.html),
@@ -69,7 +70,7 @@ we would run:
 
 ```
 cd tutorials/gas_compact
-python ../../scripts/inactive_study.py -i unit_cell -input openmc.i
+python ../../scripts/inactive_study.py -i unit_cell -input openmc.i --method=all
 ```
 
 since the Python script used to generate the OpenMC model is named
@@ -82,9 +83,10 @@ of cell layers and number of inactive cycles to consider:
 When the script finishes running, plots of Shannon entropy and $k$, along
 with the saved statepoint files from the simulations, will be available
 in a directory named `inactive_study`. Below are shown example images of
-the Shannon entropy and $k$ plots. You would then select the number of
-inactive cycles to be the point at which both the Shannon entropy and $k$
-reach stationarity.
+the Shannon entropy and $k$ plots. Based on the method of detecting stationarity,
+the script will recommend a minimum number of inactive batches to be used 
+in order to converge the fission source for each layer. Picking the most conservative
+answer will help ensure your simulation is discarding enough batches before tallying.
 
 !media example_entropy_plots.png
   id=entropy
--- a/doc/content/tutorials/convergence.md
+++ b/doc/content/tutorials/convergence.md
@@ -69,7 +69,7 @@ we would run:
 
 ```
 cd tutorials/gas_compact
-python ../../scripts/inactive_study.py -i unit_cell -input openmc.i --method=all
+python ../../scripts/inactive_study.py -i unit_cell -input openmc.i
 ```
 
 since the Python script used to generate the OpenMC model is named
@@ -82,10 +82,16 @@ of cell layers and number of inactive cycles to consider:
 When the script finishes running, plots of Shannon entropy and $k$, along
 with the saved statepoint files from the simulations, will be available
 in a directory named `inactive_study`. Below are shown example images of
-the Shannon entropy and $k$ plots. Based on the method of detecting stationarity,
-the script will recommend a minimum number of inactive batches to be used
-in order to converge the fission source for each layer. Picking the most conservative
-answer will help ensure your simulation is discarding enough batches before tallying.
+the Shannon entropy and $k$ plots. If a method to detect convergence of
+the fission source is supplied (the default is none), then the script will
+report whether the chosen metric succeeded or not. Note, there could be various
+factors affecting whether the metric succeeded or not, and based on user choices,
+it is possible for a metric to succeed when more batches or more particles per
+batch are truly needed. These metrics should be considered along side inspection
+of the plots of the Shannon Entropy, which typically takes longer than the
+eigenvalue to converge. Having a converged fission source is important so that
+when the active batches begin tallying, there is no contamination from the 
+source's initial guess impacting the tallies.
 
 !media example_entropy_plots.png
   id=entropy
--- a/doc/content/tutorials/convergence.md
+++ b/doc/content/tutorials/convergence.md
@@ -5,10 +5,12 @@ In this tutorial, you will learn how to:
 - Use helper scripts in Cardinal to assess convergence of the OpenMC solution
   and of a general MOOSE application
 
-To access this tutorial,
+To access this tutorial, enter the following directory and then also generate
+the necessary mesh file,
 
 ```
 cd cardinal/tutorials/gas_compact
+cardinal-opt -i mesh.i --mesh-only
 ```
 
 !alert! note title=Computing Needs
--- a/doc/content/tutorials/convergence.md
+++ b/doc/content/tutorials/convergence.md
@@ -8,7 +8,7 @@ In this tutorial, you will learn how to:
 To access this tutorial,
 
 ```
-cd cardinal/gas_compact
+cd cardinal/tutorials/gas_compact
 ```
 
 !alert! note title=Computing Needs
@@ -83,7 +83,7 @@ When the script finishes running, plots of Shannon entropy and $k$, along
 with the saved statepoint files from the simulations, will be available
 in a directory named `inactive_study`. Below are shown example images of
 the Shannon entropy and $k$ plots. Based on the method of detecting stationarity,
-the script will recommend a minimum number of inactive batches to be used 
+the script will recommend a minimum number of inactive batches to be used
 in order to converge the fission source for each layer. Picking the most conservative
 answer will help ensure your simulation is discarding enough batches before tallying.
 
--- a/doc/content/tutorials/convergence.md
+++ b/doc/content/tutorials/convergence.md
@@ -62,7 +62,6 @@ To determine an appropriate number of inactive batches, you can use the
   of cell divisions to create
 - A Cardinal input file that runs your OpenMC model given some imposed
   temperature and density distribution
-- Criteria used to assess the convergence of the fission source
 
 For example, to run an inactive cycle study for the model developed
 for [the gas compact tutorial](https://cardinal.cels.anl.gov/tutorials/gas_compact.html),
--- a/scripts/inactive_study.py
+++ b/scripts/inactive_study.py
@@ -20,18 +20,21 @@
 # in order to determine how many inactive cycles are needed to converge both the
 # Shannon entropy and k. This script is run with:
 #
-# python inactive_study.py -i <script_name> -input <file_name> [-n-threads <n_threads>]
+# python inactive_study.py -i <script_name> -input <file_name> --method=method [--winow_lenth=<LENGTH>]  [-n-threads <n_threads>]
 #
-# - the script used to create the OpenMC model is named <script_name>.py;
-#   this script MUST accept '-s' as an argument to add a Shannon entropy mesh
-#   AND '-n' to set the number of layers; please consult the tutorials for
-#   examples if you're unsure of how to do this
-# - the Cardinal input file to run is named <file_name>
-# - by default, the number of threads is taken as the maximum on your system;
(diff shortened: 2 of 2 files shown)


scripts/inactive_study.py
```diff
@@ -64,14 +67,23 @@
...
 ap = ArgumentParser(description=program_description)
 
-ap.add_argument('-i', dest='script_name', type=str,
-                help='Name of the OpenMC python script to run (without .py extension)')
 ap.add_argument('-n-threads', dest='n_threads', type=int,
                 default=multiprocessing.cpu_count(), help='Number of threads to run Cardinal with')
-ap.add_argument('-input', dest='input_file', type=str,
+ap.add_argument('-i', dest='script_name', type=str, required=True,
+                help='Name of the OpenMC python script to run (without .py extension)')
+ap.add_argument('-input', dest='input_file', type=str, required=True,
```
aprilnovak, line 74: For future PRs, try not to unecessarily add diffs if the behavior isn't any different. i.e. the `-i` arg parser line isn't actually any different, you just moved which line it's on. This can make things harder to review.

lewisgross1296: Ah sorry about that. I was getting an error, and for a second thought I needed to put all the non required arguments at the top (but I think that wasn't the problem and my brain was thinking of c++ and defaults). I can put it back if you want. I forgot to move it back once the script wasn't giving errors.

lewisgross1296:
I just grep'd and realized there is a tutorial that mentions this script. Should I update it as well? I added to the documentation in the script itself, but not in `doc/content/tutorials/convergence.md`

aprilnovak:
Yes, if there is documentation associated with something it should be updated.

lewisgross1296:
Ready for final review (or to be merged if it looks good)

doc/content/tutorials/convergence.md
```diff
@@ -62,14 +62,15 @@ To determine an appropriate number of inactive batches, you can use the
   of cell divisions to create
 - A Cardinal input file that runs your OpenMC model given some imposed
   temperature and density distribution
+- Criteria used to assess the convergence of the fission source
```
aprilnovak, line 65: I would remove this - it's not really a "need" in the same way that an input file is a hard, tangible need.

scripts/inactive_study.py
```diff
@@ -20,18 +20,21 @@
 # in order to determine how many inactive cycles are needed to converge both the
 # Shannon entropy and k. This script is run with:
 #
-# python inactive_study.py -i <script_name> -input <file_name> [-n-threads <n_threads>]
+# python inactive_study.py -i <script_name> -input <file_name> --method=method [--winow_lenth=<LENGTH>]  [-n-threads <n_threads>]
```
aprilnovak, line 23: `winow` -> `window`

scripts/inactive_study.py
aprilnovak: Yes, please move it back so that this reflects new content.

Also, is it possible to set a default for `--method`? Good practice is to avoid the user needing to set options when there can be a sensible default. Will be less typing.

scripts/inactive_study.py
```diff
@@ -64,14 +67,23 @@
...
+                help='Name of the OpenMC python script to run (without .py extension)')
+ap.add_argument('-input', dest='input_file', type=str, required=True,
                 help='Name of the Cardinal input file to run')
-
+ap.add_argument('--method', dest = 'method', choices =['all','half','window'], required=True,
+                help = "The type of way to detect steady state. Options are all, half, or window")
+opts, rem_args = ap.parse_known_args()
+if(opts.method == "window"):
+    ap.add_argument('--window_length', dest = 'window_length',required = True, type = int,
+                    help =" The window length must be specified if window is the detectio method")
```
aprilnovak, line 81: detectio -> detection

aprilnovak: Please also change the documentation string here to document what this parameter is, not necessarily the conditions when you need it.

scripts/inactive_study.py
```diff
@@ -64,14 +67,23 @@
...
 ap.add_argument('-n-threads', dest='n_threads', type=int,
                 default=multiprocessing.cpu_count(), help='Number of threads to run Cardinal with')
-ap.add_argument('-input', dest='input_file', type=str,
+ap.add_argument('-i', dest='script_name', type=str, required=True,
+                help='Name of the OpenMC python script to run (without .py extension)')
+ap.add_argument('-input', dest='input_file', type=str, required=True,
                 help='Name of the Cardinal input file to run')
-
+ap.add_argument('--method', dest = 'method', choices =['all','half','window'], required=True,
+                help = "The type of way to detect steady state. Options are all, half, or window")
```
aprilnovak, line 77: Can you add some extra information on what the 3 options do?

scripts/inactive_study.py
```diff
@@ -64,14 +67,23 @@
...
+opts, rem_args = ap.parse_known_args()
+if(opts.method == "window"):
+    ap.add_argument('--window_length', dest = 'window_length',required = True, type = int,
```
aprilnovak, line 80: Don't add parameters inside of if-statements. If the user does `./inactive_study.py --help`, then they would never see that `window_length` is a valid parameter. Please instead leave it as a top-level parameter, but throw an error if the user does _not_ specify if when using the window detection method.

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+  nl = n_layers[i]
+  if(method == "window"):
+      # detect when entropy of current batch is within window (mean +/- std of window)
+      for j in range(window_length,len(entropy[i])):
+          window = entropy[i][(j-window_length):j]
+          window_mean = np.average(window)
+          window_dev = np.std(window)
+          window_low = window_mean - window_dev
+          window_high = window_mean + window_dev
+          # print("window bounds: ", window_low , " to ", window_high )
```
aprilnovak, line 190: This can be deleted

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+          if(window_high - entropy[0][j] > 0 and entropy[0][j]-window_low > 0):
```
aprilnovak, line 191: Shouldn't entropy be indexed as `[i][j]` here?

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+              # if entropy[j] is less than window_high and greater than
+              # window_low, then it is within the window and we've succeeded
+              print("For layer", nl, ", batch", j, "produced a value within one "
+                    "standard deviation of the window mean. It is recommended to do "
+                    "at least this many inactive cycles for this layer.")
+              break
+          else:
+              continue
+  elif(method == "half"):
+      #  Accordinig to Brown, the below is done in MCNP5:
```
aprilnovak, line 201: Can you put a specific reference here, in case someone wants to cite this MCNP method?

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+      # find the first cycle when Hsrc is within one standard deviation of its average for the last half of cycles
+      # compute on the fly the average of the eigenvalue (last half batches) +/- the std dev of the last half cycles
+      # and see if the current k is within this interval
+      for j in range(1,len(entropy[i])):
+          last_half_idx = int(np.floor(j/2))
+          last_half_vals = entropy[0][last_half_idx:j]
```
aprilnovak, line 207: same question about 0 vs. i indexing

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+          stdev = np.std(last_half_vals)
+          mean = np.mean(last_half_vals)
+          low =  mean - stdev
+          high = mean + stdev
+          if(high - entropy[0][j] > 0 and entropy[0][j]-low > 0):
```
aprilnovak, line 212: same question about 0 vs. i indexing

scripts/inactive_study.py
```diff
@@ -162,3 +174,60 @@
...
+  else:
+      # use all batches as data points
+      for j in range(1,len(entropy[i])):
+          entropy_slice = entropy[i][0:j]
+          stdev = np.std(entropy_slice)
+          mean = np.mean(entropy_slice)
+          low =  mean - stdev
+          high = mean + stdev
+          if(high - entropy[i][j] > 0 and entropy[i][j]-low > 0):
+              print("For layer", nl, ", batch", j, "produced a value within one "
```
aprilnovak, line 228: I am wondering how these multiple print statements look when shown to the screen. If the user has, say, 8 different choices for layers, this might get quite busy. What do you think about showing the same info, but in a table? 

I'd also suggest showing some information telling the user if they didn't run enough inactive batches at all to meet the criteria. I think right now, that currently nothing will print out at all (which would be confusing).

aprilnovak (commented):
That's all my comments for the first round of review - thanks for putting this together! Mostly small things related to wording.

scripts/inactive_study.py
lewisgross1296: Do you want this extra info in the `help` doc string or in the comments at the top of the file?

scripts/inactive_study.py
aprilnovak: I think a default of "none" is preferred - those methods you are adding are not foolproof, and I prefer to err on the side of caution of not giving over-confidence to the user

scripts/inactive_study.py
aprilnovak: in the help doc string

lewisgross1296: Let me know if the most recent push addressed your concerns.

scripts/inactive_study.py
lewisgross1296: Any other places in Cardinal that make a table the way you want I could copy / things to `grep` for?

I can add a message for the case that convergence is not detected with the user settings.

aprilnovak: Nope, go ahead and just make a table you think looks nice

lewisgross1296: I've been looking into python packages that automate table formatting, but it seems they all require a pip install. Do you have reservations about relying on something like this or do you think it is fine for the user to run the script, see they are missing the package and then figure out they need to install it? I can put some kind of note in the script that just tells them to pip install as well for extra support.

doc/content/tutorials/convergence.md
```diff
@@ -82,9 +82,16 @@ of cell layers and number of inactive cycles to consider:
 When the script finishes running, plots of Shannon entropy and $k$, along
 with the saved statepoint files from the simulations, will be available
 in a directory named `inactive_study`. Below are shown example images of
-the Shannon entropy and $k$ plots. You would then select the number of
-inactive cycles to be the point at which both the Shannon entropy and $k$
-reach stationarity.
+the Shannon entropy and $k$ plots. If a method to detect convergence of
```
lewisgross1296, line 85:
```suggestion
the Shannon Entropy and $k$ plots. If a method to detect convergence of
```

aprilnovak:
Hi @lewisgross1296 - thank you for this contribution. I just pushed one commit on top of yours, which compresses the changes made to the `inactive_study.py` script. There's a lot of code duplication there - the three methods really just differ in how they select some window for computing the entropy, and that can be consolidated and reduce code maintenance efforts long-term. Please take a look.

I also changed the table print so that it shows actual numbers there and shows the results for every batch. For example, if someone was using the `half` method and their 3rd inactive batch happened to be within 1 standard deviation of the previous half (which is statistically quite possible), the original message printed to them could be misleading. The new table print eliminates any type of suggestion, and simply shows the user the calculation they asked for. This is generally better than providing subjective advice, because there's no chance of misleading someone.

lewisgross1296:
Thanks for getting to this! The elimination of code duplication is definitely an improvement. Tested locally and printout table looks great. I like the `extra_str = "--> "` to indicate which batches satisfy the criteria the user chose.

Only thing left I'd like to include is a print statement that explains `"--> "`. I'll add that in a commit now

PR #599, opened 2023-02-17
Reset tallies between Picard iterations

### LINKED ISSUE #461: Move initializeTallies() and checkMeshTemplateAndTranslations() from the constructor to OpenMCCellAverageProblem::setupProblem()
Currently, the OpenMCCellAverageProblem constructor initializes the tallies and checks the mesh template element ordering and centroids to make sure the mesh centroids are as expected. However, to run Adaptive Mesh Refinement simulations, it would make more sense to move these functions to the `setupProblem()` function that would be called each time the mesh changes. When the mesh changes, these steps will need to be done again. As is, they currently do not. Moving these functions will fix this issue and since the constructor calls `setupProblem()`, it will still do these two important steps when the OpenMCCellAverageProblem constructor is called.

aprilnovak:
If the `[Mesh]` is changing (from adaptivity or spatial deformation), then the tallies created by Cardinal need to be erased and re-created. For mesh tallies, this is obvious because we are tallying on the `[Mesh]`. For cell tallies, it is unlikely, but possible that as the `[Mesh]` changes that new cells are "exposed" to the coupling (because we only add tallies to fissile cells that map to some MOOSE elements). In this case, we'd need to add those cells to the cell instance filter.

These changes simply (i) delete the local and global tallies, (ii) delete the tally filters, and (iii) delete the meshes (if using mesh tallies), then re-creates them. I tested this by turning on this step-by-step regeneration on an existing test, to check that this has no influence on the results. 

The other changes in this PR are just moving existing code into the `setupProblem()` function so that it is called on every time step (if needing to re-init the coupling for moving-mesh reasons).

Closes #461

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -505,7 +505,7 @@ OpenMCCellAverageProblem::initialSetup()
 {
   OpenMCProblemBase::initialSetup();
 
-  if (_adaptivity.isOn() && _need_to_reinit_coupling)
+  if (_adaptivity.isOn() && !_need_to_reinit_coupling)
     mooseError("When using mesh adaptivity, 'fixed_mesh' must be false!");
 
   if (isParamValid("symmetry_mapper"))
@@ -518,15 +518,22 @@ OpenMCCellAverageProblem::initialSetup()
   }
 
   setupProblem();
-
-  initializeTallies();
-
-  checkMeshTemplateAndTranslations();
 }
 
 void
 OpenMCCellAverageProblem::setupProblem()
 {
+  // establish the local -> global element mapping for convenience
+  _local_to_global_elem.clear();
+  for (unsigned int e = 0; e < _mesh.nElem(); ++e)
+  {
+    const auto * elem = _mesh.queryElemPtr(e);
+    if (!isLocalElem(elem))
+      continue;
+
+    _local_to_global_elem.push_back(e);
+  }
+
   initializeElementToCellMapping();
 
   getMaterialFills();
@@ -548,6 +555,10 @@ OpenMCCellAverageProblem::setupProblem()
   }
 
   subdomainsToMaterials();
+
+  initializeTallies();
+
+  checkMeshTemplateAndTranslations();
 }
 
 void
@@ -1755,6 +1766,8 @@ OpenMCCellAverageProblem::storeTallyCells()
   cellInfo first_tally_cell;
   Real mapped_tally_volume;
 
+  _tally_cells.clear();
+
   for (const auto & c : _cell_to_elem)
   {
     auto cell_info = c.first;
@@ -1850,7 +1863,7 @@ OpenMCCellAverageProblem::meshFilter()
   tally_mesh->output_ = false;
   _mesh_template = tally_mesh.get();
 
-  int32_t mesh_index = openmc::model::meshes.size();
+  _mesh_index = openmc::model::meshes.size();
   openmc::model::meshes.push_back(std::move(tally_mesh));
 
   std::vector<openmc::Filter *> mesh_filters;
@@ -1859,7 +1872,7 @@ OpenMCCellAverageProblem::meshFilter()
   {
     const auto & translation = _mesh_translations[i];
     auto meshFilter = dynamic_cast<openmc::MeshFilter *>(openmc::Filter::create("mesh"));
-    meshFilter->set_mesh(mesh_index);
+    meshFilter->set_mesh(_mesh_index);
     meshFilter->set_translation({translation(0), translation(1), translation(2)});
     mesh_filters.push_back(meshFilter);
   }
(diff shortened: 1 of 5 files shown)


src/base/OpenMCCellAverageProblem.C
```diff
@@ -1835,39 +1872,87 @@ OpenMCCellAverageProblem::meshFilter()
...
 }
 
+void
+OpenMCCellAverageProblem::resetTallies()
+{
+  if (_needs_global_tally)
+  {
+    // erase tally
+    auto idx = openmc::model::tallies.begin() + _global_tally_index;
+    openmc::model::tallies.erase(idx);
```
pshriwise, line 1890: We'll want to remove entries in the `model::tally_map` as well.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1835,39 +1872,87 @@ OpenMCCellAverageProblem::meshFilter()
...
+    {
+      // erase tallies
+      openmc::model::tallies.erase(idx, idx + _mesh_translations.size());
+
+      // erase filters
+      auto fidx = openmc::model::tally_filters.begin() + _filter_index;
+      openmc::model::tally_filters.erase(fidx, fidx + _mesh_translations.size());
+
+      // erase meshes
+      auto midx = openmc::model::meshes.begin() + _mesh_index;
```
pshriwise, line 1916: I'm not sure we need to erase the meshes. If the `libMesh` mesh updates, the OpenMC mesh should automatically update (i.e. `n_bins`, `get_bin`, and such should all behave correctly.

pshriwise: Hrmm though I'm not sure if libMesh point locators will update automatically. Lemme check...

pshriwise (changes requested):
Just a couple of small comments from me. I think ideally we wouldn't need to delete meshes along with the tally objects, but I'm following up on something there so I'll let you know.

In the long run, we should add some command to OpenMC like `update_mesh` or something if any data structures need to change on that side IMO. It'll make bookkeeping here easier.

src/base/OpenMCCellAverageProblem.C
aprilnovak: In general I think we will need to erase the meshes, because this code is developed for the purpose of either (i) adaptive mesh refinement or (ii) moving meshes. In both cases, the actual mesh is different, so I think we'd just be wasting memory storing an old mesh?

pshriwise: The safest thing to do for now is to recreate them though, so we can always come back and remove this extra work later.

aprilnovak: That may be the case, good point. But at least right now, I think the particular functions being called in OpenMC are not designed for adaptive mesh refinement. For instance,

```
int LibMesh::n_bins() const
{
  return m_->n_elem();
}
```

I'm pretty sure this is going to return the sum of the "active" elements and "inactive" elements for AMR, and not the "actual" elements I currently want to tally on in my mesh.

aprilnovak: > Will removing & re-adding the mesh help with that though?

No, I guess it wouldn't - sorry, I'm confusing a few different (closely-related) topics we're discussing.

> to ensure the point locators attached to the OpenMC mesh object are updated correctly.

Do you mean that you talked with libMesh devs and they said the point locators wouldn't be updated? Just want to be sure I understand if I should put a TODO comment in Cardinal to remind us that we might remove it in the future.

aprilnovak: Posting slack convo here for posterity, re: point locators

Screen Shot 2023-02-27 at 11 49 20 AM

pshriwise: It's my hope that we can remove it in the future, so if you're okay with it let's pop a TODO in there or make an issue

src/base/OpenMCCellAverageProblem.C
aprilnovak: Shouldn't that quantity stay the same? If I'm going to re-init the same tally, just with potentially a different filter?

pshriwise: That's a map of from tally ID to the tally's index into the `openmc::model::tallies` vector. You may be right that Cardinal will generally replace all of the entires in this map and there won't be any invalid entires, but as a matter of good practice I'd recommend removing the entries for the removed tallies from this map as well to protect from a potential bug in the future if we change how tallies are handled for some reason.

aprilnovak: Actually, does this code not already handle the `tally_map` removal? Once I delete the unique ptr, and it goes out of scope?

```
Tally::~Tally()
{
  model::tally_map.erase(id_);
}
```

pshriwise: Oh yeah you're totally right! Thanks for spotting that. In fact, I'm now noticing that there are other classes that would benefit from this line in their constructor.

Onward!

pshriwise (approved):
Thanks @aprilnovak!

PR #621, opened 2023-03-09
Add PETSC_ARCH logic, Add DAGMC/NekRS link libraries

### LINKED ISSUE #620: simplify the build requirement process a bit
Alter several Makefiles to relax some restrictions on what build environment the user must have in order to achieve a successful build.

While building the Cardinal Conda package #594, I have found several CMAKE lines which were not necessary and in some cases hindered builds in some environments (like Conda).

milljm:
When PETSC_ARCH is physically not available, and is using the default arch-moose, use PETSC_DIR instead. Error if PETSC_ARCH is not available and is not the default.

Add additional DAGMC and NekRS link lines.

Remove -DCMAKE_* defaults already being set in NekRS config. Remove -DCMAKE_* args that are no longer valid for the project.

Remove overrides for CC, CXX, CFLAGS, etc. These are being set and honored within build.mk appropriately. I also found these were overriding flags necessary when building within a Conda Build environment.

Closes #620


Makefile
```diff
@@ -303,7 +303,6 @@ GEN_REVISION       := no
 DEP_APPS           := $(shell $(FRAMEWORK_DIR)/scripts/find_dep_apps.py $(APPLICATION_NAME))
 
 ifeq ($(ENABLE_DAGMC), yes)
-  ENABLE_DAGMC     := ON
```
aprilnovak, line 304: I did this so that I could use `ENABLE_DAGMC` to pass "ON" into OpenMC.

Line 8 here: https://github.com/neams-th-coe/cardinal/blob/devel/config/openmc.mk

Do I not need to do that? I'm not sure I ever tried seeing if some other "truthy" value would be equivalent to ON.

milljm: Oh I didn't realize the variable was again used on line 8. Okay, I can revert this, and use 'ON' instead of 'yes' on line 340

milljm: I even forgot to change it in unit/Makefile anyway... very sloppy of me.

milljm:
While attempting to enable several controlled apps, I am unable to build. So this is still a WIP.

milljm:
Closing this, I believe current HEAD has this all sorted... as we are now building a Cardinal container.

PR #623, opened 2023-03-11
Allow different orders in bc->wrk

aprilnovak:
This FY we are adding support for Nek to interface with MOOSE's stochastic tools module. This means that we need to add another "slot" to `nrs->usrwrk` to hold that incoming information. If we were to keep our current convention where the "ordering" in `nrs->usrwrk` never changes, this means that we're going to end up with quite a lot of unused space allocated (e.g. the first 6 slots) if we only are interested in stochastic modeling.

This PR changes our convention to NOT retain the same order in `nrs->usrwrk` all the time. In other words, if there is a slot in `nrs->usrwrk` that a particular case does not need, we shift all the other slots forward by one, so that there's no "unused gaps" in this array.

I _think_ I caught all places in the documentation and test suite where we need to change this. @anshchaube @ahuxford what do you think of this design change? I think it is nominally only really going to pop up in your use cases, since your situations had more occurrences where there were dummy, empty slots. Are there any considerations I haven't thought of here? If you are both agreeable to this change:

- @anshchaube would you please do a read through `NekRSProblem.md` and make sure I didn't miss anything referring to a rigid order in scratch space?
- @ahuxford would you please do a read through `NekRSSeparateDomainProblem.md` and make sure I didn't miss anything referring to a rigid order in scratch space?

Please note that you'll both probably have to update any local files you have on unmerged branches. Hopefully this isn't too onerous.

### DIFF
--- a/src/base/NekRSProblem.C
+++ b/src/base/NekRSProblem.C
@@ -83,35 +83,35 @@ NekRSProblem::NekRSProblem(const InputParameters & params)
   if (!_boundary)
     checkUnusedParam(params, "conserve_flux_by_sideset", "coupling NekRS solely through a volume mesh mirror");
 
-  // Determine an appropriate default usrwrk indexing; the ordering will always be
-  //   0: flux             (if _boundary is true)
-  //   1: heat_source      (if _volume is true and _has_heat_source is true)
-  //   2: mesh_velocity_x   (if nekrs::hasElasticitySolver() is true)
-  //   3: mesh_velocity_y   (if nekrs::hasElasticitySolver() is true)
-  //   4: mesh_velocity_z   (if nekrs::hasElasticitySolver() is true)
-  //
-  // The most we will do is skip allocating terms at the end of this ordering if we
-  // don't need them. We never change the ordering of "earlier" terms.
-  std::vector<std::string> str_indices =
-    {"flux", "heat_source", "mesh_velocity_x", "mesh_velocity_y", "mesh_velocity_z"};
-  indices.flux = 0 * nekrs::scalarFieldOffset();
-  indices.heat_source = 1 * nekrs::scalarFieldOffset();
-  indices.mesh_velocity_x = 2 * nekrs::scalarFieldOffset();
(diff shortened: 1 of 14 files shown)


ahuxford (approved):
Yes I think this change works fine for NekRSSeparateDomainProblem. I think the updated table + updated example is clear for the new usage, so I don't have any changes to recommend there.

I also don't see anything that could cause issues with the problem class. So it's good on my end.

src/base/NekRSProblem.C
```diff
@@ -83,35 +83,35 @@ NekRSProblem::NekRSProblem(const InputParameters & params)
...
+  if (_volume && _has_heat_source)
+  {
+    indices.heat_source = start++ * nekrs::scalarFieldOffset();
+    str_indices.push_back("heat_source");
+  }
 
-      if (!_boundary)
-        str_indices.erase(str_indices.end());
-    }
+  if (nekrs::hasElasticitySolver())
```
anshchaube, line 107: One minor comment: for `user` mesh solver cases (volume-mirror based, uses displacements directly), should we also include a similar `if` block using `hasUserMeshSolver()`? Everything could be the same as the `elasticity` solver case but we can print `_displacement_x/y/z` instead of `mesh_velocity_x/y/z`, so that the user sees what is congruent with the `user` mesh solver.

The user mesh solver has limited utility (mainly debugging) so it's not very important to me.

aprilnovak: Yes, I think we should - but I guess the code did not have it before, because I didn't see it here. Maybe it is pending merge in another open PR?

anshchaube: I don't see it in my other PR. It is possible it was overlooked. If it sounds okay, I can create an issue and approve this PR. We can circle back to the issue soon.

anshchaube: Sorry, my bad, we don't write displacements to `usrwrk`, we add them directly to the mesh.

aprilnovak: Thanks @anshchaube!

PR #636, opened 2023-03-24
One pebble CHT test case

jma6166:


aprilnovak:
I don't think this solution is correct, actually. If you look at the output files, the Nek mesh mirror surface is still the pebble surface (the fluid-solid interface), when it should be the pseudo-surface within the solid region.

aprilnovak:
Closing because of inactivity.

PR #638, opened 2023-03-26
Skin by density in OpenMC

aprilnovak:
This adds density skinning to `OpenMCCellAverageProblem`, identical in implementation to the temperature skinning, but now which re-inits the OpenMC materials to append `_<density bin>` so that we can change their density.

## How did I test this
skinning

The regression test does a 1:1 comparison between the skinned DAGMC model and an equivalent CSG version. For example, in the image above, an initial OpenMC model with 2 cells is re-skinned into 8 new cells at different densities. I built an identical CSG version of this same 8-cell geometry (with the same density values in each cell as in the skinned model). I was able to match exactly the values for mesh tally bins and k-eigenvalue, which proves equivalent behavior.


aprilnovak:
@pshriwise this is ready for review (finally) - I discovered when trying to apply this to the MSR model, that I had overlooked the scenario where there's just a single density bin, but you still need to create N materials to fill each new iso-temperature cell. So I added a new option in Cardinal that will essentially allow you to bypass some error checks we had where we'd abort if the same material was found in > 1 fluid cell (for our CSG geometries, this was probably a mistake -- but not the case for the DAGMC models).

pshriwise:
> @pshriwise this is ready for review (finally) - I discovered when trying to apply this to the MSR model, that I had overlooked the scenario where there's just a single density bin, but you still need to create N materials to fill each new iso-temperature cell. So I added a new option in Cardinal that will essentially allow you to bypass some error checks we had where we'd abort if the same material was found in > 1 fluid cell (for our CSG geometries, this was probably a mistake -- but not the case for the DAGMC models).

Ah that's tricky, yeah. If we wanted to get really fancy we could check the universe in which the repeated material occurs. An easier version would be to check if the model is still composed entirely of CSG and still enforce the check. That can be left for a later PR though IMO.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -3116,4 +3134,50 @@ OpenMCCellAverageProblem::reloadDAGMC()
...
+#ifdef ENABLE_DAGMC
+  // We currently only re-init the materials one time, because we create one new
+  // material for every density bin, even if that density bin doesn't actually
+  // appear in the problem. TODO: we could probably reduce memory usage
+  // if we only re-generated materials we strictly needed for the model.
+  if (!_first_transfer)
+    return;
+
+  // only need to create new materials if we have density skinning
+  if (_skinner->nDensityBins() == 1)
```
pshriwise, line 3149: Will the skinner try to assign a material name like `water_0` even with one density bin?

aprilnovak: No, the materials will not be changed if there is a single density bin (we `return` just after this line, in effect not modifying anything about the names). We _could_ change the names though if we wanted - I just thought it made more sense to skip that if density skinning was effectively not doing anything.

include/base/OpenMCCellAverageProblem.h
```diff
@@ -839,7 +845,7 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
    * Whether to map density according to each individual OpenMC cell (in which case an
    * error is thrown if you don't have a unique material in each cell) or by material.
    */
-  const bool & _map_density_by_cell;
+  bool _map_density_by_cell;
```
pshriwise, line 848: Now that we're making our own copy of this variable, it's going to be out of sync with the one we get using the MOOSE input parsing system. I don't generally think this will be a problem based on the changes below, but is there a way to programmatically set the value of that variable inside the MOOSE parameters system instead?

pshriwise: And if we don't feel like it's worth the trouble, this comment thread can be a placeholder/reminder in case it does become a problem in the future.

src/base/OpenMCCellAverageProblem.C
pshriwise: I think that's okay for now. Just making a mental note. Thanks!

src/base/OpenMCCellAverageProblem.C
```diff
@@ -469,6 +473,19 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
...
+    // it is impossible for us to know with 100% certainty a priori how many materials
+    // we would need to create. It's also true that even if we had a single density
+    // bin, that we'd still need to create one material for every new temperature
+    // cell. So, we require the user to have map_density_by_cell to false if there
+    // is going to be any density feedback at all
+    if (_map_density_by_cell && _has_fluid_blocks)
+    {
+      _map_density_by_cell = false;
+      mooseWarning("When skinning by density, 'map_density_by_cell' is overridden to be false!\n"
+        "You can silence this warning by setting 'map_density_by_cell' to false.");
```
pshriwise, line 487: Just wanted to make it abundantly clear to the user that densities are still incorporated into feedback here.

```suggestion
      mooseWarning("When skinning by density, material densities are fed back as part\n"
        "of the skinning process and 'map_density_by_cell' is overridden to be false!\n"
        "You can silence this warning by setting 'map_density_by_cell' to false.");
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -652,14 +671,10 @@ OpenMCCellAverageProblem::initialSetup()
...
 
-    // If the density bins are > 1, we need to re-init the OpenMC materials once (between
-    // the first time we read the materials.xml and when we run the first skinned OpenMC
-    // geometry) so that we have unique materials that can be set to individual densities.
-    // This is just not yet implemented. I think it'd be worthwhile to just allow OpenMC
-    // to set densities of cells filled by the same material, i.e. have the "density"
-    // construct attached to the cell, not material.
-    if (_skinner->nDensityBins() > 1)
-      paramError("skinner", "Density binning is not currently supported for the OpenMC wrapping!");
+    if (_skinner->hasDensitySkinning() != _has_fluid_blocks)
```
pshriwise, line 674: Can you elaborate on the "vice versa" piece of this message for my benefit?

src/base/OpenMCCellAverageProblem.C
pshriwise: > Material densities are sent back regardless of the `map_density_by_cell` setting...

This is what I was trying to make extra clear in the warning message. Has my addition muddied the water?

I'm just thinking that a new user still working their way through the docs might have a problem with their run, turn on this setting, get a warning and be confused.

include/base/OpenMCCellAverageProblem.h
pshriwise: > I don't believe that we can actually change the value in the MOOSE parameters system

This is the piece of info I was missing. Didn't know we couldn't change those! Makes sense to take ownership of the value here in that case, though it is surprising we can't change those values through some call into the MOOSE params.

src/base/OpenMCCellAverageProblem.C
pshriwise: I guess this isn't quite clear to me (though maybe it should be). I know this will only appear if the skinning method is being used, but it reads to me as though you must use the skinner to get fluid feedback in fluid blocks w/ Cardinal, which isn't the case right?

include/base/OpenMCCellAverageProblem.h
aprilnovak: Maybe I don't fully understand what you're proposing? When fetching from the input file, parameters are grabbed as `const &`, so there's no way to change the values if they're const.

src/base/OpenMCCellAverageProblem.C
aprilnovak: This is saying that, if the user specified that their whole OpenMC model should only receive temperature feedback (and not density feedback), that we should throw an error if they are trying to apply density skinning. It is incongruous with the settings they specified. In one place in the input file, they are saying they want to apply density feedback, but in another place they are saying they don't want to apply density feedback.

src/base/OpenMCCellAverageProblem.C
pshriwise: > Your addition isn't really explaining why the warning is being thrown - what you have written is true, but it's not why the warning is being thrown.

I don't disagree with that, but it does seem to provide some helpful context as to how the simulation is going to behave. Maybe it's getting in the way of the point though?

I'd prefer (i). I think displaying a warning makes perfect sense in this case -- just wanted to give the user some assurance that density is still getting back into the OpenMC model.

pshriwise: If you feel this isn't a useful change, we can take or leave it. No big.

src/base/OpenMCCellAverageProblem.C
pshriwise: Okay, thanks for clarifying! Same page now.

src/base/OpenMCCellAverageProblem.C
pshriwise: To make sure I understand correctly, `elem` would be the equivalent of `map_density_by_cell` being `False` and `cell` is equivalent to `True`?

In the context to which that setting applies (mapping onto a CSG/DAMC geometry without skinning), the Boolean option makes more sense to me after reading the description of this setting more carefully in the documentation.
I wonder if it would be most accurate to say in the warning that this setting is ignored regardless of its value when a skinner is present.

Thanks for the thorough consideration of this!!

aprilnovak: That's correct.

`map_density_by_cell` is not ever ignored - the valid user options just depend on whether they have skinning or not.

- **No skinning (for CSG or DAGMC):** density can be sent from MOOSE to OpenMC by either (i) trying to set a unique density in every cell (`map_density_by_cell = true` == `map_density_by = cell`) or (ii) trying to set a unique density only on each unique material, which could technically be filling multiple cells (`map_density_by_cell = false` == `map_density_by = material`)
- **Skinning:** user MUST send density according to material (`map_density_by_cell = false` == `map_density_by = material`). I am just trying to save some headache by overriding their setting if they tried `map_density_by_cell = true`.

I'll defer to you on which you think is clearer - the boolean vs. the enumeration. What do you think?

pshriwise: I guess I'm still a bit confused. Isn't the nature of the temperature and density feedback entirely determined by the properties of the skinner object if one is used?

(really feeling the need for a term other than skinner today hah...)

aprilnovak: Yeah this is kind of in the weeds. With skinning, it is possible for the same material to fill multiple fluid cells, if the skinner ends up making non-contiguous cells. i.e. in the figure below, there are four light blue cells, each filled with the same material. `map_density_by = material` is going to set the density of that _single_ material by averaging over the elements which map to all four cells together.

Screen Shot 2023-04-30 at 1 07 23 PM

Otherwise, with `map_density_by = cell`, Cardinal is going to throw an error because the user put the same material in multiple cells, preventing density feedback on a cell-wise level.

pshriwise: And in the later case we're overriding that value when a skinner is present, right?

pshriwise: > I just thought it'd be more user-friendly to do that automatically

Totally. I think I'd still describe this parameter as being ignored with the presence of a skinner. So if it's present in the input file I think a warning like "`map_density_by_cell` is present in the input but will be ignored. Density mapping will be governed by the model skinning parameters." 

Sorry for all the nitpicking. There are just a lot of capabilities now and I'm hoping to make things as clear as possible for the growing user base.

aprilnovak: Ok I like that - that's clearer than what I was trying to convey.

Thanks for the detailed review! You're absolutely right, the code is getting more and more complex and be clear in documentation is important.

aprilnovak: Ok, change has been implemented - let me know if I translated this convo correctly.

pshriwise: Looks great!! 🙌🏻  Thanks again for the discussion

pshriwise (approved):
Thanks @aprilnovak! I'd say this is good to merge once tests pass.

aprilnovak:
Thanks so much for the careful review!

PR #684, opened 2023-05-12
WIP: Update NekRS version

aprilnovak:
- I am not sure why, but the `wait_for_first_app_init` does not fix the two-channel case (but still works for the stochastic runs). This is okay, as it is only a slight regression in user-friendliness. But it would be good to understand if it was just a fluke why `wait_for_first_app_init` worked for about a week.


aprilnovak:
Closing because this has been addressed in #745.

PR #691, opened 2023-05-16
Simplify separate domain tests

aprilnovak:
These tests break in an upcoming NekRS release, but the diffs are just due to a slightly different NekRS solve itself - the actual functionality we are trying to test (with Cardinal correctly sending different postprocessors to/from NekRS) is still done correctly, we were just diffing on the actual values of pressure drop.


ahuxford (approved):
All good with me.

PR #700, opened 2023-05-19
Volume operations over solid domain

aprilnovak:
This adds support for postprocessing operations on the solid parts of the NekRS domain (this is a revival of #532).


aprilnovak:
Cannot merge until we update the NekRS version, which has a critical bug fix for this to work in parallel.

PR #702, opened 2023-05-19
Update mesh-velocity-areas test

aprilnovak:
This updates the `deformation/mesh-velocity-areas` test to make us more robust to future NekRS changes.

- Use a different MOOSE transfer, which is not as susceptible to `MultiAppNearestNodeTransfer` to parallel problems when one node is equidistant to > 1 nearest node candidate
- Change the criteria we monitor in the test to the _difference_ in surface areas. For absolute area values, the diffs are very small and could be missed depending on criteria for a CSVDiff test.
- Changed the output file writing so that we only write the mesh areas on the last time step, because the diffs are exceptionally small at earlier time steps due to the transient nature of the deformation.


aprilnovak:
Closing, as these commits were cherry-picked on #745.

PR #734, opened 2023-06-01
Simplify tutorial options

aprilnovak:
The `face_centered` option should not be used in these two tutorials - that's an artifact of how we originally made these files through the Center of Excellence.

### DIFF
--- a/tutorials/gas_compact/unit_cell.py
+++ b/tutorials/gas_compact/unit_cell.py
@@ -47,6 +47,11 @@ matrix_density = 1700                    # graphite matrix density (kg/m3)
 
 def coolant_temp(t_in, t_out, l, z):
     """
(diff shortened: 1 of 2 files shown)


loganharbour:
Are those failures expected?

aprilnovak:
Was the environment on CIVET changed?

loganharbour:
Yep. This is in the rocky 9 container that you were using; completely new environment

aprilnovak:
The failing tests are _super_ simple. They are `--mesh-only`-type tests. If I comment out that `--mesh-only`, the tests finish without a stack trace (though the test themselves still diff because I didn't make the `input_in.e` files). 

So it seems like something with `--mesh-only` here. But I can't think of what it could possibly be.

aprilnovak:
Merging as-is because the two Nek failures are due to the new GCC 9 CIVET environment, independent of this PR's content.

PR #738, opened 2023-06-14
Allow initial shift in scratch space slots used by Cardinal

aprilnovak:
This allows an initial offset in the scratch space written by Cardinal so that other NekRS activities (such as velocity recycling functions) can use earlier slots in the scratch space.


aprilnovak:
Merging as-is, because the test failures are just due to an older environment on CIVET.

PR #745, opened 2023-07-09
Update NekRS version to v23

aprilnovak:
Explanation of changes to test suite:

- the `cht/nondimensional` test was previously using a pressure postprocessor in the CSV gold file. In this pseudo-steady case, pressure is not a reliable metric because it will be very sensitive to random numerical things during the first few time steps. Removed pressure as a criterion (adequate coverage is provided by other, less susceptible quantities).
- Cherry-picked the changes to `mesh-velocity-areas` test from #702.
- Remove the derived BC versions of the deformation standalone test, since they were not adding any coverage beyond existing tests.
- `nek_separatedomain/transfers_temperature` was re-golded due to very small diffs in the simulation itself.
- Unskipped the `nek_standalone/channel` test and re-golded to make test more meaningful coverage.
- `nek_standalone/conj_ht` was changed to only look at the last time step, to circumvent small diffs in the transient. Also remove min_T as a comparison parameter, because it is essentially zero and prone to large relative diffs. Coverage is still ensured in the volume extrema using max_T.


aprilnovak:
Merging as-is. Stochastic OpenMPI bug is causing statepoint issues with OpenMC.

PR #749, opened 2023-07-14
Modify material compositions from UserObject

### LINKED ISSUE #746: Add accessor to send material compositions into OpenMC C++ end
For coupling to depletion calculations, we want to send material compositions from the Python depletion solver into the C++ end to update them for the next multiphysics calculation.

aprilnovak:
This adds the capability to change OpenMC material compositions from a UserObject, which can be set from the input file. Alternatively, we can set the `_names` and `_densities_` members from some other piece of C++ code, as we see fit.

This is the functionality we will be using to send data FROM OpenMC's Python depletion module TO the C++ in Cardinal (i.e. this is half of the data transfers we need for the GAIN work). fyi @loganharbour the data is a `std::vector<std::string>` and `std::vector<double>` for each OpenMC material we will change. 

### How did I test this

I tested this by doing 1:1 comparisons against OpenMC standalone runs. For the Cardinal cases, all the OpenMC XML files start with some dummy material composition which we change via MOOSE (so I compared these against standalone OpenMC runs which already loaded those material compositions).

Closes #746


aprilnovak:
fyi @pshriwise ignore the test failure, this is ready for review.

pshriwise (approved):
No real comments from me here. Thanks @aprilnovak!

PR #764, opened 2023-08-09
compacts should use k_compacts

### LINKED ISSUE #763: Potential typo in Solid and Fluid Coupling Tutorial
I believe this [line](https://github.com/neams-th-coe/cardinal/blob/0e606acbcfccb3b8a1bd4c38523493a83e8e5f4e/tutorials/gas_assembly/solid.i#L87) from the `[Materials]` block in the Solid and Fluid Coupling tutorial may have a typo.  The relevant part of the  block looks like this 
```
[Materials]
  [graphite]
    type = HeatConductionMaterial
    thermal_conductivity_temperature_function = k_graphite
    temp = T
    block = 'graphite'
  []
  [compacts]
    type = HeatConductionMaterial
    thermal_conductivity_temperature_function = k_TRISO
    temp = T
    block = 'compacts'
  []
...
[]
```
Here are the relevant functions in this input file
```
[Functions]
  [k_graphite]
    type = ParsedFunction
    expression = '${matrix_k}'
  []
  [k_TRISO]
    type = ParsedFunction
    expression = '${kernel_fraction} * ${kernel_k} + ${buffer_fraction} * ${buffer_k} + ${fparse ipyc_fraction + opyc_fraction} * ${PyC_k} + ${sic_fraction} * ${SiC_k}'
  []
  [k_compacts]
    type = ParsedFunction
    expression = '${triso_pf} * k_TRISO + ${fparse 1.0 - triso_pf} * k_graphite'
    symbol_names = 'k_TRISO k_graphite'
    symbol_values = 'k_TRISO k_graphite'
  []
...
[]
```
There seems to be a function for computing the compact homogenized thermal conductivity, so I believe the line in the  Materials block should read
```
    thermal_conductivity_temperature_function = k_compacts
```

Feel free to close the issue if I am misinterpreting the above line. Otherwise, I can make PR to fix this.

lewisgross1296:
Closes #763.


aprilnovak:
Can you please also check the other TRISO tutorials and make the fix there + re-gold the tests

lewisgross1296:
I found a few other instances of `k_TRISO` likely needing to be `k_compacts`, so I've pushed what I believe is every case. This means that the following tutorials should be re-golded:
-  TRISO Compacts (`cardinal/tutorials/gas_compact`)
- Conjugate Heat Transfer for Turbulent Channel Flow (`cardinal/tutorials/gas_compact_cht`)
- Solid and Fluid Coupling of OpenMC and MOOSE (`cardinal/tutorials/gas_assembly`)

Let me know if you agree.

In order to re-gold, do I just re-run each simulation, copy the relevant outputs to `tutorial_name/gold` and push the new outputs? This is my first time golding something, so please let me know if there are other steps involved.

aprilnovak:
Yep, looks good! You will just need to re-run those cases and copy into the `gold` directories the new golded files. You can check the `tests` file in each directory to see what the names of those files are.

lewisgross1296 (commented):
@aprilnovak I noticed that when I tried to run the `[run_solid]` test using the given `cli_args`. I was getting negative temperatures. I ran 
```
~/cardinal/cardinal-opt -i solid.i Executioner/num_steps=1 Mesh/file=coarse_mesh.e Outputs/file_base=solid_test_out AuxVariables/power/initial_condition=1e8
```
from`cardinial/tutorials/gas_assembly`.

What I think is happening is that since there's no `initial_condition` for the `thm_temp` AuxVariable in `solid.i`, it defaults to zero and "loses heat," so the internal temp goes down.

When I run the main app (`openmc.i`), this issue does not occur. When I add `AuxVariables/thm_temp/initial_condition=598.0` to the command line args, there is no error, but the results in `solid_test_out.csv` are quite different. Were you able to run this test as standalone without providing an `initial_condition` for `thm_temp`?

tutorials/gas_compact/openmc.i
```diff
@@ -136,7 +136,7 @@ mdot = 0.011                             # fluid mass flowrate (kg/s)
 
 [Executioner]
   type = Transient
-  num_steps = 4
+  num_steps = 2
```
aprilnovak, line 139: Why was this changed? We should not be decreasing this.

lewisgross1296: My bad, I originally changed this so that the number of steps equaled the number of time steps for the gold files (2, see tests). I did this before I realized `num_steps` could be set from the command line. I can put it back

lewisgross1296:
I noticed there's no gold directory for `gas_compact_cht`. In that case, everything is ready for final review.

If you'd like me to create a gold directory, I can, but it seems like the only test is whether thm can run (and it did).
cht_csv_outputs
The full `solid_thm.i` simulation generated a number of csv outputs. Some of them seem similar to other golded files (see attached, but maybe `thm.csv` and `solid_thm.csv`?). I could add a `[test]` for them too.  (I didn't run `solid_nek.i` due to lack of quick HPC.)

If everything looks good to you, I'll rebase to shrink the number of commits, since I added quite a few here. Otherwise, when you get a chance, let me know what you'd like changed.

aprilnovak:
@lewisgross1296 this needs further updates before it is ready for review. Not all the tests are passing.

tutorials/gas_compact_multiphysics/gold/openmc_thm_test_out.csv
```diff
@@ -0,0 +1,4 @@
+time,flux_integral,heat_source,k,k_std_dev,max_power,max_tally_err,min_power
```
aprilnovak, line 1: This file isn't used anywhere - it doesn't show up in any `tests`. Why is it here? There is already a test which is covering the `openmc_thm.i` case.

tutorials/gas_compact_multiphysics/gold/openmc_thm_test_out_power_avg_0002.csv
```diff
@@ -0,0 +1,51 @@
+power_avg
```
aprilnovak, line 1: Same comment, not used anywhere.

tutorials/gas_compact_multiphysics/gold/solid_mesh_out.csv
```diff
@@ -1,3 +1,3 @@
 time,area_walls,volume_coolant,volume_fuel
 0,0,0,0
-1,0.080483429170927,0.00032169908772759,0.00040536598327801
```
aprilnovak, line 3: Please undo the changes to this file. This test was not covering anything related to the wrong thermal conductivity, this is just a mesh test and these diffs are just floating point noise on your machine.

tutorials/gas_compact_multiphysics/gold/openmc_thm_test_out.csv
lewisgross1296: My bad, I mistook the `Outputs/csv/file_base=openmc_thm_test_out` from `[openmc_thm_moose]`. I will rebase this commit out of the history, since all three issues were introduced by the same commit.

tutorials/gas_compact_multiphysics/gold/openmc_thm_test_out_power_avg_0002.csv
lewisgross1296: Will fix this one too.

tutorials/gas_compact_multiphysics/gold/solid_mesh_out.csv
lewisgross1296: I'll fix this too. Assuming the random noise is below the test threshold, which is why the test passes with this change.

lewisgross1296 (commented):
Removing these issues.

PR #768, opened 2023-08-09
allow SAM+THM compilation

ahuxford:
SAM removed conflicts and allows for compilation with THM now. Removes Cardinal's objection to SAM+THM compilation.


aprilnovak:
Thanks for the contribution!

PR #790, opened 2023-09-20
v23 picard changes

regerdavid:
Implemented changes to enable picard iterations of NekRS. These can be enabled by setting fixed_point_iterations=true in the NekRSProblem block. A postprocessor named "fp_iteration" must exist in the nek.i file that will receive the current iteration number from the main app.

### DIFF
--- a/src/base/NekRSProblem.C
+++ b/src/base/NekRSProblem.C
@@ -70,6 +70,8 @@ NekRSProblem::validParams()
     "vector postprocessors, and places restrictions on how the sidesets are set up.");
   params.addParam<Real>(
       "initial_mesh_vel", 0.0, "Initial mesh velocity to pass to NekRS on the first timestep");
+  params.addParam<bool>("calculate_filtered_velocity", false,
+    "Whether to conserve determine the filtered velocity for FSI calculations. This can be useful if there is significant mesh compression in the solid.");
   return params;
 }
 
@@ -79,7 +81,8 @@ NekRSProblem::NekRSProblem(const InputParameters & params)
     _conserve_flux_by_sideset(getParam<bool>("conserve_flux_by_sideset")),
     _abs_tol(getParam<Real>("normalization_abs_tol")),
     _rel_tol(getParam<Real>("normalization_rel_tol")),
-    _initial_mesh_vel(getParam<Real>("initial_mesh_vel"))
+    _initial_mesh_vel(getParam<Real>("initial_mesh_vel")),
+    _calc_filtered_velocity(getParam<bool>("calculate_filtered_velocity"))
 {
   nekrs::setAbsoluteTol(getParam<Real>("normalization_abs_tol"));
   nekrs::setRelativeTol(getParam<Real>("normalization_rel_tol"));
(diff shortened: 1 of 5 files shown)


src/base/NekRSProblem.C
```diff
@@ -109,9 +112,15 @@ NekRSProblem::NekRSProblem(const InputParameters & params)
     indices.mesh_velocity_x = start++ * nekrs::scalarFieldOffset();
     indices.mesh_velocity_y = start++ * nekrs::scalarFieldOffset();
     indices.mesh_velocity_z = start++ * nekrs::scalarFieldOffset();
+    indices.filtered_velocity_x = start++ * nekrs::scalarFieldOffset();
```
aprilnovak, line 115: Hi David, would you please elaborate upon what these filtered velocity components are? Would someone want both the `mesh_velocity_x` and `filtered_velocity_x`, or does this instead represent some option we should have for computing `mesh_velocity_x`?

regerdavid: Sure, from my understanding these are important if there is any compression in the solid mesh, as with an incompressible flow solver this will cause huge instabilities. The filtered velocity calculates the integral of mesh_v dotted with the surface normal, and then subtracts this from mesh_v (so filtered_v = mesh_v - integral(mesh_v dot n). We apply the filtered v to the velocity boundary condition in Nek and then apply the unfiltered mesh velocity to the mesh boundary condition in Nek. For the cases I've run so far, this the integral term is really small (like 1e-12 or something), but Dr. Merzari has said that it is important for other FSI problems that they plan on doing.

aprilnovak: Thanks! I'm going to add a commit on top of this branch then, some changes need to happen in order to pass the existing test suite.

What do you think of making the filtered velocity computation an optional setting, which maybe defaults to off? This will reduce the amount of memory we need to allocate in the scratch space all the time, and may be useful for FSI problems to test turning it on/off.

regerdavid: Sure, I can make it optional!

aprilnovak: Thanks! Once you do that, I'll add a commit on top of yours. Please ping me once the option has been added.

src/base/NekRSProblem.C
```diff
@@ -68,6 +68,13 @@ NekRSProblem::validParams()
     "Whether to conserve the heat flux by individual sideset (as opposed to lumping all sidesets "
     "together). Setting this option to true requires syntax changes in the input file to use "
     "vector postprocessors, and places restrictions on how the sidesets are set up.");
+  params.addParam<Real>(
+      "initial_mesh_vel", 0.0, "Initial mesh velocity to pass to NekRS on the first timestep");
```
aprilnovak, line 72: Is there ever a situation where you would set this to something other than zero? Also I am a bit confused on the dimensions here. Is this interpreted as `velocity * hat(n)`?

regerdavid: I had added this in at one point for testing purposes, it doesn't get dotted with the normal, no, it just applies to all fields. I can just remove it for this pull request actually

aprilnovak: Yes, I think let's remove it for simplicity. We can always re-add it if necessary

aprilnovak: You'll need to remove it as an option here, too - otherwise people may think this does something, when it does not.

PR #801, opened 2023-10-12
v23 picard changes

aprilnovak:
Same PR as #790, but I did not want to overwrite @regerdavid's branch to split out into 2 PRs.

### DIFF
--- a/src/base/NekRSProblemBase.C
+++ b/src/base/NekRSProblemBase.C
@@ -109,8 +109,12 @@ NekRSProblemBase::validParams()
     "is mapped to/receives data from the mesh mirror for every time step.");
   params.addParam<unsigned int>("constant_interval", 1,
     "Constant interval (in units of number of time steps) with which to synchronize the NekRS solution");
-  params.addParam<bool>("fixed_point_iterations",false,"Whether or not fixed point iterations will be used. Cardinal will look for a Postprocessor called "
-  " 'fp_iteration' that receives the current iteration number from the top-level application.");
+  params.addParam<bool>(
+      "fixed_point_iterations",
+      false,
+      "Whether or not fixed point iterations will be used. Cardinal will look for a Postprocessor "
+      "called "
+      " 'fp_iteration' that receives the current iteration number from the top-level application.");
   return params;
 }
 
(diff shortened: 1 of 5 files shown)


src/base/NekRSProblemBase.C
```diff
@@ -629,23 +636,39 @@ NekRSProblemBase::externalSolve()
...
+  if (_fp_iteration)
   {
-    converged = nekrs::runStep(corrector++);
-  } while (!converged);
-
-  // TODO: time is somehow corrected here
-  nekrs::finishStep();
+    const PostprocessorValue * iter = &getPostprocessorValueByName("fp_iteration");
+    std::cout << "Current fixed point iteration number read in NekRSProblemBase is " << *iter
+              << std::endl; // JUST FOR TESTING
```
anshchaube, line 645: Is this being kept in the final code? Comment indicates otherwise

src/base/NekRSProblemBase.C
```diff
@@ -629,23 +636,39 @@ NekRSProblemBase::externalSolve()
...
+    nekrs::initStep(_timestepper->nondimensionalDT(step_start_time),
+                    _timestepper->nondimensionalDT(_dt),
+                    _t_step);
+    int corrector = 1;
+    bool converged = false;
+    do
+    {
+      converged = nekrs::runStep(corrector++);
+    } while (!converged);
+    // TODO: time is somehow corrected here
```
anshchaube, line 669: Has `TODO` item been accomplished i.e. the time correction? @regerdavid

src/base/NekRSProblemBase.C
regerdavid: I think this should be kept in the final code. Depending on how the postprocessor for the iteration is passed into nek.i, it can be possible to mess it up (if the correct execute_on flags are not used), so I think this is necessary for the user to check that the correct iteration count is being received.

src/base/NekRSProblemBase.C
regerdavid: I don't believe that that is my TODO comment (looks like it exists in the current master branch of cardinal) so I'm not sure if that is resolved or not.

src/base/NekRSProblemBase.C
anshchaube, line 645:
```suggestion
              << std::endl;
```

anshchaube: Maybe the comment should be removed, otherwise everything looks good to me, thanks @regerdavid @aprilnovak

regerdavid:
As an update, further testing is needed to make sure that the fixed point iterations are working correctly for non-FSI applications. I was trying to use them with a CHT case today and was not getting expected behavior.

aprilnovak:
This is finally ready to consider for merging, now that the CHT bug has been fixed in NekRS's `next` (which I updated last week). However, I'd like to see at least one test for these changes to make sure we maintain this functionality. @regerdavid is this something you/another student are still working on? What would you like to see happen with this pull request?

anshchaube:
I think I have found a workaround for the `execute_on` flags, but some issues still remain (with the number-of-iteration postprocessor transfer lagging at the first fixed point iteration,and some issues with subcycling and data-transfers). The path forward for that would require attempting to reproduce these issues without NekRS. I can take a look at this in the coming weeks.

PR #816, opened 2023-11-22
Remove diff in `tutorials/gas_contact:gas_contact`

### LINKED ISSUE #815: Do not use csvdiff `tutorials/gas_contact:gas_contact`
## Reason
The diff in `tutorials/gas_contact:gas_contact` is not particularly useful because we test it with so few histories. The purpose of the test is really to ensure that the syntax works, and checking the result is not appropriate.

## Design
Make the `tutorials/gas_contact:gas_contact` test a `RunApp` test.

## Impact
Minor because the diff means little.

loganharbour:
Closes #815

This will enable using the new container as well.


loganharbour:
@aprilnovak this is good to go

PR #817, opened 2023-12-01
Update moose submodule and use common container

loganharbour:
We should merge https://github.com/idaholab/civet_recipes/pull/1580 before merging this so that the new container is used on devel.


loganharbour:
Once this in, we can start testing cardinal regularly on moose! Still need to figure out how to skip some of the nek tests, but we're getting there.

PR #835, opened 2024-02-07
Allow independent trigger settings for individual tally scores

aprilnovak:
This allows individual triggers to be applied to the tally scores, if you want to fine tune trigger settings.


aprilnovak:
fyi @meltawila

PR #844, opened 2024-02-19
Fix tritium score for TallyRelativeError

aprilnovak:
When we recently added the enumeration to the `TallyRelativeError`, I didn't notice that `MultiMooseEnum` get input as all caps, but `MooseEnum` are lower case. So we had the conversion wrong going from the tritium score `H3_production` because the logic was still based on the assumption the incoming field was all caps.

Also added some deprecated syntax updates in the tutorials, which are unrelated to the error fix.


aprilnovak:
@meltawila fyi

PR #847, opened 2024-02-19
Update cht1.md

MendesMichael:
Adding instructions on how to run .jou files in Cubit.


aprilnovak:
Looks great! Can you also check to see if there are any other tutorials which use `.jou` files? I think there are a few. You can find all the files by doing

```
cd tutorials
grep -rI "*.jou" .
```

and then figure out what `.md` file corresponds to those tutorials.

MendesMichael:
I checked, and the only time where a .jou file is mentioned is in the reflector bypass flow example.

aprilnovak:
Great work for first PR!

PR #877, opened 2024-05-03
sawtooth docs extra info and launch script adjustment

### LINKED ISSUE #876: Enhance Sawtooth entry on HPC docs
## Reason
The main reason is for performance. After discussing with some INL HPC users, it seems beneficial to include the environment variable `OMP_PROC_BIND=true` in every launch script. I'd also like to add a little more description about running jobs on Sawtooth regarding the hardware resources and advice to leverage them as best as possible.

## Design
Very minimal addition to sawtooth submit script [here](https://cardinal.cels.anl.gov/hpc.html) and some extra description about using it.

## Impact
New users will get better performing jobs by default and have a little more guidance on using Sawtooth

lewisgross1296:
Closes #876. Open to feedback on the docs comments/submit script.

Also, is there a reason we load `cmake` in the job launch script? Does the cardinal executable need `cmake` to execute the job? I'd think no, but I am not an expert in that department

### DIFF
--- a/doc/content/hpc.md
+++ b/doc/content/hpc.md
@@ -193,6 +193,12 @@ export PYTHONPATH=$MOOSE_DIR/python:${PYTHONPATH}
 
 [Sawtooth](https://nsuf.inl.gov/Page/computing_resources)
  is an [!ac](HPC) system at [!ac](INL) with 99,792 cores (48 cores per node).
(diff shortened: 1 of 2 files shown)


aprilnovak:
I think you're right, that `cmake` is not needed -- can you do a quick confirmation by removing it and relaunching your job?

lewisgross1296:
Works without cmake!

PR #880, opened 2024-05-08
Fix Sawtooth script

aprilnovak:
FYI @lewisgross1296 actually cmake is needed for Nek. Also, you cannot have hashtags on PBS lines, they are not interpreted as comments.


lewisgross1296:
Thanks for the catch! I must have added the comment after running, my bad.

PR #894, opened 2024-05-30
Enforce collision estimator for photon heating.

aprilnovak:
Thanks Shikhar for noting this incompatibility! Photon heating scores cannot be used with tracklength estimators due to some approximations made for thick-target bremmstrahlung.


shikhar413:
Thanks for the quick patch @aprilnovak!

Issue #899, opened 2024-06-06
Add support to use moab skinner on moving meshes


meltawila:
When the mesh is moving currently the moab skinner doesn't use the displaced problem mesh to generate new DAGMC meshes. This will lead to openmc running on the non-deforming mesh. This capability (and with further development still needed) will allow Cardinal to run multiphysics deforming meshes problems.

aprilnovak:
Let's remember to test this moving mesh capability when we have `scale /= 1` in `OpenMCCellAverageProblem`. This will belong in a separate PR from #900, since it'll need the changes you're putting in `OpenMCCellAverageProblem`.

PR #900, opened 2024-06-06
Skin displaced mesh

meltawila:
Update Moab Skinner to use displaced meshes
(Ref #899)


src/userobjects/MoabSkinner.C
```diff
@@ -6,6 +6,9 @@
 #include "BinUtility.h"
 #include "GeometryUtils.h"
 #include "UserErrorChecking.h"
+#include "DisplacedProblem.h"
+#include "FEProblemBase.h"
+#include "MooseMesh.h"
```
aprilnovak, line 11: I think if you move this line into the header, you can kill two birds with one stone and eliminate the `class MooseMesh` thing.

src/userobjects/MoabSkinner.C
```diff
@@ -215,9 +222,25 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
   return _fe_problem.getAuxiliarySystem().getFieldVariable<Real>(0, name).number();
 }
 
+MooseMesh &
+MoabSkinner::getMooseMesh()
+{
+  if (_use_displaced)
```
aprilnovak, line 228: Can we auto-detect a good default for this? For instance, is `_fe_problem.getDisplacedProblem()` a null pointer if the mesh is not moving? If so, we can use an intelligent default so that users don't make a mistake and forget to enable this when they want it.

meltawila: I added that exactly in line 299 after I kept getting a seg fault while testing which I realized later was because I was `_fe_problem.getDisplacedProblem()->updateMesh();` a non-displaced problem. Do you mean I should do this everywhere as well and remove the `use_displaced_mesh = false` option? I added it this way because I found other solid mechanics module objects use it as an input parameter as well so just following the same logic

aprilnovak:
Can you remove the MOOSE submodule update you added? I don't think it's necessary. Trying to figure out what the sockeye thing is all about, since I've got another PR which is running the test suite fine...

src/userobjects/MoabSkinner.C
```diff
@@ -6,6 +6,8 @@
 #include "BinUtility.h"
 #include "GeometryUtils.h"
 #include "UserErrorChecking.h"
+#include "DisplacedProblem.h"
+#include "FEProblemBase.h"
```
aprilnovak, line 10: Hmm ok I think we may have a problem here -- this is going to cause some circular logic, because we're later wanting to `#include` the `MoabSkinner` inside of a class which derives from `FEProblemBase`. 

How was this working for you in your other branch where you are modifying the `OpenMCCellAverageProblem` as well?

src/userobjects/MoabSkinner.C
```diff
@@ -97,7 +101,8 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
     _output_full(getParam<bool>("output_full")),
     _scaling(1.0),
     _n_write(0),
-    _standalone(true)
+    _standalone(true),
+    _use_displaced(getParam<bool>("use_displaced_mesh"))
```
aprilnovak, line 105: it's customary to put parameters which fetch references before the ones which are just setting placeholder defaults

test/tests/userobjects/moab_skinner/expanding-cube/all_bins.i
```diff
@@ -0,0 +1,139 @@
+[Mesh]
+ [cube]
+    type = FileMeshGenerator
+    file = '../../../neutronics/meshes/tet_cube.e'
+ []
+ [id1]
```
aprilnovak, line 6: Is there a reason why the subdomains are being modified like this? Is this essential for the test? If not, let's remove it because it is extraneous.

test/tests/userobjects/moab_skinner/expanding-cube/all_bins.i
```diff
@@ -0,0 +1,139 @@
...
+    verbose = true
+    material_names = "mat mat"
+
+    temperature = temp
+    n_temperature_bins = 2
+    temperature_min = 300.0
+    temperature_max = 1000.0
+
+    output_skins = true
+    use_displaced_mesh = true
```
aprilnovak, line 127: You don't need to repeat this here, since you already have it in `GlobalParams`

test/tests/userobjects/moab_skinner/expanding-cube/all_bins.i
```diff
@@ -0,0 +1,139 @@
...
+  []
+[]
+
+[Executioner]
+  type = Transient
+  num_steps = 2
```
aprilnovak, line 133: `num_steps` is unused here, since the main app is controlling the time stepping. We can remove this

test/tests/userobjects/moab_skinner/expanding-cube/all_bins.i
```diff
@@ -0,0 +1,139 @@
...
+    type = SkinnedBins
+    variable = bins
+    skinner = moab
+  []
+[]
+
+[Functions]
+  [temp]
+    type = ParsedFunction
+    expression = 400+x*100+100*t
```
aprilnovak, line 107: It looks like as the mesh is deforming, that the deformation doesn't cause any change in the actual skinning (not that anything is wrong with your code, but I think we can design this test better). Can you modify the functions used for temperature and density in order to make something that will change how the bins are assigned based on the mesh movement? I'd remove the time dependence, and make the temperature dependence such that, as the mesh moves, the skins also change.

aprilnovak: You may just want to use the `temperature_min` and `temperature_max` to exactly enclose the min/max temperature so that you can see the bins move as the mesh moves.

src/userobjects/MoabSkinner.C
aprilnovak: After thinking about this more I think your design is fine -- let's use the same terminology as in the solid mechanics module for consistency.

test/tests/userobjects/moab_skinner/expanding-cube/constant_expansion_coeff.i
```diff
@@ -0,0 +1,167 @@
...
+  solve_type = 'PJFNK'
+
+  l_max_its = 50
+  nl_max_its = 50
+  nl_rel_tol = 1e-12
+  nl_abs_tol = 1e-10
+  l_tol = 1e-9
+
+  start_time = 0.0
+  end_time = 0.075
```
aprilnovak, line 104: If we end up only comparing the first 2 time steps in our test suite, let's remove the later time steps. Doing so will reduce the size of our gold files, helping to reduce repository bloat (and also lets the test remain hyper-focused on what we're trying to prove).

aprilnovak:
For future PRs, make sure to remove any of the deprecated `./` and `../` stuff from input files -- that will eventually no longer be supported, and we're trying to keep Cardinal with latest MOOSE syntax, especially with new file adds.

src/userobjects/MoabSkinner.C
```diff
@@ -215,9 +220,25 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+  {
+    return _fe_problem.getDisplacedProblem()->mesh();
+  }
+  else
+  {
+    return _fe_problem.mesh();
+  }
+}
 MeshBase &
 MoabSkinner::mesh()
```
aprilnovak, line 236: Are the `getMooseMesh()` and `mesh()` functions here doing basically the same thing? Can we consolidate them into a single function call instead?

src/userobjects/MoabSkinner.C
```diff
@@ -273,6 +294,10 @@ MoabSkinner::update()
   // Clear MOAB mesh data from last timestep
   reset();
 
+  if (_use_displaced && _fe_problem.getDisplacedProblem() != nullptr)
```
aprilnovak, line 297: Should we throw an error if the user writes `use_displaced = true` but `getDisplacedProblem` returns a null ptr? I think that could go in the constructor, and then we'd not have to check this nullptr here.

Issue #906, opened 2024-06-13
Add support to allow Cardinal to run OpenMC on deforming meshes problems 


meltawila:
Using moab skinner updates from Ref. #900 (described in Ref. #899) the displaced problem mesh is now used to generate new DAGMC meshes. Here we edit OpenMCCellAverageProblem to update the displaced problem mesh as well so we can run OpenMC with multiphysics deforming meshes problems

PR #907, opened 2024-06-13
OpenMCCellAverageProblem updates

meltawila:
Update OpenMCCellAverageProblem to use the displaced problem mesh 

(Ref. #906)

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -756,8 +756,8 @@ OpenMCCellAverageProblem::storeElementPhase()
   for (const auto & s : excl_density_blocks)
     _n_moose_density_elems += numElemsInSubdomain(s);
 
-  _n_moose_none_elems =
-      getMooseMesh().nElem() - _n_moose_temp_density_elems - _n_moose_temp_elems - _n_moose_density_elems;
+  _n_moose_none_elems = getMooseMesh().nElem() - _n_moose_temp_density_elems - _n_moose_temp_elems -
+                        _n_moose_density_elems;
 }
 
 void
@@ -1339,9 +1339,10 @@ OpenMCCellAverageProblem::initializeElementToCellMapping()
     mooseError("Did not find any overlap between MOOSE elements and OpenMC cells for "
                "the specified blocks!");
 
-  _console << "\nMapping between " + Moose::stringify(getMooseMesh().nElem()) + " MOOSE elements and " +
-                  Moose::stringify(_n_openmc_cells) + " OpenMC cells (on " +
-                  Moose::stringify(openmc::model::n_coord_levels) + " coordinate levels):"
+  _console << "\nMapping between " + Moose::stringify(getMooseMesh().nElem()) +
+                  " MOOSE elements and " + Moose::stringify(_n_openmc_cells) +
+                  " OpenMC cells (on " + Moose::stringify(openmc::model::n_coord_levels) +
+                  " coordinate levels):"
            << std::endl;
 
   VariadicTable<std::string, int, int, int, int> vt(
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -20,6 +20,7 @@
 
 #include "OpenMCCellAverageProblem.h"
 #include "DelimitedFileReader.h"
+#include "DisplacedProblem.h"
 #include "TallyBase.h"
 #include "CellTally.h"
 #include "AddTallyAction.h"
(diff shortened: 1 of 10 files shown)


src/base/OpenMCCellAverageProblem.C
```diff
@@ -3063,6 +3085,11 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
         if (_volume_calc)
           _volume_calc->resetVolumeCalculation();
 
+        if (_use_displaced)
```
aprilnovak, line 3088: let's figure out just the single place where this call needs to go -- we shouldn't need it more than once in this class.  I think this is the right place to have it. Can you try one-by-one progressively deleting, re-compiling, and re-running your test case in order to confirm we can have just this single call?

aprilnovak:
@meltawila can you please rebase this? There's been a lot of changes to `OpenMCCellAverageProblem` recently

meltawila:
@aprilnovak would you check this?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -164,7 +165,9 @@ OpenMCCellAverageProblem::validParams()
   params.addParam<int>("first_iteration_particles",
                        "Number of particles to use for first iteration "
                        "when using Dufek-Gudowski relaxation");
-
+  params.addParam<bool>("use_displaced_mesh",
```
aprilnovak, line 168: @nuclearkevin does this need to be modified at all to be consistent with what you did with `fixed_mesh` being auto-detected? 

@meltawila can you please also remind me, was there a use case where we wanted to disable the displaced mesh updates in OpenMC, even if the mesh is actually moving? Or was this user option only existing because we didn't know how to detect it automatically?

aprilnovak:
Thanks for updating the PR! We'll want to have a few tests before merging this. How about something like a CSG cylinder in OpenMC, and a cube mesh which translates around in space over and beyond the cylinder? We can visualize the `CellIDAux` to see the "shadow" of the CSG on the moving mesh.

src/base/OpenMCCellAverageProblem.C
meltawila: This is the logic used in MOOSE solid mechanics, to be able to set `use_displaced_mesh = false` for certain post processors/user objects only for example

nuclearkevin: @aprilnovak At the moment this doesn't need to be modified since the changes made to auto-detect `fixed_mesh` are coming in the AMR PR (#964). If this gets merged first, I can handle the necessary changes in #964. If the AMR PR gets merged ahead of this one, a single line of code will need to be added to this PR to get auto-detection working:
https://github.com/neams-th-coe/cardinal/pull/964/files#diff-7321f62ca7502f529d9917464fed4643c5e57c97c27d295f78db837d1a35a366R219

nuclearkevin: For reference @meltawila this is what's necessary to automatically detect if the mesh is moving and setup the `OpenMCCellAverageProblem` accordingly. This include is required:
```
#include "CreateDisplacedProblemAction.h"
```
and this needs to get added to the constructor of `OpenMCCellAverageProblem`:
```
// Check to see if a displaced problem is being initialized.
const auto & dis_actions = getMooseApp().actionWarehouse().getActions<CreateDisplacedProblemAction>();
for (const auto & act : dis_actions)
{
  auto has_displaced = act->isParamValid("displacements") && act->getParam<bool>("use_displaced_mesh");
  _need_to_reinit_coupling |= (has_displaced && _use_displaced_mesh);
}
```

meltawila:
We have `test/tests/userobjects/moab_skinner/expanding-cube` which I wanted to extend to use OpenMCCellAverageProblem as the test for this. I don't think CSG will work, right? either way we need to conserve the mass and update material densities with mesh movements as well

aprilnovak:
The mesh motion aspect doesn't necessarily need to be interpreted as the OpenMC geometry changing. Although it's sort of weird why anyone would intentionally do this, the test I proposed would just check that the cell <--> element mapping is happening correctly.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -368,6 +372,16 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
   }
 }
 
+const MooseMesh &
+OpenMCCellAverageProblem::getMooseMesh() const
+{
+  if (_use_displaced && !_displaced_problem && !_first_transfer)
```
aprilnovak, line 378: What is the motivation for having `_first_transfer` as part of this?

aprilnovak:
@meltawila please close this

PR #908, opened 2024-06-13
Add test to cover user-provided bounding box for volume calculations.

aprilnovak:
This adds a test that shows that the OpenMC volume calculation correctly uses a bounding box provided by the user. We did have an error-check test which would throw an error if the lower left was greater than the upper right, but this test provides a user-given bounding box of volume 0.001, and we show that the stochastic calculation by OpenMC gives the same value.


meltawila (approved):
Looks good to me

PR #913, opened 2024-06-20
Update OpenMC tutorials

nuclearkevin:
Updating the LWR pincell and solid UO<sub>2</sub> pebbles tutorials to fix errors that pop up when using the up-to-date OpenMC Python API. This PR also swaps all instances of `openmc.source.source` to `openmc.IndependentSource` and re-generates the settings.xml files to avoid any future deprecations in the OpenMC source API.


aprilnovak:
Thanks @nuclearkevin! In order for the test suite to launch, you'll have to accept the invitation I just sent to join `cardinal-devs` (you might need to force push to re-trigger the tests)

tutorials/gas_assembly/settings.xml
```diff
@@ -13,7 +13,7 @@
   <temperature_range>294.0 1500.0</temperature_range>
   <volume_calc>
     <domain_type>cell</domain_type>
-    <domain_ids>42120</domain_ids>
+    <domain_ids>42171</domain_ids>
```
aprilnovak, line 16: do you know what is causing this diff?

nuclearkevin: I have no idea, running assembly.py seems to change it every time. I believe it's caused by the specification of a packing fraction instead of a number of spheres in `openmc.model.pack_spheres`, so the uids of the domains change every time the model is regenerated.

aprilnovak: Can you try adding `seed=1` to that case in the `pack_spheres` function? IIRC, the behavior in OpenMC changed to make the pack_spheres call non-deterministic, but that wasn't the case when we first made this test.

nuclearkevin: I'll revert the volume calculation domain back to `42120` since this would likely cause the volume calculation to either fail or produce a volume estimate for the wrong cell.

nuclearkevin: Sure thing. I'll also add it to the other TRISO tutorials if that's the case there.

nuclearkevin: I'm also committing all of the new xml files (not just settings) to make sure everything matches. Was hoping to not need to do it since it results in a massive diff, but it is what it is...

Issue #850, opened 2024-02-25
average tally relative error


aprilnovak:
## Reason
The `TallyRelativeError` postprocessor currently computes either the max or min tally relative error. It would be useful to also compute the average.

## Design
Add `average` as an enumeration option, and compute the average relative error.

## Impact
New feature.

aprilnovak:
fyi @nuclearkevin, in the future you can write `Closes #850` instead of just `Refs #850` on a PR, and when it is merged it will automatically also close the issue.

PR #917, opened 2024-06-21
Averaging in `TallyRelativeError`

nuclearkevin:
Added averaging to `TallyRelativeError` and modified the fission_tally_relative_error tests to include averaging post-processors.
(Ref. #850)


src/postprocessors/TallyRelativeError.C
```diff
@@ -86,20 +86,24 @@ TallyRelativeError::getValue() const
...
+      post_processor_value = std::numeric_limits<Real>::max();
+      break;
+    case operation::average:
+      post_processor_value = 0.0;
       break;
     default:
       mooseError("Unhandled OperationEnum!");
   }
 
+  unsigned int num_values = 0u;
```
aprilnovak, line 106: I don't think you need this `0u`, but `0` would work fine. Is there a specific scenario you were thinking of by using `0u`, or just habit?

nuclearkevin: It's a habit of mine from other projects with styles that abuse `auto`. I can change it to `0` and try to break the habit for future PRs if that's preferred.

aprilnovak: I think convention in MOOSE is to use type-agnostic syntax as much as possible, so I'd prefer to remove it. I think you can also safely remove the `static_cast<Real>` line later as well, I think the type conversion should happen fine without it (but lmk if that's not the case).

nuclearkevin: Pushed style changes. Tests are still passing on my end so it looks like the `static_cast` was unnecessary.

PR #921, opened 2024-06-26
OpenMC depletion coupling

loganharbour:
Things still to do

- [ ] Actually verify the test, or setup actual verification?
- [ ] Also add a test with the openmc-only operator?
- [ ] Clean up the python imports in `depletion.py`
- [ ] Add more comments, maybe a little documentation
- [ ] Setup a proper clean exit of the `CardinalOperator` that checks to see if MOOSE is also done, otherwise error
- [ ] Clean up the state cleanup before running (hdf5 and tallies)

PR #923, opened 2024-07-08
Tally and Domain Filter Editor UOs

### LINKED ISSUE #837: Allow control of OpenMC depletion tallies
## Reason
In order for the MOOSE webserver depletion control to function properly, we need to be able to edit the tallies OpenMC automatically sets up for depletion. This includes:

- Pass the "possible" depleted nuclides (nuclides which may be created from depletion of a material) up from the depletion Python side, into Cardinal's C++ side. These will be used to then properly add those nuclides as filters on the reaction rate tallies needed for depletion.
- Create new tallies
- Change tally filters

## Design
Create a user object (or similar concept) to update the tallies and filters for OpenMC's depletion solver.

## Impact
New feature to support depletion coupling.

pshriwise:
This adds a `UserObject` for editing OpenMC Tally and some Filter objects. The order of execution in `OpenMCProblemBase` should ensure they can be used in conjunction to create custom, controllable tallies in a Cardinal input file (for the set of filters supported, see below).

OpenMCTallyEditor
---

This UO can be used to either edit an existing tally object or create a new one via the Cardinal input file. A tally's scores, nuclides, filters (set by ID) can all be specified in the input file. These parameters can also be controlled using the MOOSE control system. The tally's `multiply_density` attribute, governing whether or not tally results are multiplied by atom density, can be set as well, but as this is used less often it is set to `false` by default. `multiply_density` can also be controlled via the MOOSE control system. Other parameters are required.

If the `create_tally` parameter of an `OpenMCTallyEditor` object is `true` and a tally already exists with the specified ID, a warning is shown. If `create_tally` is `false` and a tally with the specified ID doesn't exist, an error is raised.

The tally UO provides functionality that can replace the `OpenMCTallyNuclides` and I've done so in this PR. I'm not sure that deprecation of the existing `OpenMCTallyNuclides` object is necessary given its short lifetime, but I'm happy to provide that here.

OpenMCDomainFilterEditor
---

The domain filter editor allows for the creation of tally filters and modification of their associated bins (by cell/material/universe/mesh ID). Like the tally properties above, filter bins can be set in the input file and modified between OpenMC executions as part of the MOOSE control system.

The supported filter types for this UO include: `CellFilter`, `MaterialFilter`, `UniverseFilter`, and `MeshFilter`. These are all spatial/domain filters that have the same generic bin structure and are important for depletion/activation workflows that may be attached to Cardinal multiphysics problems.

The filter editor UO is complicated by the necessity of a filter type. Additional checks have been added that ensure:

  - if `create_filter` is `true` and a tally filter with the specified ID already exists:
      - when `filter_type` parameter is set, the specified `filter_type` is checked against the existing `Filter`'s type and an error is raised on mismatch
      - if `filter_type` is not set or the types match, a warning is shown
  - if `create_filter` is `false` and a filter with a matching ID exists:
    - when `filter_type` parameter is set, the specified `filter_type` is checked against the existing `Filter`'s type and an error is raised on mismatch
      - if `filter_type` is not set, no additional check is performed other than to check that the filter is in the set of types supported by `OpenMCDomainFilter`

TODO:
- [x] Complete TallyEditor testing
- [x] Complete DomainFilterEditor testing
- [x] tally editor docs
- [x] domain filter editor docs

Closes #837


src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,204 @@
...
+
+registerMooseObject("CardinalApp", OpenMCDomainFilterEditor);
+
+InputParameters
+OpenMCDomainFilterEditor::validParams()
+{
+  InputParameters params = OpenMCUserObject::validParams();
+  params.addParam<bool>("create_filter", false, "Whether to create the tally if it doesn't exist");
+  params.addRequiredParam<int32_t>("filter_id", "The ID of the filter to modify");
+  params.addParam<std::string>("filter_type", "", "The type of filter to create");
```
aprilnovak, line 38: Let's make this a `MooseEnum` -- it will then render the options automatically in the documentation. If you haven't used that type before, you can follow examples in the `CardinalEnums.C`/`CardinalEnums.h` files and how MooseEnum's are used in `OpenMCCellAverageProblem`.

doc/content/source/userobjects/OpenMCDomainFilterEditor.md
```diff
@@ -0,0 +1,20 @@
+# OpenMCDomainFilterEditor
+
+!syntax description /UserObjects/OpenMCDomainFilterEditor
+
+## Description
+
+This user object can be used to create or control new domain filter in OpenMC.
+Currently suppored filter types are "cell", "material", "universe", or "mesh"
```
aprilnovak, line 8: In the interest of technical debt, I'd revise this sentence to "Currently supported filter types are specified with the 'filter_type' parameter."

include/userobjects/OpenMCDomainFilterEditor.h
```diff
@@ -0,0 +1,67 @@
...
+   */
+  void check_filter_type_match() const;
+
+  std::string long_name() const { return "OpenMCDomainFilterEditor \"" + this->name() + "\""; }
+
+  // Accessors
+  int32_t filter_id() const { return _filter_id; }
+
+protected:
+  bool _first_execution{true};
```
aprilnovak, line 63: MOOSE convention is to put these defaults in the member initialization list

include/userobjects/OpenMCDomainFilterEditor.h
```diff
@@ -0,0 +1,67 @@
...
+  int32_t _filter_id;
+  std::string _filter_type;
+  const std::set<std::string> _allowed_types{"cell", "universe", "material", "mesh"};
```
aprilnovak, line 66: Convention is to use a MooseEnum if possible

include/userobjects/OpenMCTallyEditor.h
```diff
@@ -0,0 +1,54 @@
...
+  void first_execution();
+  virtual void finalize() override {}
+
+  std::string long_name() const { return "OpenMCTallyEditor \"" + this->name() + "\""; }
+
+  // Accessors
+  int32_t tally_id() const { return _tally_id; }
+
+protected:
+  bool _first_execution{true};
```
aprilnovak, line 52: Would this be worth moving into the base class? I see it's in two of the new UOs

include/userobjects/OpenMCUserObject.h
```diff
@@ -20,29 +20,19 @@
 
 #include "GeneralUserObject.h"
 
+// forward declarations
+class OpenMCProblemBase;
+
 /**
- * User object to modify the nuclides in an OpenMC tally.
+ * User object to modify an OpenMC tally
```
aprilnovak, line 27: I think this documentation is not referring to this class? This looks like a generic base class

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,204 @@
...
+  : OpenMCUserObject(parameters),
+    _filter_id(getParam<int32_t>("filter_id")),
+    _filter_type(getParam<std::string>("filter_type"))
+{
+}
+
+void
+OpenMCDomainFilterEditor::initialize()
+{
+  bool create_filter = getParam<bool>("create_filter");
```
aprilnovak, line 55: Is there a reason this object isn't fetched from the member initialization list?

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,204 @@
...
+    for (const auto & type : _allowed_types)
+      msg += "\"" + type + "\"";
+    mooseError(msg);
+  }
+
+  // if the filter doesn't exist at this point and no other errors have been raised,
+  // create the filter
+  if (!filter_exists)
+  {
+    openmc_problem()->_console << long_name() << ": Creating Filter " << _filter_id << std::endl;
```
aprilnovak, line 104: why does this use `openmc_problem()->_console ` instead of just `_console`? I think they are the same?

doc/content/source/userobjects/OpenMCDomainFilterEditor.md
pshriwise: Good call! I'll use this.

include/userobjects/OpenMCUserObject.h
pshriwise: Good catch. Thanks!

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: It's used often in this method, so I assigned a variable to decrease code verbosity.

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: Didn't know that `_console` was available on the user objects. I'll use that instead!

nuclearkevin (commented):
After going over the changes, I don't think there are any clashes between these userobjects and the tally system when xml tallies/filters are modified. I also don't believe this will cause any issues with the filter system coming in #951 since those are thin wrappers the return values stored in the OpenMC tally objects.

I can see issues arising if the user attempts to modify tallies/filters added through the `[Problem/Tallies]` block since the changes made by these userobjects don't get propagated to either `CellTally` or `MeshTally` wrapper classes. It might be worth restricting the use of both userobjects so they only modify xml tallies or create new ones, unless you want to find a way to propagate changes in the filter/tallies to `CellTally` and `MeshTally`.

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,206 @@
...
+{
+  InputParameters params = OpenMCUserObject::validParams();
+  params.addParam<bool>("create_filter", false, "Whether to create the tally if it doesn't exist");
+  params.addRequiredParam<int32_t>("filter_id", "The ID of the filter to modify");
+  params.addParam<MooseEnum>("filter_type", getFilterTypeEnum(), "The type of filter to create");
+  params.addRequiredParam<std::vector<std::string>>("bins", "The bins to modify in the filter");
+  params.declareControllable("bins");
+  params.addClassDescription("A UserObject for creating and managing OpenMC domain tally filters");
+  return params;
+}
```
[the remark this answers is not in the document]
pshriwise: Very true @nuclearkevin. I wasn't intending these for use with mapped tallies. Unless there's a use case I'm not seeing, I don't think we'd want users controlling those given the amount of programmatic mapping we do in the `OpenMCCellAverageProblem`. For now, I think it's best if we assume these are limited to unmapped tallies.

I could also add some code to preclude their use for mapped Cardinal tallies.

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,206 @@
...
+int32_t
+OpenMCDomainFilterEditor::filter_index() const
+{
+  if (!filter_exists())
+    mooseError(long_name() + ": Filter " + std::to_string(_filter_id) + " does not exist");
+  return openmc::model::filter_map.at(_filter_id);
+}
+
+void
+OpenMCDomainFilterEditor::execute()
```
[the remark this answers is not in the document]
pshriwise: I'll check back on this depending on the discussion above

https://github.com/neams-th-coe/cardinal/pull/923#discussion_r1725657073

src/userobjects/OpenMCTallyEditor.C
```diff
@@ -0,0 +1,143 @@
...
+{
+  if  (!tally_exists()) {
+    mooseError(long_name() + ": Tally " + std::to_string(_tally_id) +
+               " does not exist in the OpenMC model");
+  }
+
+  openmc::Tally * tally = openmc::model::tallies[tally_index()].get();
+
+  std::vector<std::string> scores = getParam<std::vector<std::string>>("scores");
+  if (scores.size() > 0)
```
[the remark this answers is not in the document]
pshriwise: Action on this will depend on the discussion above

https://github.com/neams-th-coe/cardinal/pull/923#discussion_r1725657073

pshriwise:
@aprilnovak I'm seeing similar Nek-related test failures on other PRs and the changes here don't seem relevant (correct me if I'm wront) so I'm not going to dig into that right away.

aprilnovak:
You can ignore the container failures -- the Nek ones are all passing, the JIT seems to have gotten slower so there's just some timeouts but they are fine.

@loganharbour, anything special we should be doing about the container failures? I believe this is also related to the failure on #952

src/userobjects/OpenMCDomainFilterEditor.C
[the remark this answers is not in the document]
nuclearkevin: I also can't think of any reasons why `OpenMCTallyEditor`/`OpenMCDomainFilterEditor` should be able to modify mapped tallies/filters, so I'm good with restricting their use. It would probably be best to perform a programmatic check when constructing / initializing the userobjects (since they're added after tallies during initialization) to make sure the ids supplied by the users don't conflict with any of the mapped tally/filter ids just to be sure.

If the plan is to avoid the modification of mapped tallies the docs would need to be modified as they [explicitly state](https://github.com/neams-th-coe/cardinal/pull/923/files#diff-fa2f5618663f9bed68907812f88950bd0409609b945497c0c917f25d2f9a7fa2R394) that this is one of the use cases for these userobjects.

doc/content/source/problems/OpenMCCellAverageProblem.md
```diff
@@ -378,6 +384,15 @@ already are exactly the same volume, then no special thought is needed for the t
...
 used in OpenMC.
 
+## Tally and Filter Editors
+
+Cardinal provides UserObjects for editing tallies and filters.
+
+  - [OpenMCTallyEditor](https://cardinal.cels.anl.gov/source/userobjects/OpenMCTallyEditor.html)
+  - [OpenMCDomainFilterEditor](https://cardinal.cels.anl.gov/source/userobjects/OpenMCTallyEditor.html)
+
+These objects provide online control of tally and filter parameters, respectively. These objects can be used to interact with tallies and/or filters that are mapped in Cardinal as part of the `OpenMCCellAverageProblem` or are present in a `tallies.xml` file for the problem being run (unmapped tallies). These objects can be used to create tallies as well -- in this case the new objects will **not be mapped** to the mesh mirror.
```
nuclearkevin, line 394: I would recommend revising this section of the documentation to make it clear to the user that mapped tallies cannot be modified by either `OpenMCTallyEditor` or `OpenMCDomainFilterEditor`.

src/userobjects/OpenMCDomainFilterEditor.C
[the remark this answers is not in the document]
pshriwise: Agreed! I think that's a good plan. I'll implement a check that ensures the editor IDs and mapped IDs aren't overlapping. I'll look to the documentation too once I've done that. 👍🏻

pshriwise:
@nuclearkevin @aprilnovak I believe I've addressed all outstanding issues at this point. I've added additional checks and tests that the tallies with editors aren't tallies generated by Cardinal for data mapping. 

I started to add this for filters as well, but I realized that it isn't really possible for filters to overlap because the filters currently supported by the tally editors are currently in the set of XML filter types that are not allowed for application with Cardinal tallies, so I think we're okay there. I'm happy to add the code to double-check this if it would make everyone more comfortable, but it wouldn't be straightforward to test really. I'm thinking that it's something to note for later. (I do have a proposal on how we might be able to unify these features with the nice work Kevin has done on the current tally system as a follow-on).

@aprilnovak @loganharbour I'm not sure why CI isn't running on this PR. It seems it isn't happy with the SHA (not the one in the git history though I'll note). Thoughts?

aprilnovak:
Thanks @pshriwise! I'll give this another round of review today.

The CI isn't running because there's a merge conflict on the `OpenMCTallyNuclides.md` file. If you fix it and push up a change, the tests should launch.

nuclearkevin:
@pshriwise Thanks for addressing my concerns! I'll go over the changes again later today.

pshriwise:
> The CI isn't running because there's a merge conflict on the `OpenMCTallyNuclides.md` file. If you fix it and push up a change, the tests should launch.

Whoops! Thanks. Just addressed that. I'll keep an eye on CI now.

nuclearkevin (commented):
Requesting a few small changes to ensure that the PR meets MOOSE's C++ style guide. Otherwise it looks good to me!

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1487,6 +1488,8 @@ OpenMCCellAverageProblem::initializeElementToCellMapping()
 
   // Check that each cell maps to a single phase
   checkCellMappedPhase();
+
+  openmc_set_seed(1);
```
aprilnovak, line 1492: @pshriwise can you please explain this line to me? Would this preclude a user wanting to run Cardinal N times, with different starting seeds?

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,206 @@
...
+#include "openmc/tallies/filter_universe.h"
+#include "openmc/tallies/filter_mesh.h"
+
+registerMooseObject("CardinalApp", OpenMCDomainFilterEditor);
+
+InputParameters
+OpenMCDomainFilterEditor::validParams()
+{
+  InputParameters params = OpenMCUserObject::validParams();
+  params.addParam<bool>("create_filter", false, "Whether to create the tally if it doesn't exist");
```
aprilnovak, line 36: Should this line say "filter" instead of "tally"?

src/userobjects/OpenMCTallyEditor.C
```diff
@@ -0,0 +1,144 @@
...
+  this->initialize();
+}
+
+void
+OpenMCTallyEditor::initialize()
+{
+  // this is placed here, instead of the constructor, because Cardinal initializes
+  // some tallies. Depending on the order of initialization of UserObjects vs.
+  // other classes, those tallies might not exist yet in OpenMC's data space
+  // (but they will by the time this method is called).
```
aprilnovak, line 60: Does this comment refer to both of lines 61 and 62, or just line 62?

src/userobjects/OpenMCUserObject.C
```diff
@@ -0,0 +1,57 @@
...
+  {
+    std::string extra_help = _fe_problem.type() == "FEProblem" ? " (the default)" : "";
+    mooseError("This user object can only be used with wrapped OpenMC cases! "
+               "You need to change the\nproblem type from '" +
+               _fe_problem.type() + "'" + extra_help + " to OpenMCCellAverageProblem.");
+  }
+}
+
+void
+OpenMCUserObject::execute()
```
aprilnovak, line 44: Why is this method defined here? As far as I can tell, the only two derived classes of this base class are `OpenMCDomainFilterEditor` and `OpenMCTallyEditor`, both of which override `execute()` and do not call the parent class method at all. It seems like this `OpenMCUserObject::execute()` method never gets called.

Is this the case, or am I missing something?

test/tests/userobjects/openmc_tally_editor/tests
```diff
@@ -0,0 +1,132 @@
...
+  [nonexistent_nuclide]
+    type = RunException
+    input = tally_param_checks.i
+    cli_args = 'UserObjects/tally_editor_uo/nuclides="fake fake2"'
+    expect_err = "In attempting to set tally nuclides in the 'tally_editor_uo' UserObject, OpenMC reported:\n\n"
+                 "Nuclide 'fake' is not present in library."
+    requirement = "The system shall error if trying to add a nuclide not accessible in the cross section library."
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [nuclides]
```
aprilnovak, line 27: What is this test doing? The `CheckFiles` test type only looks to see if a file named `tallies.out` was created. But from the description of this test (and some others in this file), I think what you actually want is a `CSVDiff` or something which actually passes the test based on some numbers being correct? Is that the case? (if so, I can help to figure out how to change these tests)

test/tests/userobjects/openmc_tally_editor/tests
```diff
@@ -0,0 +1,132 @@
...
+    cli_args = 'UserObjects/new_filter/filter_type="universe"'
+    requirement = 'Ensure that filters specified by a tally editor UO are present in the tally output'
+    file_expect_out = 'Material'
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [invalid_filter_type]
+    type = RunException
+    input = add_filter.i
+    cli_args = 'UserObjects/new_filter/filter_type="invalid"'
+    expect_err = 'Invalid option "invalid" in MooseEnum.'
```
aprilnovak, line 90: We don't need to have this test; MOOSE handles error checking for invalid enums, now that we use the MooseEnum type we get this for free. I'll remove it in the commit I push.

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,206 @@
...
+  return params;
+}
+
+OpenMCDomainFilterEditor::OpenMCDomainFilterEditor(const InputParameters & parameters)
+  : OpenMCUserObject(parameters),
+    _filter_id(getParam<int32_t>("filter_id")),
+    _filter_type(getParam<MooseEnum>("filter_type").getEnum<OpenMCFilterType>())
+{
+
+  if (!parameters.isParamSetByUser("filter_type"))
```
aprilnovak, line 51: Is there a reason why we don't just have `filter_type` be required? Then we would not need to enforce this ourselves

```
params.addRequiredParam<MooseEnum>("filter_type", getFilterTypeEnum(), "The type of filter to create");
```

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,206 @@
...
+void
+OpenMCDomainFilterEditor::initialize()
+{
+  bool create_filter = getParam<bool>("create_filter");
+  bool filter_exists = this->filter_exists();
+
+  // if create_filter is set to true, but the filter already exists, display a warning
+  if (create_filter && filter_exists)
+  {
+    mooseWarning(long_name() + ": Filter " + std::to_string(_filter_id) +
```
aprilnovak, line 68: What is the intention for the user to understand when seeing this warning? I feel like this should be an error -- I guess the use case we are allowing is for a user to override an existing filter? 

Maybe we can add something to the error, like "You will be overwriting this filter."

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,202 @@
...
+bool
+OpenMCDomainFilterEditor::filterExists() const
+{
+  return openmc::model::filter_map.find(_filter_id) != openmc::model::filter_map.end();
+}
+
+int32_t
+OpenMCDomainFilterEditor::filterIndex() const
+{
+  if (!filterExists())
```
aprilnovak, line 105: I don't think this error message is tested anywhere. @pshriwise would you please add a test to cover it?

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,202 @@
...
+  for (auto bin_id : getParam<std::vector<std::string>>("bins"))
+  {
+    ids.push_back(std::stoi(bin_id));
+  }
+
+  if (_filter_type == OpenMCFilterType::cell)
+  {
+    openmc::CellFilter * cell_filter = dynamic_cast<openmc::CellFilter *>(filter);
+    if (!cell_filter)
+      mooseError(longName() + ": Filter " + std::to_string(_filter_id) + " is not a cell filter");
```
aprilnovak, line 126: Same here, for the cell, material, universe, and mesh filter checks -- I don't think there are tests covering these

src/base/OpenMCCellAverageProblem.C
pshriwise: Ah, that's a mistake. Nice catch thanks! I think I was trying to rule out something during testing.

src/userobjects/OpenMCTallyEditor.C
pshriwise: It refers to the entirety of the block following. I'll remove the antecedent in the comment for clarity.

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: Yep! Addressed.

src/userobjects/OpenMCUserObject.C
pshriwise: You...are not missing something. This feels like a holdover from when I allowed the tally editors to be attached to the Cardinal tallies. I'm going to look into the control flow of this again.

test/tests/userobjects/openmc_tally_editor/tests
pshriwise: Nice! Removing...

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: Good call. Updating.

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: What I'd like them to know is that they specified that the filter should be created but that one with the spec provided is already present in the model to the `create_filter` was redundant. In that case there's nothing wrong (that I can see) with simply using that filter, but it seems useful for the user to know it was already there.

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: Because the `initialize()` method is now called right away we shouldn't need that check anymore. Removing.

src/userobjects/OpenMCDomainFilterEditor.C
pshriwise: I'm honestly not sure how I'd trigger these errors from the test system, which might indicate that we don't need them, but with `dynamic_cast`'s going on here I think it's wise to keep them so we get some kind of output instead of a segfault if something goes wrong.

src/userobjects/OpenMCTallyEditor.C
```diff
@@ -0,0 +1,153 @@
...
+  bool create_tally = getParam<bool>("create_tally");
+
+  if (create_tally)
+  {
+    if (tallyExists())
+    {
+      mooseWarning(longName() + ": Tally " + std::to_string(_tally_id) +
```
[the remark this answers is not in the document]
pshriwise: Good call. I'll add some content to this message 👍🏻

test/tests/userobjects/openmc_tally_editor/tests
pshriwise: As I understand it, this test ensures that the string `"U238"` is present in the `tallies.out` file resulting from the OpenMC simulation by using the `file_expect_out` entry. Without the tally editor it wouldn't be there otherwise.

https://mooseframework.inl.gov/python/testers/CheckFiles.html

test/tests/userobjects/openmc_tally_editor/tests
```diff
@@ -0,0 +1,130 @@
...
+    check_files = tallies.out
+    requirement = 'Ensure that scores specified by a tally editor UO are present in the tally output'
+    file_expect_out = 'Scattering Rate'
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [absorption_u238]
+    type = CheckFiles
+    input = nuclide_absorption.i
+    check_files = tallies.out
+    requirement = 'Ensure that scores specified by a tally editor UO are present in the tally output'
```
aprilnovak, line 100: For SQA purposes, no two tests can have identical "requirement" tags. Can you check these two (and maybe others) and add more detail to the `requirement` so that they are not duplicates?

test/tests/userobjects/openmc_tally_editor/tests
aprilnovak: Oh neat!

At least one test using `CheckFiles` sound like they are relying on comparing numeric values though:

> This is compared by selecting just one of the existing nuclides, and comparing to the sum of the total nuclides.

Does this test actually capture those numeric values and look at them, or is it just looking for a regex in the `tallies.out`? Just trying to insulate from the possible scenario where this numeric behavior breaks in the future, but we might not catch it in any of our tests.

pshriwise: Ah! Good point. I borrowed this test block from the `TallyNuclides` object that was removed here. 

https://github.com/neams-th-coe/cardinal/blob/a347610bcda9d7f7662de5f6f2ce50331dc44f1a/test/tests/userobjects/openmc_tally_nuclides/tests#L28

The original test did what looks like some spot checking of the kappa-fission values. I don't see where it's checking the tally sum against the nuclide values. Happy to add the CSV check back in though!

aprilnovak: Yes, let's make sure all the original tests are retained in the new system.

I _think_ the original test was made by doing a manual check that the values sum appropriately to some number `X` (i.e., `Y+Z=X`). Then, the test is likely actually just reinforcing that `Y` and `Z` match those historical values of `Y` and `Z` (and thereby implicitly also matching the sum that they originally had to match).

src/userobjects/OpenMCTallyEditor.C
```diff
@@ -0,0 +1,161 @@
...
+  params.declareControllable("filter_ids");
+  params.declareControllable("multiply_density");
+  params.addClassDescription("A UserObject for creating and managing OpenMC tallies");
+  return params;
+}
+
+OpenMCTallyEditor::OpenMCTallyEditor(const InputParameters & parameters)
+  : OpenMCUserObject(parameters), _tally_id(getParam<int32_t>("tally_id"))
+{
+  this->initialize();
```
aprilnovak, line 51: Does this `initialize()` have to be here? I'd have thought that MOOSE properly calls the `initialize()` method in the right place. 

Also I guess I'm confused on the comment on line 57 - that function is called from the constructor in practice based on line 51, but the comment makes it sound like it isn't?

src/userobjects/OpenMCDomainFilterEditor.C
```diff
@@ -0,0 +1,201 @@
...
+  return params;
+}
+
+OpenMCDomainFilterEditor::OpenMCDomainFilterEditor(const InputParameters & parameters)
+  : OpenMCUserObject(parameters),
+    _create_filter(getParam<bool>("create_filter")),
+    _filter_id(getParam<int32_t>("filter_id")),
+    _filter_type(getParam<MooseEnum>("filter_type").getEnum<OpenMCFilterType>())
+{
+  this->initialize();
```
aprilnovak, line 51: Same question here about `initialize()` -- does this need to be here, or is it sufficient for MOOSE to call `initialize()` during the typical setup patterns?

aprilnovak:
I added a commit which makes `OpenMCUserObject` the base class for all the OpenMC-related userobjects (2 new ones in this PR, but also some pre-existing user objects). This allowed me to remove two of the tests being added in this PR, since now we can use a single test to check the `wrong_problem.i` catch (that user does not try to apply these OpenMC-only objects without `OpenMCCellAverageProblem`).

test/tests/userobjects/openmc_tally_editor/tests
pshriwise: Makes sense! The ability to evaluate OpenMC results is a little limited, so using a proxy value is a good idea.

In the original test it also seems that, based on the geometry and material compositions, it's checking that the nuclides of the tally are set correctly by ensuring that the tally value in Cell 1, which is filled with a material composed of only U235, is zero.

A snag though -- as I was looking at this today I realized that since we've decided that the tally editor shouldn't apply to Cardinal-mapped tallies I can't see a way to generated a CSV file with the same expectations. I recall now that's why I didn't include it in the tests. I could write a simple Python file to check the results and use the test harness, but that would require the OpenMC Python API to be installed (or at least h5py if we want to get a little hacky). 

I'll keep thinking on this, but let me know if you have any immediate thoughts!

src/userobjects/OpenMCTallyEditor.C
pshriwise: > Does this `initialize()` have to be here? I'd have thought that MOOSE properly calls the `initialize()` method in the right place.

Hrmm yeah you're probably right. Let me try the tests without that.

pshriwise: > Also I guess I'm confused on the comment on line 57 - that function is called from the constructor in practice based on line 51, but the comment makes it sound like it isn't?

Good point. That comment isn't really applicable anymore.

test/tests/userobjects/openmc_tally_editor/tests
aprilnovak: hmmm, yeah good point, it may be something we can't carry through in this PR. Do you think we could have this type of test in https://github.com/neams-th-coe/cardinal/pull/921?

src/userobjects/OpenMCTallyEditor.C
aprilnovak: What happens if the content inside `initialize()` is just moved into the constructor, and then the `initialize()` method is deleted? Does that also run into issues?

test/tests/userobjects/openmc_tally_editor/tests
pshriwise: Given that it's going to require the Python API in one way or another I think we should be able to support more detailed OpenMC testing yeah. I can make an issue to track this in the meantime though?

Maybe it would be good to target the Python API installation (at least for the test suite) as it's own task and re-implement this test as proof?

aprilnovak: Yeah, let's make an issue. Can you attach the original test files to it?

If we could automatically build the Python API as part of the Cardinal compilation process, that would be awesome! My naive approach would be to source a bash script from Cardinal's makefile which just calls `pip intall .` in the right place, but there's probably a more elegant way to do it.

pshriwise: Issue is in place: https://github.com/neams-th-coe/cardinal/issues/1032

I think that would do the trick, but I'll think on it some more. We'll want to be as careful as we can be to use the user's intended python environment when doing so.

src/userobjects/OpenMCTallyEditor.C
aprilnovak: If it works, then yes -- `initialize()` is something which MOOSE also calls somewhere in its execution of the `GeneralUserObject` base class, so at a minimum this block of code is being called twice (and possibly cause unexpected things to happen later, depending on what we put in it). Thanks!

test/tests/userobjects/openmc_tally_editor/tests
pshriwise: All set. Sorry for the duplication/inaccurate descriptions.

src/userobjects/OpenMCTallyEditor.C
pshriwise: True true. It doesn't seem as though it was being called at all, but maybe. I didn't dig. The content has been moved into the UO constructors and that seems to work just fine.

src/userobjects/OpenMCDomainFilterEditor.C
aprilnovak: Sounds good, that's fine by me!

test/tests/userobjects/openmc_tally_editor/tests
aprilnovak: no worries, there's no way to know that unless generating the SQA documentation. Thanks!

aprilnovak:
Is this ready to go?

pshriwise:
As far as I'm concerned yep

aprilnovak:
LGTM!

PR #931, opened 2024-07-15
Addition of a tally system and `[Tallies]` block

nuclearkevin:
## Reason
This PR implements a new tally system and an associated block in Cardinal's input syntax based on a discussion with @aprilnovak (ref. #922). This allows for the addition of multiple different tallies at once which have different scores, an example of which can be found below (tests/neutronics/tally_system/multi_diff.i):
```
[Problem]
  verbose = true
  power = 1e4
  temperature_blocks = '100 200'
  density_blocks = '200'
  cell_level = 0
  initial_properties = xml

  # The global tally check is disabled because we have a loosely fitting unstructured mesh tally.
  normalize_by_global_tally = false

  source_rate_normalization = 'kappa_fission'

  [Tallies]
    [Cell]
      type = CellTally
      score = 'kappa_fission'
      blocks = '100 200'
    []
    [Mesh]
      type = MeshTally
      score = 'flux'
      mesh_translations = '0 0 0
                           0 0 4
                           0 0 8'
      mesh_template = ../meshes/sphere.e
    []
  []
[]
```
This is advantageous if different spatial filters are required for different scores. This PR will also (hopefully) make it easier to add new filters in the future (such as an `EnergyFilter`) without requiring the user to manually post-process tallies in the input file when they don't need certain scores binned according to a new filter.

## Design
The new tally system adds the following new objects:

- `TallyBase`: A class which wraps a single OpenMC tally to provide common data types and virtual functions required by spatially dependant tallies. `TallyBase` classes are constructed by an `OpenMCCellAverageProblem` through the `addTallyObject(...)` function, which mirrors how MOOSE handles `MooseObject` construction in `FEProblemBase`. The virtual functions that must be overloaded by derived tallies (`CellTally` and `MeshTally`) are:
  - `initializeTally(...)`: A function where derived tallies can implement functionality required to initialize an OpenMC tally. Called by the `OpenMCCellAverageProblem` in `initializeTallies(...)`.
  - `resetTally(...)`: A function where derived tallies can implement functionality required to reinit an OpenMC tally. Called by the `OpenMCCellAverageProblem` in `resetTallies(...)`.
  - `storeResults(...)`: A function where derived tallies can store their tally results in a field auxvariable. Called by the `OpenMCCellAverageProblem` in `syncSolutions(...)`.
  - `storeExternalResults(...)`: A function where derived tallies can store other requested tally results (unrelaxed / unrelaxed standard deviation) in a field auxvariable. Called by the `OpenMCCellAverageProblem` in `syncSolutions(...)`.
- `CellTally`: A class derived from `TallyBase` which implements distributed cell tallies. Replaces the functionality of `tally_type = cell` in the `OpenMCCellAverageProblem`.
- `MeshTally`: A class derived from `TallyBase` which implements unstructured mesh tallies that may also be translated in space. Replaces the functionality of `tally_type = mesh` in the `OpenMCCellAverageProblem`.
- `AddTallyAction`: An `Action` which is responsible for adding `TallyBase` objects to an `OpenMCCellAverageProblem`. This action calls `addTallyObject(...)` when the simulation reaches the new `add_tallies` task during execution.
  - When the tally to be added is a `CellTally`, a single tally object is added.
  - When the tally to be added is a `MeshTally`, multiple tally objects may be added if the user specifies mesh translations. These multiple tallies all accumulate in the same set of auxvariables.

Multiple tallies of the same type are allowed by this system so long as they do not share scores. In general, multiple tally objects that share scores are not allowed as it becomes difficult to normalize the tallies due to possible bin overlap.

At present time this new system is **NOT** backwards compatible with previous versions of Cardinal. Attempting to use an input deck where a single tally is added in the `[Problem]` block will result in an error telling the user to use the `[Tallies]` block. If backwards compatibility is desirable, I can emulate the old system by having `OpenMCCellAverageProblem` add an `AddTallyAction` in it's constructor with the input parameters of the required tally.

## Impact
- Allows for greater tally flexibility, both for users and for the addition of new features.
- Reduces the number of input parameters in `OpenMCCellAverageProblem` and decreases the overall complexity of the class.
- Breaks input decks which don't use the new tally system (at the moment).
  - This can be fixed if backwards compatibility is desired.

## Other
This PR is unfortunately quite large due to the need to update the input syntax for every tutorial and test. None of the tests required re-golding, so I don't believe updating the syntax will cause any CI/CD related issues. The documentation website has also been updated to reflect the new system (including the tutorial explanations). I believe that I caught every reference of the old system, but if I missed something please let me know and I'll get it changed over.


aprilnovak:
Awesome!! I'm also requesting a review from @pshriwise to get his feedback on this.

aprilnovak:
How are the tallies normalized when we have > 1? I see there's a new parameter, `global_tally_estimator`. Would it instead make sense to normalize each tally individually? If you have a cell tally with a tracklength estimator, but it's normalized by a collision estimator, power won't be preserved as the user might expect, right?

nuclearkevin:
The default at the moment is that tallies are normalized by a single global tally. If there is a mismatch between the global estimator and the estimators in the added tallies, the user is warned (lines 2727 to 2760 in `OpenMCCellAverageProblem.C`). 

Thinking about it now, it would make more sense to set the default to normalizing by the local tallies due to this change.

aprilnovak:
I think the default of normalizing by a global tally does help to catch user mistakes where they are missing part of the heating contribution in their coupling.

Could we retain the original behavior, but just apply it individually to each tally?

aprilnovak:
Another minor request -- would you please add a summary of the changes to the `news.md` (make an entry for July 2024)?

nuclearkevin:
> Could we retain the original behavior, but just apply it individually to each tally?

I think the easiest way to maintain that behaviour would be to add multiple global tallies (each with their own estimators, as required) and mapping those to tallies added by the tally system. It shouldn't be too difficult to add.

> Another minor request -- would you please add a summary of the changes to the news.md (make an entry for July 2024)?

Sure thing!

pshriwise:
I like these changes @nuclearkevin! Unless you're strongly opposed, one thing @aprilnovak and I ended up on the same page about was keeping these in the `OpenMCCellAverageProblem` block for clarity. The input block is very out of context otherwise as these tally constructs involve mapping logic in `OpenMCCellAverageProblem`.

nuclearkevin:
@pshriwise I'm not opposed to `[Tallies]` becoming a sub-block of `[Problem]`, especially if it improves clarity for users. I don't think MOOSE allows for input syntax to explicitly be a sub-block of a MOOSE object (`OpenMCCellAverageProblem`), so the best that we can get is something like this:
```
[Problem]
  verbose = true
  power = 1e4
  temperature_blocks = '100 200'
  density_blocks = '200'
  cell_level = 0
  initial_properties = xml

  # The global tally check is disabled because we have a loosely fitting unstructured mesh tally.
  normalize_by_global_tally = false

  source_rate_normalization = 'kappa_fission'

  [Tallies]
    [Cell]
      type = CellTally
      tally_score = 'kappa_fission'
      tally_blocks = '100 200'
    []
    [Mesh]
      type = MeshTally
      tally_score = 'flux'
      mesh_translations = '0 0 0
                           0 0 4
                           0 0 8'
      mesh_template = ../meshes/sphere.e
    []
  []
[]
```
If that looks good to you I can get it implemented when I have the time.

pshriwise:
@nuclearkevin yeah that's exactly what I had in mind. Thank you!!

doc/content/source/tallies/CellTally.md
```diff
@@ -0,0 +1,33 @@
+# CellTally
+
+!alert note
+`CellTally` can only be added to problems when the input files contains a [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
+in the `[Problem]` block. Otherwise, attempting to add this tally will result in an error.
+
+## Description
+
+The `CellTally` class wraps an OpenMC tally with a distributed cell filter to enable data transfer
+from CSG cells to a [MOOSEMesh](https://mooseframework.inl.gov/source/mesh/MooseMesh.html) mirror
```
aprilnovak, line 10: suggest to replace `MOOSEMesh` by `MooseMesh` to match the class name

doc/content/source/tallies/CellTally.md
```diff
@@ -0,0 +1,33 @@
...
+of the OpenMC geometry. The cell to element mapping established by the [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
+is used to facilitate this data transfer to the mesh mirror. When using a `CellTally`, `tally_blocks`
+specifies which blocks in the `[Mesh]` should be tallied. Then, any OpenMC cells that map to those
+blocks are added to a cell tally, with one bin for each unique cell ID/instance combination.
+`check_equal_mapped_tally_volumes` is used to confirm whether the elements mapped to a specific cell
```
aprilnovak, line 15: This explanation isn't quite correct - this parameter doesn't check if the mapped element volume matches the CSG cell volume. Instead, it checks that if we find 99 CSG cells to tally, that they all map to the same volume on the MOOSE side.

include/base/OpenMCCellAverageProblem.h
```diff
@@ -360,9 +402,9 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
    * @param[in] score tally score
    * @return sum of the tally
    */
-  Real getCellTally(const unsigned int & var_num,
-                    const std::vector<xt::xtensor<double, 1>> & tally,
-                    const unsigned int & score);
+  // Real getCellTally(const unsigned int & var_num,
```
aprilnovak, line 405: If we don't need them anymore, let's remove the comments. Same applies to `getMeshTally`

test/tests/neutronics/tally_system/tests
```diff
@@ -0,0 +1,49 @@
...
+    exodiff = multi_out.e
+    requirement = "The system shall correctly apply and normalize two different local tallies with different scores. "
+                  "The gold file was generated using an input that had a single tally with multiple scores."
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [duplicate_scores]
+    type = RunException
+    input = multi.i
+    cli_args = 'Problem/Tallies/Cell_2/tally_score="flux kappa_fission"'
+    expect_err = "You have added 2 tallies which score kappa-fission!"
```
aprilnovak, line 21: What is the motivation for having this error? I could (maybe) see someone interested in having the same score, but with different spatial representation (e.g., cell vs. mesh). It's definitely an edge case and could be fine to keep this as an error, but I'd like to get a better sense on whether there's something fundamental here that I'm missing, or if this is just for convenience?

test/tests/neutronics/tally_system/tests
```diff
@@ -0,0 +1,49 @@
...
+    input = multi_estimator.i
+    exodiff = multi_estimator_out.e
+    requirement = "The system shall correctly normalize local tallies with different estimators using multiple global tallies."
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [multiple_tallies]
+    type = Exodiff
+    input = multi.i
+    exodiff = multi_out.e
+    requirement = "The system shall correctly apply and normalize two different local tallies with different scores. "
```
aprilnovak, line 13: Nice! This is a great test for cell tallies. It may be useful to do the same type of test, but for (i) mesh tallies, (ii) relaxation, and (iii) triggers just to be extra sure we're not missing anything.

test/tests/neutronics/tally_system/tests
```diff
@@ -0,0 +1,49 @@
...
+    expect_err = "The local tallies added in the \[Tallies\] block do not contain the requested heating score"
+    requirement = "The system shall error if more than one tally is provided and the requested heating score is in "
+                  "none of the tallies."
+    required_objects = 'OpenMCCellAverageProblem'
+  []
+  [multiple_different_tallies]
+    type = Exodiff
+    input = multi_diff.i
+    exodiff = multi_diff_out.e
+    requirement = "The system shall allow calculations with multiple different tallies."
```
aprilnovak, line 38: Ideally, you will have run the legacy input file 2x, one with the cell tally and one with the mesh tally, and then confirm visually in paraview that having both tallies simultaneously now gives identical results (I'm not sure on a more programmatic way to do that via gold file generation directly, as you have done in the `multiple_tallies` test).

test/tests/neutronics/triggers/multi_rel_err.i
```diff
@@ -22,15 +22,20 @@
...
-  tally_score = 'damage_energy kappa_fission'
-
-  tally_trigger = 'rel_err rel_err'
-  tally_trigger_threshold = '2e-2 2e-2'
   max_batches = 200
+
+  [Tallies]
+    [Cell]
+      type = CellTally
+      tally_blocks = '100'
```
aprilnovak, line 32: What do you think of shortening some of the names of these parameters, now that they're nested inside these blocks? Users may find the following more straightforward/faster, but perhaps there could be a good reason not to change the names, to minimize effort to port things over.

```
  [Cell]
      type = CellTally
      blocks = '100'
      score = 'damage_energy kappa_fission'

      trigger = 'rel_err rel_err'
      trigger_threshold = '2e-2 2e-2'
  []
```

include/base/OpenMCCellAverageProblem.h
nuclearkevin: Oops, thanks for catching this! I didn't realize some remnants of the old tally system slipped through. I've also swept through and removed some other functions in `OpenMCCellAverageProblem` and `OpenMCProblemBase` which aren't used in the `Tallies` sub-block.

test/tests/neutronics/tally_system/tests
nuclearkevin: The motivation behind this error is the case where tallies are normalized by the sum of a local tally as opposed to a global tally. If two different tallies share a score they could have overlapping bins, so when the sum of the local tally is computed we might double count tally results. 

This could be alleviated if each tally is normalized by it's own sum as opposed to being normalized by a sum over all of the local tally results for a score. This would lead to complications when trying to accommodate multiple distributed mesh tallies where you need to normalize them by the sum over each "instance" of the tally when performing local normalization; the new tally system is designed such that a single `TallyBase` implies a single `openmc::Tally`. Other complications include how different tallies would store their variables on the mesh mirror if they have the same scores, and which tally's score would be used for normalizing all of the tallies when running in eigenvalue mode (i.e., which `kappa_fission` do you use if both tallies score `kappa_fission`?).

I made the decision to avoid all of these issues by not allowing multiple tallies to share a score, with the compromise that some users wouldn't be able to perform comparisons of different spatial tallies in a single input file.

test/tests/neutronics/triggers/multi_rel_err.i
nuclearkevin: I like the idea quite a bit, though I was hesitant to change parameter names precisely because of that reason (and to minimize the effort required to update all of the tests and tutorials). If you think it's worthwhile, I can shorten the syntax.

test/tests/neutronics/tally_system/tests
nuclearkevin: Sounds good, I'll add some extra tests to cover those cases. When it comes to tally relaxation, should I add a case for each relaxation scheme or would constant relaxation work just fine?

test/tests/neutronics/tally_system/tests
nuclearkevin: I don't believe that works for the `MeshTally` as it is normalized to a source rate (`kappa_fission`) scored by a CellTally in this test. At the very least, I compared the `CellTally` results to a legacy cell tally in Paraview and they are identical.

nuclearkevin:
@aprilnovak I believe I've addressed all of your requested changes. Feel free to let me know if there's anything else you'd like changed or added!

aprilnovak:
Thanks @nuclearkevin! I think there's just one last item which occurred to me. It might be complicated.

We populate the `_cell_to_elem` map (and other objects in `OpenMCCellAverageProblem`) based on an intersection of the blocks given by `tally_blocks` (old syntax), `temperature_blocks`, and `density_blocks`. I'm wondering what happens if you have two `CellTally` objects, each defined on the same blocks, but with different scores. Are we re-adding duplicate elements to the `_cell_to_elem` map?

If so, I think you'd be able to detect this using the `CellVolumeAux`, to display the element volume each cell maps to. If things are wrong, then `CellVolumeAux` will give you different values depending on whether you had 1 tally with N scores, vs. N tallies with 1 score each (same block definitions for all of them). 

Can you please test this out and see what happens?

nuclearkevin:
@aprilnovak It looks like the cell volume reported by the `CellVolumeAux` doesn't change if a single `CellTally` with two scores is added vs two `CellTally` with a single score each (on shared blocks).

Postprocessor Values (single `CellTally`):
```
+----------------+----------------+----------------+----------------+
| time           | Cell_1_Vol     | Cell_2_Vol     | Cell_3_Vol     |
+----------------+----------------+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   1.322128e+01 |   1.322128e+01 |   1.322128e+01 |
+----------------+----------------+----------------+----------------+
```

Postprocessor Values (two `CellTally` with different scores on shared blocks):
```
+----------------+----------------+----------------+----------------+
| time           | Cell_1_Vol     | Cell_2_Vol     | Cell_3_Vol     |
+----------------+----------------+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   1.322128e+01 |   1.322128e+01 |   1.322128e+01 |
+----------------+----------------+----------------+----------------+
```
Also adding this to `tests/neutronics/tally_system/multi_cell.i` as an additional check just to be sure.

aprilnovak:
Which `volume_type` did you use for `CellVolumeAux`? Just check that you used `volume_type = mapped`

nuclearkevin:
Yep, I used `volume_type = mapped`for the check.

aprilnovak:
Great! This looks good-to-go. I'm just going to do a local checkout and run some of the tests, see if I can think of other error checks which may be missing.

@pshriwise how would you prefer to handle the timing of this PR in contrast to #923? #923 is for a GAIN project which has a close-out approaching soon, so I'll defer to Patrick on what order we should merge these in order to minimize any schedule delays with GAIN.

aprilnovak:
Looks like each of the test gold files is rather large. Can we test this equivalently using just 1 pebble, instead of 3 with the `CombinerGenerator`? We can squash all the commits when it's merged, so don't worry about removing big files from the git history.

nuclearkevin:
Updated the tests. The only major change is that moving to a single pebble requires `check_tally_sum = false` as the xml files still model three pebbles (so the global tally ends up being larger than the local tally).

aprilnovak:
Was `check_tally_sum=true` working before? I'd be surprised because the meshes are not conformal to the pebbles, so even with all 3 included you'd still have slivers of pebbles outside the mesh but inside the CSG spheres which would contribute to the global tally.

aprilnovak:
oh gotcha! How about we do this -- let's actually keep all 3 pebbles for the cell tests, since they would catch anything going wrong with instancing. But instead of an exodiff test, for those we can easily do a CSVDiff (`PointValue` for the tally value at the centroid of each pebble).

nuclearkevin:
`CellTally` tests have been changed to `CSVDiffs`. I took the liberty of reverting `multi_mesh.i` back to three pebbles to make sure we cover the case where there are distributed mesh tallies.

aprilnovak:
Sorry for the kind of piecemeal review process, but another thought came to mind. Can we consolidate some of the duplicate behavior in `CellTally` and `MeshTally` into the base class? For instance, the `initializeTally` method has a lot of duplicate code - I'd suggest we have virtual functions in the base class which derived classes can override. This should make code maintenance more straightforward while also streamlining an upcoming FET implementation for Cardinal that another student is working on.

nuclearkevin:
No worries! It's a rather large PR that was submitted near the end of the summer, it was destined to take a while to review.

I could move the `initializeTally(...)` function into the base class and instead replace it with a virtual function called `spatialFilter(...)` (or something similar) which the derived tally objects can use to create the spatial filter that the class corresponds to. I can also consolidate the `resetTally(...)` function into the base class with a default implementation, but still allow derived tally objects to override it so they can add custom behavior (`MeshTally` needs to reset the mesh it uses, while `CellTally` does not).

I don't know if I can really move the functionality in `storeResults(...)` to the base class since it heavily depends on the derived tally object. I definitely find a way to merge `storeResults()` and `storeExternalResults()` into a single function per derived tally to decrease the amount of code duplication though.

aprilnovak:
That sounds like a good plan to me - I don't anticipate we'd be able to move absolutely everything, I'll trust your judgement on what makes sense to consolidate into the base class (a few obvious places stood out to me, which you've touched on).

PR #940, opened 2024-07-24
Install xs into the container

loganharbour:
Installs the current cross sections into the cardinal container. Needed so that trainees don't need to when using a docker or apptainer container.


loganharbour:
@aprilnovak this is ready. Includes the XS in the container.

PR #951, opened 2024-08-17
Addition of a filter system and `[Filters]` block

### LINKED ISSUE #602: Add energy binning for tallies

nuclearkevin:
## Reason 

This PR implements a filter system, an associated block in Cardinal's input syntax (`[Problem/Filters]`), and four new filters which tallies added through the `[Problem/Tallies]` block can access. The added filter objects are:
- `EnergyFilter`: a thin wrapping of an OpenMC [EnergyFilter](https://docs.openmc.org/en/stable/pythonapi/generated/openmc.EnergyFilter.html).
- `PolarAngleFilter`: a thin wrapping of an OpenMC [PolarFilter](https://docs.openmc.org/en/stable/pythonapi/generated/openmc.PolarFilter.html).
- `AzimuthalAngleFilter`: a thin wrapping of an OpenMC [AzimuthalFilter](https://docs.openmc.org/en/stable/pythonapi/generated/openmc.AzimuthalFilter.html).
- `FromXMLFilter`: makes an arbitrary OpenMC filter (added in the `tallies.xml` file) available to Cardinal tallies through the filter's id.

The input syntax is similar to the recently added tally system:
```
[Problem]
  verbose = true
  power = 1e4
  temperature_blocks = '100'
  cell_level = 0
  initial_properties = xml

  source_rate_normalization = 'kappa_fission'

  [Tallies]
    [Heating]
      type = CellTally
      score = 'kappa_fission'
      blocks = '100 200'
    []
    [Flux]
      type = CellTally
      score = 'flux'
      blocks = '100 200'
      filters = 'Energy Polar'
    []
  []

  [Filters]
    [Energy]
      type = EnergyFilter
      # CASMO 2 group structure for testing. May result in some missed particles
      energy_boundaries = '0.0 6.25e-1 2.0e7'
    []
    [Polar]
      type = PolarAngleFilter
      num_equal_divisions = 2
    []
  []
[]
```
(`tests/neutronics/filters/multi_filter.i`)

## Design

The new filters system has the following objects other than the previously mentioned filters:
- `FilterBase`: A class which wraps a single OpenMC filter. `FilterBase` classes are constructed by an `OpenMCCellAverageProblem` through the `addFilter(...)` function, which mirrors the tally system and other MOOSE object construction patterns. Filters derived from `FilterBase` must override the `binName(...)` function, which returns a unique name for each bin in the filter to enable output formatting. Otherwise, classes which derive from `FilterBase` must ensure that `_filter` and `_filter_index` are populated by the derived filter's constructor.
- `AddFilterAction`: An Action which is responsible for adding `FilterBase` objects to an OpenMCCellAverageProblem. This action calls `addFilter(...)` when the simulation reaches the new add_filters task during execution. Filters are added before tallies and after the initialization of the problem.

The filters added through the filters system must use discrete bins as the tallies are still normalized by a sum over all bins. This restriction could be lifted, however this would require a substantial change to tally normalization to allow bins that contain coefficients to be skipped when computing the sum over the entire tally.

In addition to the above, this PR modifies the tally system to reduce the amount of duplicate code and fixes a bug in both `MeshTally` and `CellTally` which caused a memory leak when using a `MOABSkinner`. These changes can be removed from this PR and submitted in their own PR to expedite merging, if that's preferred.

## Impact

- Enables multi-physics coupling to other codes which require fast fluxes / fluences (such as Bison).
- Allows for greater flexibility when post-processing tallies.
- Opens the door to the generation of surface / volume sources for deterministic radiation transport (https://github.com/openmc-dev/openmc/issues/3031).
- Allows for the comparison of multi-group reaction rates / fluxes between deterministic radiation transport codes (such as Griffin) and Cardinal.

Closes #602.


doc/content/source/actions/AddFilterAction.md
```diff
@@ -0,0 +1,22 @@
+# AddFilterAction
+
+!alert note
+`AddFilterAction` can only add filters to problems which contain a [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
+in the `[Problem]` block. Otherwise, attempting to add a tally will result in an error.
```
aprilnovak, line 5: tally -> filter ?

doc/content/source/filters/AzimuthalAngleFilter.md
```diff
@@ -0,0 +1,27 @@
+# AzimuthalAngleFilter
+
+!alert note
+`AzimuthalAngleFilter` can only be added to problems when the input files contains a [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
+in the `[Problem]` block. Otherwise, attempting to add this filter will result in an error.
+
+## Description
+
+The `AzimuthalAngleFilter` object provides a thin wrapper around an OpenMC `AzimuthalFilter` which local tallies (added through the
```
aprilnovak, line 9: In the interest of technical maintenance, do we want to link to the OpenMC documentation instead of re-writing it? In case something changes in the future, this could slip through the cracks (same comment applies for other filters)

doc/content/source/filters/EnergyFilter.md
```diff
@@ -0,0 +1,27 @@
+# EnergyFilter
+
+!alert note
+`EnergyFilter` can only be added to problems when the input files contains a [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
```
aprilnovak, line 4: files -> file

doc/content/source/filters/EnergyFilter.md
```diff
@@ -0,0 +1,27 @@
...
+The `EnergyFilter` object provides a thin wrapper around an OpenMC `EnergyFilter` which local tallies (added through the
+[tally system](AddTallyAction.md)) can access. This filter bins particles based off of the particle energy before the collision
+has occured (for collision or analog estimators) or the particle energy (for tracklength estimators). The ordering of the bins in
+`EnergyFilter` can be reversed (set to be descending in terms of energy) by setting `reverse_bins = true` if you wish to use
+the conventional multi-group radiation transport formalism.
+
+## Example Input File Syntax
+
+As an example, a [MeshTally](MeshTally.md) named `Flux` applies an `EnergyFilter` named `Energy` to break the total flux up
+into multi-group fluxes using the CASMO-2 group structure.
```
aprilnovak, line 18: Can we expose the built-in group structures in OpenMC? So that if the user wants many groups, they don't need to type them out, but could use the enumeration from OpenMC?

doc/content/source/filters/PolarAngleFilter.md
```diff
@@ -0,0 +1,27 @@
+# PolarAngleFilter
+
+!alert note
+`PolarAngleFilter` can only be added to problems when the input files contains a [OpenMCCellAverageProblem](OpenMCCellAverageProblem.md)
```
aprilnovak, line 4: files -> file

include/actions/AddTallyAction.h
```diff
@@ -40,11 +40,7 @@ class AddTallyAction : public MooseObjectAction
    */
   void addMeshTally(unsigned int instance = 0, const Point & translation = Point(0.0, 0.0, 0.0));
 
-  /**
-   * This function adds a a tally which may be an offset into an array
-   * of translations with index instance. Used for translated mesh tallies
-   * that all use the same base mesh for the spatial filter.
-   */
+  /// This function adds a non-mesh tally.
```
aprilnovak, line 43: Can this doc string be made more specific? Is this true? If so, it may be clearer for folks reading Doxygen

`This function adds a cell tally`

include/filters/FilterBase.h
```diff
@@ -0,0 +1,77 @@
...
+/// Forward declarations.
+class OpenMCCellAverageProblem;
+
+/**
+ * A class which provides a thin wrapper around an OpenMC Filter for
+ * the purposes of adding a tally via the MOOSE input syntax. This
+ * class does not need init / reset functions as non-spatial tallies
+ * don't change during simulation execution (at the moment).
+ *
+ * The filter system should NOT wrap spatial filters as the problem
```
aprilnovak, line 34: What are your thoughts on supporting ALL of the OpenMC filters, but requesting the user to create them from the Python API, and refer to them using a Cardinal class which gets passed an `int` to refer to the filter ID? (or something along those lines).

doc/content/source/filters/EnergyFilter.md
nuclearkevin: @aprilnovak AFAIK the built-in group structures in OpenMC are purely on the [Python side of the API](https://github.com/openmc-dev/openmc/blob/develop/openmc/mgxs/__init__.py) so there's no "easy" way of exposing them through OpenMC headers. I could copy and paste the structures into C++ code (as a constexpr array) if you're OK with the `EnergyFilter.h` getting rather bloated, but I can't think of a more streamlined way to go about this...

I think it's still worthwhile to add, just wanted to make sure you're OK with this implementation approach before I start pasting the structures in.

include/filters/FilterBase.h
nuclearkevin: I think it's a good idea and could probably be implemented in a class called `ArbitraryFilter` (or something similar). I'd want to put some padding on using an arbitrary filter with a few `mooseWarnings` and `mooseErrors` letting the user know that they cannot use local normalization or tally sum checking with an arbitrary filter;  a sum over all bins for some of the tally filter options (i.e. spherical harmonics filters, Legendre filters) don't make sense as you're summing function expansion coefficients (which are not guaranteed to equal the global tally sum).

include/actions/AddTallyAction.h
nuclearkevin: At the moment it's true in that the only non-translatable tallies in Cardinal are `CellTally` objects. My thought is that if other tally objects (such as ones which implement a ZernikeFilter) would also get added by that function.

I can definitely try to make the doc string clearer though. Adding a list of tallies which would be added by that function and a better description of what it does would help a lot compared to a single line.

doc/content/source/filters/EnergyFilter.md
aprilnovak: I think pasting them in is fine -- I assume they are not going to be revised in the future so it'll be a lot of line changes now, but shouldn't be something we need to really maintain a bunch.

include/filters/FilterBase.h
aprilnovak: right, there are downsides to having a fully opaque implementation, in terms of normalization -- but it could automatically support all of the filters, and we can continue to have those which we use frequently in Cardinal (like the ones you've already added) as thin wrappers.

nuclearkevin: My solution to the normalization problem was to have `FromXMLFilter` (the new filter filter that pulls from `tallies.xml`) throw an error if the filter at `id` is a functional expansion filter. The user can override this behaviour, but they're warned against it and informed that they would need to manually normalize all Cardinal tallies which use the functional expansion filter during post-processing if they do.

tjlaboss:
👀

src/filters/FromXMLFilter.C
```diff
@@ -0,0 +1,98 @@
...
+                        false,
+                        "Whether functional expansion filters are allowed or not. Tallies with "
+                        "these filters are likely to fail "
+                        "automatic normalization as a sum over all bins may not make sense for a "
+                        "certain functional expansion.");
+
+  return params;
+}
+
+FromXMLFilter::FromXMLFilter(const InputParameters & parameters)
```
aprilnovak, line 47: We should probably throw an error if the user is trying to add a spatial filter, right? Since we're handling those via Cardinal?

src/filters/AzimuthalAngleFilter.C
```diff
@@ -0,0 +1,109 @@
...
+
+    // Warn the user if there is the possibility of missed particles.
+    if (_azimuthal_angle_bnds.front() > (-1.0 * libMesh::pi) ||
+        _azimuthal_angle_bnds.back() < libMesh::pi)
+      mooseWarning(
+          "The bin boundaries don't contain one of the following: -pi or pi. Some particles may be "
+          "missed during normalization.");
+  }
+  else
+    mooseError("No bins have been specified! Please set either 'num_equal_divisions' or "
```
aprilnovak, line 92: I like to combine line 92 with line 53 using 

```
 if (isParamValid("num_equal_divisions") == isParamValid("azimuthal_angle_boundaries"))
    mooseError(
        "You have set either both or noen of 'num_equal_divisions' and 'azimuthal_angle_boundaries'! Please set one of these variables.");
```

or something along those line

src/filters/AzimuthalAngleFilter.C
```diff
@@ -0,0 +1,109 @@
...
+    // Make sure we have at least two boundaries to form bins.
+    if (bnds.size() < 2)
+      paramError("azimuthal_angle_boundaries",
+                 "At least two azimuthal angles are required to create bins!");
+
+    // Sanity check the boundaries to make sure they're provided in increasing order.
+    for (unsigned int i = 0; i < bnds.size() - 1; ++i)
+    {
+      if (bnds[i] > bnds[i + 1])
+        paramError("azimuthal_angle_boundaries",
```
aprilnovak, line 77: what if we sorted them ourselves? Mostly I am wondering about the use case if someone inserts the bins in reverse order which might happen.

src/filters/EnergyFilter.C
```diff
@@ -0,0 +1,877 @@
...
+      "groups are arranged in descending order (i.e. group 1 is fast, group 2 is thermal).");
+
+  return params;
+}
+
+EnergyFilter::EnergyFilter(const InputParameters & parameters)
+  : FilterBase(parameters), _reverse_bins(getParam<bool>("reverse_bins"))
+{
+  if (isParamValid("energy_boundaries") && isParamValid("group_structure"))
+    mooseError("You have set both 'energy_boundaries' and 'group_structure'! Only one bin "
```
aprilnovak, line 56: same comment as above regarding switching `&&` to `==` to hit two birds with one stone regarding error checking.

src/filters/EnergyFilter.C
```diff
@@ -0,0 +1,877 @@
...
+
+  // Two boundaries are required at minimum to form energy bins.
+  if (_energy_bnds.size() < 2)
+    paramError("energy_boundaries",
+               "At least two energy values are required to create energy bins!");
+
+  // Sanity check the energy boundaries to make sure they're increasing monotonically.
+  for (unsigned int i = 0u; i < _energy_bnds.size() - 1; ++i)
+    if (_energy_bnds[i] >= _energy_bnds[i + 1])
+      paramError("energy_boundaries",
```
aprilnovak, line 76: thoughts on sorting ourselves?

src/filters/EnergyFilter.C
```diff
@@ -0,0 +1,877 @@
...
+    5.7544e7,  6.0256e7,  6.3096e7,  6.6069e7,  6.9183e7,  7.2444e7,  7.5858e7,  7.9433e7,
+    8.3176e7,  8.7096e7,  9.1201e7,  9.5499e7,  1.0000e8,  1.0471e8,  1.0965e8,  1.1482e8,
+    1.2023e8,  1.2589e8,  1.3183e8,  1.3804e8,  1.4454e8,  1.5136e8,  1.5849e8,  1.6596e8,
+    1.7378e8,  1.8197e8,  1.9055e8,  1.9953e8,  2.0893e8,  2.1878e8,  2.2909e8,  2.3988e8,
+    2.5119e8,  2.6303e8,  2.7542e8,  2.8840e8,  3.0200e8,  3.1623e8,  3.3113e8,  3.4674e8,
+    3.6308e8,  3.8019e8,  3.9811e8,  4.1687e8,  4.3652e8,  4.5709e8,  4.7863e8,  5.0119e8,
+    5.2481e8,  5.4954e8,  5.7544e8,  6.0256e8,  6.3096e8,  6.6069e8,  6.9183e8,  7.2444e8,
+    7.5858e8,  7.9433e8,  8.3176e8,  8.7096e8,  9.1201e8,  9.5499e8,  1.e9};
+
+constexpr double S_ECCO_1968[]{
```
aprilnovak, line 577: This is really minor, but I'm slightly in favor of moving these group structures into a header file. Maybe we'll want to include them elsewhere in the code base.

src/filters/FromXMLFilter.C
```diff
@@ -0,0 +1,98 @@
...
+               "added in the OpenMC model and you've supplied the correct filter id.");
+
+  _filter_index = openmc::model::filter_map.at(_filter_id);
+  _filter = openmc::model::tally_filters[_filter_index].get();
+
+  // Check to see if the filter is an expansion filter.
+  bool is_exp;
+  switch (_filter->type())
+  {
+    case openmc::FilterType::LEGENDRE:
```
aprilnovak, line 65: This can be shortened to

```
    case openmc::FilterType::LEGENDRE:
    case openmc::FilterType::SPATIAL_LEGENDRE:
    case openmc::FilterType::SPHERICAL_HARMONICS:
    case openmc::FilterType::ZERNIKE:
    case openmc::FilterType::ZERNIKE_RADIAL:
    case openmc::FilterType::ENERGY_FUNCTION:
      is_exp = true;
      break;
```

test/tests/neutronics/filters/azimuthal/tests
```diff
@@ -0,0 +1,67 @@
+[Tests]
+  [azimuthal_cell_provided_bins]
+    type = CSVDiff
+    input = cell.i
+    csvdiff = cell_provided_out.csv
+    cli_args = 'Problem/Filters/Azimuthal/azimuthal_angle_boundaries="-3.1415926535897932384626433832795029 0.0 3.1415926535897932384626433832795029" Outputs/file_base=cell_provided_out'
```
aprilnovak, line 6: does `-${pi}$` work the same?

aprilnovak:
just needs a rebase and then I am good to merge

nuclearkevin:
@aprilnovak I've finished rebasing this, it should be good to go now!

PR #954, opened 2024-09-05
Allow OpenMC tally triggers to ignore bins with zeros

### LINKED ISSUE #875: Explore `ignore_zeros` option for the tallies added by Cardinal
## Reason
A new capability has been added to OpenMC, `ignore_zeros` (https://github.com/openmc-dev/openmc/pull/2928), which will ignore zero tally bins when applying triggers. We should expose this option through the Cardinal interface so that we can control it via the MOOSE inputs.

## Design
Add `ignore_zeros` as an option to `OpenMCCellAverageProblem` that will get passed into OpenMC when we create tallies.

## Impact
New feature, nice for @meltawila's fusion cases where we use mesh tallies but certain scores (e.g., tritium production) necessarily have zero values in some locations.

nuclearkevin:
This PR exposes the `ignore_zeros` option in OpenMC tally triggers. This allows tallies to skip empty bins which never get tallied on when computing trigger metrics. Examples include kappa fission or tritium production on mesh tallies.

Closes #875.


src/tallies/TallyBase.C
```diff
@@ -48,6 +48,9 @@ TallyBase::validParams()
       "this same trigger is applied to all scores.");
   params.addRangeCheckedParam<std::vector<Real>>(
       "trigger_threshold", "trigger_threshold > 0", "Threshold for the tally trigger");
+  params.addParam<std::vector<bool>>(
+      "trigger_ignore_zeros",
```
aprilnovak, line 52: My gut reaction is to assign a default to everything, so that there's kind of a suggestion to beginner users on what these parameters typically are.

aprilnovak: Then we also don't need to check `isParamValid`, because it will have a value even if the user doesn't set it

nuclearkevin: Sounds good. I've added a default of `false` to `trigger_ignore_zeros` and added functionality to allow users to specify a single value of `trigger_ignore_zeros` which is than applied to all tally scores.

src/tallies/TallyBase.C
```diff
@@ -139,9 +146,17 @@ TallyBase::TallyBase(const InputParameters & parameters)
...
+    {
+      if (_tally_score.size() != _trigger_ignore_zeros.size())
+        mooseError("'trigger_ignore_zeros' (size " + std::to_string(_trigger_ignore_zeros.size()) +
+                   ") must have the same length as 'score' (size " +
+                   std::to_string(_tally_score.size()) + ")");
+    }
+    else if (_trigger_ignore_zeros.size() == 1)
+      _trigger_ignore_zeros.resize(_tally_score.size(), _trigger_ignore_zeros[0]);
+    else
+      mooseError("'trigger_ignore_zeros' has been set with no value(s)!");
```
aprilnovak, line 159: we can use `checkEmptyVector` here

src/tallies/TallyBase.C
```diff
@@ -329,7 +344,7 @@ TallyBase::applyTriggersToLocalTally(openmc::Tally * tally)
     for (int score = 0; score < _tally_score.size(); ++score)
       tally->triggers_.push_back({_openmc_problem.triggerMetric((*_tally_trigger)[score]),
                                   _tally_trigger_threshold[score],
```
aprilnovak, line 346: I may have missed this in an earlier PR, but what happens if the user does not set the `trigger_threshold`? In the constructor, a value is only put into it if the user provided it, but here it's used regardless. Is there a loophole here which would result in a segfault/something else going wrong?

nuclearkevin: I don't believe it's possible for this to segfault as `_tally_trigger_threshold` is default constructed with a length of zero. If the user doesn't set `trigger_threshold` then the length of `_tally_trigger_threshold` will not equal the length of `_tally_score`, so an error will get thrown here:
```
if (_tally_trigger_threshold.size() != _tally_score.size())
      mooseError("'trigger_threshold' (size " + std::to_string(_tally_trigger_threshold.size()) +
                 ") must have the same length as 'score' (size " +
                 std::to_string(_tally_score.size()) + ")");
```

The error is a little unintuitive since the user might not know that they need to also provide a value for  `trigger_threshold`, so I've added a `paramError` which should make it easier for the user to figure out what they need to set to use triggers.

src/tallies/TallyBase.C
```diff
@@ -124,10 +131,13 @@ TallyBase::TallyBase(const InputParameters & parameters)
...
                "'trigger_threshold'. You have specified only one.");
 
-  bool has_tally_trigger = false;
   if (_tally_trigger)
   {
-    _tally_trigger_threshold = getParam<std::vector<Real>>("trigger_threshold");
+    if (isParamValid("trigger_threshold"))
+      _tally_trigger_threshold = getParam<std::vector<Real>>("trigger_threshold");
+    else
+      paramError("trigger_threshold",
```
aprilnovak, line 139: please use `checkRequiredParam`

PR #957, opened 2024-09-18
Allow the use of mixed CSG-DAGMC models when using the `MOABSkinner`

nuclearkevin:
This PR allows for the use of mixed geometry models (CSG + DAGMC) when skinning with the `MOABSkinner`, with a few caveats:
- CSG cells cannot be mapped to the MOOSE mesh mirror when using a skinner. This is to prevent the generation of erroneous DAGMC geometry which may overlap with pre-existing CSG geometry in the OpenMC model. This restriction will be relaxed in a future PR by modifying the `MOABSkinner`.
- The OpenMC geometry can only contain a single DAGMC universe, otherwise the universe to overwrite with skinned geometry cannot be automatically detected.
- The DAGMC universe can only be contained within a single CSG cell, and that cell must exist exclusively in the root universe. This avoids the duplication of DAGMC geometry as this duplicated geometry may not be skinned. Similarly, the DAGMC geometry is not allowed in lattices when using a  `MOABSkinner`.


nuclearkevin:
Found some undefined behaviour when re-initializing the CSG geometry. Marking this PR as a draft until I figure out what's going on and resolve the issue.

nuclearkevin:
Putting a note here for later: the main CSG skinning test runs and converges to the correct output on all machines tested on (with an increasing number of particles). However, it is non-deterministic and so the test itself fails on csvdiff. This should be investigated in the future.

PR #959, opened 2024-09-19
add list of sideset IDs to assembly flow mesher

dshaver-ANL:
This adds a bulleted list of the sideset IDs for the assembly flow mesher documentation.


aprilnovak:
Added a commit to change formatting. The website looked like this (it thought your second dashes were new bullet points), so I've fixed that.


Screen Shot 2024-09-19 at 3 56 46 PM

PR #962, opened 2024-09-26
Fix the generation of the cell to element map

nuclearkevin:
This PR fixes an improper generation of the cell to element map for the case where a subset of elements in the MOOSE mesh do not map to any cells in the OpenMC geometry, but a `CellTally` and temperature/density feedback is requested by the user. This PR also adds a new test to cover this case.

@aprilnovak


aprilnovak:
FYI @snschune this will fix the mapping issue you discovered. I will merge as soon as tests pass.

PR #964, opened 2024-09-30
Enable Adaptive Mesh Refinement

nuclearkevin:
This PR enables adaptive mesh refinement for the OpenMC mesh mirror. This allows for the use of AMR (as implemented in MOOSE) within Cardinal. 

The main changes include looping over active elements instead of all elements when generating the cell -> element map and creating a copy of the mesh containing only active elements when using a `MeshTally`. The latter is required for two reasons: to avoid errors which are thrown by MOOSE because of the equation system created in the `openmc::LibMesh` class, and to prevent erroneous tallying on non-active elements / tally bin scrambling in the refinement/coarsening hierarchy. Creating a copy of the mesh without the refinement hierarchy is a severe limitation of the current approach and should be remedied in the future.

As an added bonus, the changes made in this PR also made it trivial to add block restriction for `MeshTally` objects. When `blocks` is set by the user, a submesh is created which only contains the requested tally blocks which is used by OpenMC for tallying. This is useful for cases where multiphysics feedback is required for the full mesh mirror, but the user only wishes to tally on a subset of the OpenMC geometry. Previously `MeshTally` objects would use the full mesh mirror, resulting in many (potentially unnecessary) tally bins which degrades performance in active batches.


nuclearkevin:
For some reason the AMR MeshTally test crashes in CIVET, but doesn't crash on Pinchot or my local machine. I've ran the test through gdb and Valgrind (thorough memtest) and it executes successfully, so I'm at a bit of a loss at the moment.

Since I use the MOOSE Conda environment on both my local machine and Pinchot (but CIVET builds libMesh from a specific commit hash), I'm starting to wonder if there is an incongruence between the version of libMesh in the Conda distribution and the one required for Cardinal's MOOSE submodule. I'm building from source to investigate that at the moment, hopefully it yields a simple fix so I can get the tests passing.

nuclearkevin:
@aprilnovak I finally managed to track down and patch the bug that's been plaguing this PR; it should be good for review now.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2403,12 +2406,9 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
 
         // we need to then re-establish the data structures that map from OpenMC cells to the [Mesh]
         // (because the cells changed)
+        resetTallies();
         setupProblem();
       }
-#else
```
aprilnovak, line 2408: Can you explain where this went?

test/tests/neutronics/adaptivity/tests
```diff
@@ -0,0 +1,36 @@
+[Tests]
+  [adaptive_cell]
+    type = CSVDiff
+    input = cell.i
+    csvdiff = cell_out.csv
+    requirement = "The system shall allow problems which contain adaptivity on the mesh mirror for cell tallies."
```
aprilnovak, line 6: What does this test show exactly? Can we see the outlines of a CSG cell become more refined as the mesh changes?

aprilnovak:
Would you please also add a news markdown file to announce this capability? I think it warrants it!

src/base/OpenMCCellAverageProblem.C
nuclearkevin: This `#else` statement was the cause of the segmentation faults on CIVET, and more generally it resulted in memory leaks when Cardinal is built without DAGMC.

When the mesh is changing OpenMCCellAverageProblem needs to reset tallies on every timestep (after the first step), and so it calls `resetTallies()` at the beginning of the step. This is immediately followed by a call to `setupProblem()`. Afterwards, if the mesh is moving `setupProblem()` is called a second time in that `#else` statement resulting in an initialization of OpenMC tallies twice for each time they're reset. This causes a memory leak as we don't manage the extra OpenMC tally created on each step.

The reason why this caused a segmentation fault is a little more complicated and involves the copy of the mesh created in `MeshTally` containing only active elements. This mesh copy is deleted both when `resetTally()` and `initializeTally()` are called (it's a unique pointer). Since we've been creating two OpenMC tallies per timestep (by accident) the first of those tallies still thinks the mesh copy is valid, and so it tries to tally on datastructures that have had their memory freed. This behaviour is undefined and results in a segmentation fault.

This bug only occurs when Cardinal is compiled without DAGMC as `resetTallies()` is called again when skinning the geometry (the previous usecase for the `#ifdef`/`#else` statement), so nothing breaks.

test/tests/neutronics/adaptivity/tests
nuclearkevin: This test ensures that problems can run mesh adaptivity when using cell tallies, so it's measuring the increasing number of elements. It's really just a surrogate for a `RunApp` executioner with some extra information attached.

nuclearkevin:
@aprilnovak I can definitely do that! Should I also add an entry for the filter system (since it's a pretty big addition to Cardinal as well)?

src/tallies/MeshTally.C
```diff
@@ -82,6 +98,30 @@ MeshTally::MeshTally(const InputParameters & parameters)
...
+    {
+      auto block_names = getParam<std::vector<SubdomainName>>("blocks");
+      if (block_names.empty())
+        mooseError("Subdomain names must be provided if using 'blocks'!");
+
+      auto block_ids = _mesh.getSubdomainIDs(block_names);
+      std::copy(
+          block_ids.begin(), block_ids.end(), std::inserter(_tally_blocks, _tally_blocks.end()));
+
+      // Check to make sure all of the blocks are in the mesh.
```
aprilnovak, line 113: When restricting tallying only to certain blocks, do the results still get mapped back to the [Mesh] properly? Originally, we would just map element 4 to element 4, etc. Do your tests cover the scenario where the blocks we want to tally are not the "first" in the mesh?

test/tests/neutronics/adaptivity/cell.i
```diff
@@ -1,13 +1,14 @@
...
-    # Mesh of a single pebble with outer radius of 1.5 (cm)
     type = FileMeshGenerator
-    file = ../../neutronics/meshes/sphere.e
+    file = ../meshes/sphere.e
   []
   [solid]
     type = CombinerGenerator
     inputs = sphere
-    positions = '0 0 0'
+    positions = '0 0 0
```
aprilnovak, line 9: Can we make these tests a bit smaller? Like 1 pebble? Some of the output files are 0.5 M.

src/tallies/MeshTally.C
nuclearkevin: Both test cases added in this PR covering block restriction check this. We block restrict such that tallying doesn't happen on the first pebble (which is the first set of elements in the mesh), and so the elements in the created submesh need to get mapped back to the full mesh. To confirm this is happening correctly I compared the results of these tests to an equivalent where the mesh mirror only contains two pebbles.

aprilnovak: perfect, thanks!

test/tests/neutronics/adaptivity/cell.i
nuclearkevin: I can shrink the adaptivity tests down to one pebble and drop the block restriction tests to two pebble. That should help with the PR size.

nuclearkevin:
@aprilnovak dropping the adaptivity tests to 1 pebble decreases the size of the adaptivity Exodiff tests to ~0.5 M in total. The block restriction tests is now ~0.1 M.

PR #966, opened 2024-10-04
Add common reaction rate scores

### LINKED ISSUE #912: Add `total`, `scatter`,`absorption`, `fission` scores to Cardinal tallies
## Reason
We may need these scores in order to complete a multiphysics benchmark.

## Design
Add these scores to the `TallyScoreEnum` in Cardinal. The pattern to use these scores will be identical to the `H3-production` score, as they have the same units. We may want to add a function in `OpenMCCellAverageProblem` which will tell us the units of a tally so that the code complexity on how to normalize these tallies doesn't get out of hand now that we're supporting more and more scores.

## Impact
New feature, more options for tallies.

nuclearkevin:
This PR adds some common reaction rate scores (total, scatter, absorption and fission). These are useful for solving benchmark problems, comparing Cardinal results with other transport codes and computing multi-group cross sections (to name a few uses).

In addition to the above, I believe the total and fission reaction rate scores will be useful for the NEUP AMR project (notably to use as indicators with additional post-processing).

Closes #912.


test/tests/neutronics/tallies/fission/openmc.i
```diff
@@ -0,0 +1,92 @@
+[Mesh]
```
aprilnovak, line 1: Thoughts on perhaps consolidating these three tests into a single one? Could be easier to maintain in the future, since I don't anticipate any different behavior b/w these scores which we would want to modularize.

nuclearkevin:
Assuming there aren't any other changes you want me to make, this should be good to go @aprilnovak.

PR #970, opened 2024-10-19
Output tally relative error

nuclearkevin:
This PR adds the unrelaxed relative error of tally scores as an output for `CellTally` and `MeshTally` objects. Previously this could be computed by dividing the unrelaxed tally standard deviation by the unrelaxed tally mean in the input file (using a `ParsedAux` or similar auxkernel) - this addition avoids the need for extra input file syntax. A series of tests for additional tally outputs are also added.

In addition to the above, this PR also fixes a bug involving an interaction between the filter system and heating scores Cardinal adds automatically (if the user specifies a heating score for source rate normalization when no tallies contain the score). `TallyBase` wasn't creating variables for external filter bins, which resulted in errors when extracting tallies for this case.


test/tests/neutronics/output/tests
```diff
@@ -0,0 +1,51 @@
+[Tests]
+  [unrelaxed_tally_cell]
+    type = CSVDiff
+    input = cell.i
+    csvdiff = cell_out.csv
+    requirement = "The system shall allow for the output of the unrelaxed tally value for cell tallies."
```
aprilnovak, line 6: I think we already have tests covering the output of the raw tally values and standard deviations on both cell and mesh tallies. Can you please check, and if we do, remove these duplicative tests?

nuclearkevin:
@aprilnovak The bug was introduced in the addition of the filter system (#951), it wasn't in Cardinal before that PR. 

The bug itself only occurs when an eigenvalue normalization score is requested by the user but hasn't been added a tally in the `[Tallies]` block. In this case Cardinal attempts to add that eigenvalue normalization score, but when one of the tallies contains external filters `TallyBase:addScore(...)` wouldn't add all of the required variables for the filter bins. The worst-case scenario is that the derived tallies would go out of bounds on array accesses when extracting tallies.

test/tests/neutronics/output/tests
nuclearkevin: I found tests for raw tally values and tally standard deviations for both; I'll get rid of the duplicates. 

Noting this as a future reference for myself and anyone else who plans on adding new derived tally objects: the unrelaxed tally output tests are in `tests/neutronics/relaxation/*`, the mesh tally standard deviation test is in `tests/neutronics/mesh_tally/fission_tally_std_dev.i`, and a test for cell tally standard deviations is in `tests/neutronics/tally_system/multi_diff.i` (using `cli_args` to change the output).

PR #975, opened 2024-10-25
Update moose, use new return code

loganharbour:


aprilnovak:
@loganharbour do you know what all these "Intermittent failures" are? They started appearing after I added a git submodule in the `tests` folder, but I'm not sure if that's related or not.

I'll merge since tests pass, but just curious what this means.

PR #977, opened 2024-10-27
Warn the user if the DAGMC geometry doesn't correspond to the mesh mirror

nuclearkevin:
This PR adds a check to `OpenMCCellAverageProblem` which tells the user how many DAGMC cells map to the mesh mirror when the problem is being skinned. This should aid the user in detecting situations where their OpenMC geometry is getting erroneously deleted by the MOABSkinner. 

There are situations where some DAGMC cells can be replaced by the skinned geometry without causing harm (such as graveyard / bounding void regions) so I don't think this should be a full-blown `mooseError()`; a `mooseWarning()` should suffice for model debugging.

As an example, this is the warning generated for the Krusty tutorial when the fueled region is the mesh mirror and the entire Krusty reactor is the initial DAGMC geometry:
```
Your DAGMC geometry contains unmapped cells! The skinner assumes that the DAG geometry used in the 
OpenMC model maps one to one to the mesh mirror; if that is not the case the skinner may delete some 
parts of your OpenMC model when the underlying geometry is regenerated. You have 57 unmapped DAGMC 
cells out of 58 DAGMC cells.
```

### DIFF
--- a/doc/content/source/filters/FromXMLFilter.md
+++ b/doc/content/source/filters/FromXMLFilter.md
@@ -7,7 +7,7 @@ in the `[Problem]` block. Otherwise, attempting to add this filter will result i
 ## Description
 
 The `FromXMLFilter` object provides a thin wrapper around an arbitrary OpenMC filter which
-has been added to the problem through the problem through the `tallies.xml` OpenMC input
+has been added to the problem through the `tallies.xml` OpenMC input
 file. `FromXMLFilter` fetches the filter through a filter id provided in `filter_id` which
 must match the id of a filter in `tallies.xml`. The bins of this filter are labelled with
 a string provided in `bin_label`.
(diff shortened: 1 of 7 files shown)


aprilnovak:
is there a way to tell if the one cell which has not mapped is a graveyard cell? Just wondering if we can possibly hide this warning if the behavior is indeed as intended, to avoid alarming users

nuclearkevin:
@aprilnovak good point, I've modified the check so it excludes unmapped void regions (which includes the graveyard). I tried finding a way to only exclude the graveyard region but it seems that multiple DAGMC cells can correspond to the graveyard region (depending on how regions are defined in DAGMC) and so the DAGMC ids associated with these cells don't map to the DAGMC graveyard id.

This change removes the warning for the Tokamak tutorial and slightly modifies the warning for the erroneous Krusty reactor case:
```
Your DAGMC geometry contains unmapped cells! The skinner assumes that the DAG geometry used in the 
OpenMC model maps one to one to the mesh mirror; if that is not the case the skinner may delete some parts 
of your OpenMC model when the underlying geometry is regenerated. You have 54 unmapped DAGMC cells 
out of 58 DAGMC cells.
```

PR #990, opened 2024-11-23
Update OpenMC

nuclearkevin:
This PR updates the OpenMC submodule and modifies `MeshTally` to take advantage of changes made in OpenMC to improve adaptive mesh support.

### DIFF
--- a/src/tallies/MeshTally.C
+++ b/src/tallies/MeshTally.C
@@ -148,8 +148,9 @@ MeshTally::spatialFilter()
     {
       _active_to_total_mapping.clear();
 
-      auto begin = _tally_blocks.size() > 0 ? msh->active_subdomain_set_elements_begin(_tally_blocks)
-                                            : msh->active_elements_begin();
+      auto begin = _tally_blocks.size() > 0
+                       ? msh->active_subdomain_set_elements_begin(_tally_blocks)
+                       : msh->active_elements_begin();
       auto end = _tally_blocks.size() > 0 ? msh->active_subdomain_set_elements_end(_tally_blocks)
                                           : msh->active_elements_end();
       for (const auto & old_elem : libMesh::as_range(begin, end))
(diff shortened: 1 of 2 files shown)


src/tallies/MeshTally.C
```diff
@@ -121,12 +121,6 @@ MeshTally::MeshTally(const InputParameters & parameters)
           paramError("blocks",
                      "Block '" + block_names[b] + "' specified in 'blocks' not found in mesh!");
     }
-    else
```
aprilnovak, line 124: Do we not want to keep this default behavior? If there are a lot of blocks, is this requiring the user to type out all the block names?

src/tallies/MeshTally.C
```diff
@@ -144,34 +138,41 @@ MeshTally::spatialFilter()
   // Create the OpenMC mesh which will be tallied on.
   if (!_mesh_template_filename)
   {
+    auto msh = dynamic_cast<const libMesh::ReplicatedMesh *>(_mesh.getMeshPtr());
+    if (!msh)
+      mooseError("Internal error: The mesh is not a replicated mesh.");
```
aprilnovak, line 143: Would this ever happen? I thought we had checks elsewhere that make sure the mesh is not distributed. Just wondering if this could catch some other error or if it is redundant.

src/tallies/MeshTally.C
nuclearkevin: I changed how `MeshTally` performs the block check so filling `_tally_blocks` is no longer required when tallying over all blocks. Instead, `MeshTally` checks if `_tally_blocks` is empty, and if so the full mesh is used to initialize the mesh tally.

src/tallies/MeshTally.C
nuclearkevin: This is a redundant check. It shouldn't happen, I just have it here in case something changes in libMesh (they add another derived ReplicatedMesh equivalent or something along those lines) to make debugging less painful.

PR #995, opened 2024-12-02
Update OpenMC

nuclearkevin:
Updating the OpenMC submodule fixes an adaptivity bug which occurs when running with MPI; the libMesh tally wasn't constructing the bin to element indirection map correctly for adaptive meshes. See https://github.com/openmc-dev/openmc/pull/3206 for more details.


aprilnovak:
can you try re-triggering the tests? The last remnants of pyne should be removed from civet_recipes. If re-triggering doesn't work, let's try closing and opening a new PR as a last resort.

nuclearkevin:
@aprilnovak it looks pyne still isn't working here. I'll close this PR and open a new one (hopefully the third time's the charm) to see if that works.

PR #997, opened 2024-12-02
Update OpenMC

nuclearkevin:
Updating the OpenMC submodule fixes an adaptivity bug which occurs when running with MPI; the libMesh tally wasn't constructing the bin to element indirection map correctly for adaptive meshes. See https://github.com/openmc-dev/openmc/pull/3206 for more details.


nuclearkevin:
@aprilnovak looks like it's still broken even on a new PR. I wonder if DAGMC is setting `BUILD_UWUW=on` regardless of what is getting passed to `-DBUILD_UWUW` by Cardinal's makefiles. Does Cardinal build DAGMC with CI tests?

aprilnovak:
Cardinal does not build DAGMC with CI tests. It looks like DAGMC is still pulling pyne somewhere...

PR #1001, opened 2024-12-04
Control Power and Source Strength

### LINKED ISSUE #1000: Control power and source strength
## Reason
To facilitate load-following Cardinal simulations more easily, the [Controls System](https://mooseframework.inl.gov/syntax/Controls/index.html) can be used to allow simulations with varying power.

## Design
The  `power` and `source_strength` parameters from [OpenMCCellAVerageProblem](https://cardinal.cels.anl.gov/source/problems/OpenMCCellAverageProblem.html) will be changed to controllable.

## Impact
Users will still be able to use `power` and `source_strength` as before, this will just add a new use for them.

**Additional context**
I will make a PR to add this instead of following up on #961.

lewisgross1296:
Closes #1000 (woohoo 🎉 🎉1000 issues!!)

### DIFF
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -376,10 +376,10 @@ protected:
   const bool & _verbose;
 
   /// Power by which to normalize the OpenMC results, for k-eigenvalue mode
-  const Real * _power;
+  const Real & _power;
 
   /// Source strength by which to normalize the OpenMC results, for fixed source mode
-  const Real * _source_strength;
+  const Real & _source_strength;
 
   /**
    * Whether to take the starting fission source from iteration \f$n\f$ as the
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -118,7 +118,7 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
       if (tally_actions.size() > 0)
       {
         checkRequiredParam(params, "power", "running in k-eigenvalue mode");
-        _power = &getPostprocessorValue("power");
+        _power = getPostprocessorValue("power");
       }
       else
         checkUnusedParam(params, "power", "no tallies have been added");
@@ -131,7 +131,7 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
       if (tally_actions.size() > 0)
       {
         checkRequiredParam(params, "source_strength", "running in fixed source mode");
-        _source_strength = &getPostprocessorValue("source_strength");
+        _source_strength = getPostprocessorValue("source_strength");
       }
       else
         checkUnusedParam(params, "source_strength", "no tallies have been added");
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -70,6 +70,9 @@ OpenMCProblemBase::validParams()
       false,
       "Whether to skip writing any statepoint files from OpenMC; this is a performance "
       "optimization for scenarios where you may not want the statepoint files anyways");
+
+  params.declareControllable("power");
+  params.declareControllable("source_strength");
   return params;
 }
 


lewisgross1296:
@gonuke @pshriwise 
Any idea why pyne is failed due to (unrelated) changes in OpenMCCellAverageProblem?
```
-- Submodule update
Submodule 'src/pyne/pyne' (https://github.com/pyne/pyne.git) registered for path 'src/pyne/pyne'
Cloning into '/data/civet0/build/cardinal/contrib/DAGMC/src/pyne/pyne'...
fatal: unable to access 'https://github.com/pyne/pyne.git/': Could not resolve host: github.com
fatal: clone of 'https://github.com/pyne/pyne.git' into submodule path '/data/civet0/build/cardinal/contrib/DAGMC/src/pyne/pyne' failed
Failed to clone 'src/pyne/pyne'. Retry scheduled
Cloning into '/data/civet0/build/cardinal/contrib/DAGMC/src/pyne/pyne'...
fatal: unable to access 'https://github.com/pyne/pyne.git/': Could not resolve host: github.com
fatal: clone of 'https://github.com/pyne/pyne.git' into submodule path '/data/civet0/build/cardinal/contrib/DAGMC/src/pyne/pyne' failed
Failed to clone 'src/pyne/pyne' a second time, aborting
CMake Error at CMakeLists.txt:38 (message):
  git submodule update --init --recursive failed with 1, please checkout
  submodules


-- Configuring incomplete, errors occurred!
-- Cache values
CMAKE_BUILD_TYPE:STRING=Release
CMAKE_INSTALL_PREFIX:PATH=/data/civet0/build/cardinal/install
GIT_SUBMODULE:BOOL=ON
make: *** [/data/civet0/build/cardinal/config/dagmc.mk:3: /data/civet0/build/cardinal/build/DAGMC/Makefile] Error 1

ERROR: exiting with code 2
```

gonuke:
This may be related to scaling/DAGMC#968 that was driven by a request from @nuclearkevin

gonuke:
Notably - CIVET doesn't have access to the outside world. So something more substantial must have changed than you think.

lewisgross1296:
It does look like some of the other actions failed for other reaons. I'll look into those too

nuclearkevin:
@lewisgross1296 The failures in Test, Test DAGMC, and Documentation are caused by an update to the CIVET test recipes that Cardinal uses (which removed pyne as a Cardinal dependency since we don't use the UWUW). As CIVET no longer initializes DAGMC's pyne submodule during the cloning phase, DAGMC attempts to clone it when building. This causes errors since the CIVET build boxes don't have internet access while compiling code, causing git failures when attempting to clone pyne. Once the change @gonuke made (which only clones pyne if UWUW is enabled with a compile flag) in https://github.com/svalinn/DAGMC/pull/968 is merged, we'll update Cardinal's DAGMC submodule which will fix these test failures. 

The other failures you're encountering are compiler errors. You're attempting to initialize a constant reference outside of the `OpenMCCellAverageProblem` initializer list, which is forbidden in C++. Moving the assignments to the initializer list should resolve those.

lewisgross1296:
I finally got this to compile, but wanted to ask for feedback.

We are changing `_power` and `_source_strength` to be declared as `const` references in `include/base/OpenMCProblemBase.h`. As pointers, the existing `devel` branch assigns `_power` or `_source_strength` in the `src/base/OpenMCProblemBase.C` constructor dependent on the value of `_run_mode` (and checks to make sure only one of the two is assigned in the `[Problem]` block).
https://github.com/neams-th-coe/cardinal/blob/17d18e0383db6c333a33ba4fa5a069dc87b26776/src/base/OpenMCProblemBase.C#L110-L134

So now these assignments need to be removed from the OpenMCProblemBase constructor and converted to list initialization.  However they are conditioned on  `_run_mode`, but `_run_mode` doesn't get set until inside the constructor of OpenMCProblemBase. To get around that, I initialize as such
```
OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
  : CardinalProblem(params),
    PostprocessorInterface(this),
    _verbose(getParam<bool>("verbose")),
    _power(openmc::settings::run_mode == openmc::RunMode::EIGENVALUE
               ? getPostprocessorValue("power")
               : 0),
    _source_strength(openmc::settings::run_mode == openmc::RunMode::FIXED_SOURCE
                         ? getPostprocessorValue("source_strength")
                         : 0),
    _reuse_source(getParam<bool>("reuse_source")),
    _specified_scaling(params.isParamSetByUser("scaling")),
    _scaling(getParam<Real>("scaling")),
    _skip_statepoint(getParam<bool>("skip_statepoint")),
    _fixed_point_iteration(-1),
    _total_n_particles(0)
```

Do others have opinions on setting the values to `0` dependent on run mode? I removed the assignments from the constructors, but kept all the other checks, so it should behave the same as before.

lewisgross1296:
So it looks like despite compiling, there's still a larger issue. Apparently `OpenMCCellAverageProblem::power` is of type `PostprocessorName`, which is not allowed to be controllable.
```
non-controllable type 'PostprocessorName' for parameter 'Problem/power' marked controllable: non-controllable type 'PostprocessorName' for parameter '' marked controllable
```
At this point, I think this is a MOOSE design question. Tagging @loganharbour for input

loganharbour:
Yeah, this isn't how controllable parameters work. To control something, it needs to be a value. Not a name that points to something else. I might need to understand the goal a bit more.

lewisgross1296:
As of right now `OpenMCCellAverageProblem` has a parameter named `power` (how the user specifies the power used for tally normalization) and OpenMCCellAverageProblem stores it as Postoprocessor. I can't think of a reason that it needs to stay as a Postprocessor, though there may be some. I wanted to make it controllable so I can do load-following simulations with Cardinal. Should we re-think the approach for load-following or can we change power to be something that is meant to be controllable?

aprilnovak:
`power` used to just be a `const Real`, but we modified it a few years ago to be a postprocessor so that it could be more easily set by another multiapp (e.g., point kinetics) for transients. See: https://github.com/neams-th-coe/cardinal/pull/387

It's possible that we can achieve the same thing with a redesign -- I'm not married to postprocessors if there's a way to do this solely with controls.

nuclearkevin:
@lewisgross1296 A rather clunky workaround which will allow you to specify a controllable value of `power`/`source_strength` would be to create a [ConstantFunction](https://mooseframework.inl.gov/source/functions/ConstantFunction.html) where you specify the power in `value` (which is controllable). You could then pass this to a [FunctionValuePostprocessor](https://mooseframework.inl.gov/source/postprocessors/FunctionValuePostprocessor.html), which you then use for `power` in your `OpenMCCellAverageProblem`.

I haven't tried this myself so I can't guarantee it'll work, but it's worth a shot.

lewisgross1296:
Sorry, I haven't had a chance to try it yet, but I can let you know when I do.

PR #1006, opened 2025-01-03
Point location with NekRS

aprilnovak:
This adds a postprocessor which exposes NekRS's `pointLocation`, adding an equivalent of MOOSE's `PointValue`, but using NekRS's actual solution and mesh.


AyaHegazy22 (approved):
The PR enhances the postprocessing of Cardinal, allowing the interpolation of NekRS solution onto points that do not have to be GLL points with spectral accuracy. 
The implementation is well-structured, and the tests ensure its robustness.

aprilnovak:
Ignoring VTB test failure for now since it is unrelated.

PR #1008, opened 2025-01-04
Add more postprocessors to facilitate Nek QA

### LINKED ISSUE #1015: Add the square of velocity to field
## Reason
Some NekRS CI tests require computing `\int V_iV_i d\Omega`, which we can do through postprocessors if we add the square of the velocity components as a `field`.

## Design
Add the square of the velocity components as a `field`.

## Impact
New capability.

aprilnovak:
This PR adds more postprocessors/features to allow porting of Nek's CI suite into Cardinal tests.

- Adds the squares of the velocity components as fields which can be used in any other postprocessor

Closes #1015


aprilnovak:
Ignoring VTB failure because it is unrelated.

PR #1009, opened 2025-01-05
Minor improvements to website docs

aprilnovak:
Reading through the MooseDocs documentation, discovered some neat features which I've added to our website.


aprilnovak:
Merging regardless of VTB test failure, since that is unrelated and the PRs are starting to pile up.

PR #1013, opened 2025-01-07
Update DAGMC (ref. #1012)

### LINKED ISSUE #1012: Update the DAGMC submodule
## Reason
At present, DAGMC attempts to clone PyNE (an optional dependency which is only required when using the UWUW) during the build process if submodules have not been initialized. This interacts poorly with CIVET as INL's build boxes don't have internet access while compiling, so when we attempt to remove PyNE as a Cardinal dependency tests start to fail. The recent release candidate of DAGMC fixes this issue by initializing the PyNE submodule only when `BUILD_UWUW` is set to true, and so updating the DAGMC submodule will allow us to remove PyNE in CIVET recipes.

## Design
Update the DAGMC submodule.

## Impact
Will enable the removal of PyNE from the CIVET recipes.

**Additional context**
N/A

nuclearkevin:
Closes #1012


aprilnovak:
Ignoring VTB tests for now since they are unrelated

PR #1026, opened 2025-01-16
Add a BooleanComboMarker

### LINKED ISSUE #1025: Add a marker which combines other markers using boolean logic
## Reason
When marking elements for refinement in Cardinal we often have to consider a combination of different factors, such as the relative error of tally scores and the gradients between tally variables. At present, the only option for combining marker variables together in MOOSE is the [ComboMarker](https://mooseframework.inl.gov/source/markers/ComboMarker.html) which takes the maximum of the input marker states (refinement = 2, coarsening = 1, doing nothing = 0) and applies the results to an element (often yielding refinement). This is inflexible for most Cardinal + OpenMC use cases where overrefinement has a large statistical / runtime penalty.

## Design
A `BooleanComboMarker` should be designed which can AND / OR together markers. Different combinations should be supported for refinement (where multiple markers are combined based on statistics, tally gradients, etc.) and coarsening (where we normally want to coarsen based on statistics or the element volume). The user should be allowed to select if refinement or coarsening is preferred in the event there is a collision between the two states.

## Impact
Adding a `BooleanComboMarker` increases the flexibility of AMR for neutronics simulations in Cardinal running with mesh tallies. Additionally, it adds another tool to the AMR toolbox for multi-physics problems.

nuclearkevin:
Closes #1025


test/tests/markers/boolean_combo/tests
```diff
@@ -0,0 +1,37 @@
+[Tests]
+  [prioritize_refine_and]
+    type = CSVDiff
+    input = 'prioritize_refine_and.i'
+    csvdiff = 'prioritize_refine_and_out.csv'
+    requirement = 'The system shall be able to prioritize refinement while ANDing markers together.'
+    issues = '#1025'
```
aprilnovak, line 7: For files like this, where the `issues` and `design` are repeated many times, please put them higher up in the Tests spec. Like this, they will be carried to sub-blocks.

```
[Tests]
  issues = ...
  design = ...
[]
```

aprilnovak:
Looks good to me! Do you think it might be useful to have more flexible boolean options? I could envision a world where someone may want to indicate the and/or operators within the "refine" and "coarsen" instructions.

```
    [combo]
      type = BooleanComboMarker
      refine_markers = '(box and box2) or other'
      coarsen_markers = 'box2 or box3 or other'

      # this flag gets disabled
      # boolean_operator = and

      priority = refinement
    []
```

nuclearkevin:
@aprilnovak I looked into it, and after a multi-hour adventure down the rabbit hole of binary expression parsing / evaluation (as the MOOSE framework doesn't contain a BEP library) I decided against adding it due to the explosion of complexity. The behaviour can be emulated by chaining together multiple `BooleanComboMarker`s, though it's a rather clunky workaround and I think I'll eventually bite the bullet and implement a BEP parser for this class.

aprilnovak:
LGTM!

PR #1029, opened 2025-01-16
Add an ElementOpticalDepthIndicator

### LINKED ISSUE #1028: Add an indicator based on an estimate of the per-element optical depth
## Reason
One of the objectives of AMR is to better capture gradients in underlying field variables. In radiation transports, gradients are often driven by changes in the total cross section and so estimates of the optical depth within an element may prove useful as a surrogate for gradient-based error estimators.

## Design
The error indicator shall compute the following estimate of the per-element optical depth:

```math
E = \frac{(\Sigma_{x}\Phi)}{(\Phi)}h
```
where $(\Sigma_{x}\Phi)$ is the reaction rate for reaction $x$, $(\Phi)$ is the scalar particle flux, and $h$ is an estimate of the length of the element (either the maximum vertex separation, minimum vertex separation. or the cube root of the element volume).

## Impact
The addition of an `ElementOpticalDepthIndicator` adds another capability to Cardinal for performing AMR on OpenMC unstructured mesh tallies.

nuclearkevin:
This PR also abstracts some of the necessary function calls for the `ElementOpticalDepthIndicator` into an `OpenMCIndicator` base class to decrease the amount of duplicate code that may be required for future OpenMC-based indicators.

Closes #1028


doc/content/source/indicators/ElementOpticalDepthIndicator.md
```diff
@@ -0,0 +1,30 @@
+# ElementOpticalDepthIndicator
+  id=eodi
+
+!alert note
```
aprilnovak, line 4: If we're going to add something like this, then IMO we should be consistent and do so for every postprocessor, auxkernel, etc. in Cardinal. I would lean towards not having this, just to keep documentation simpler (the user will hit a mooseError anyways). What do you think?

nuclearkevin: I'm on-board with keeping the documentation as simple as possible; I'll remove the note. I added similar notes to the filter/tally system objects, should I remove those as well or should I save them for a review/refactoring PR later down the line?

aprilnovak: I think those ones can stay, since they are defining new syntax it feels like additional explanation is okay there.

src/base/CardinalEnums.C
```diff
@@ -100,6 +100,14 @@ getTallyScoreEnum()
       "flux H3_production total absorption scatter fission");
 }
 
+MooseEnum
```
aprilnovak, line 103: Out of curiosity: is there anywhere in the code base where we might also want to use this? Do we have any error checking restricting lengths to 1, which we might replace with this?

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -0,0 +1,107 @@
...
+ElementOpticalDepthIndicator::ElementOpticalDepthIndicator(const InputParameters & parameters)
+  : OpenMCIndicator(parameters), _h_type(getParam<MooseEnum>("h_type").getEnum<HType>())
+{
+  std::string score = getParam<MooseEnum>("rxn_rate");
+  std::replace(score.begin(), score.end(), '_', '-');
+
+  // Error check to make sure the score is a reaction rate score and to make sure one of the
+  // [Tallies] has added the score and a flux score.
+  if (!_openmc_problem->isReactionRateScore(score))
+    mooseError(
```
aprilnovak, line 59: Let's try and use `paramError` as much as possible. We haven't had good practices using it consistently, but I'd say let's start now. Can you please change `mooseError`s to `paramError`, if it makes sense?

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void
+ElementOpticalDepthIndicator::computeIndicator()
+{
+  Real rxn_rate = 0.0;
+  Real scalar_flux = 0.0;
+
+  for (const auto & var : _rxn_rates)
+    rxn_rate += (*var)[0];
```
aprilnovak, line 86: What is the `[0]` index doing?

src/base/CardinalEnums.C
nuclearkevin: Off the top of my head, I'm also using this in #1031. I haven't gone through the codebase to find where this enum could replace length checks of 1, but I can do that and add the changes to this PR as well.

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -0,0 +1,107 @@
...
+  auto od = scalar_flux < libMesh::TOLERANCE ? 0.0 : rxn_rate / scalar_flux;
+
+  switch (_h_type)
+  {
+    case HType::Min:
+      od /= _current_elem->hmin();
+      break;
+    case HType::Max:
+      od /= _current_elem->hmax();
+    case HType::CubeRoot:
```
aprilnovak, line 100: add a `default` to catch a possible future scenario where we add another option so we don't forget

test/tests/neutronics/indicators/openmc.i
```diff
@@ -0,0 +1,88 @@
...
+  []
+  [combine]
+    type = CombinerGenerator
+    inputs = 'solid1 solid2 solid3'
+    positions = '0 0 0
+                 0 0 4
+                 0 0 8'
+  []
+
+  allow_renumbering = false
```
aprilnovak, line 37: Can this test be made smaller by removing some spheres? Always looking to make as small as possible :)

src/indicators/ElementOpticalDepthIndicator.C
nuclearkevin: The `[0]` index is accessing the 0'th quadrature point in `_rxn_rate`/`_scalar_fluxes` (I just optimized out a loop over all quadrature points as these variables are guaranteed to be CONSTANT MONOMIALS).

test/tests/neutronics/indicators/openmc.i
nuclearkevin: Sure thing!

src/indicators/ElementOpticalDepthIndicator.C
aprilnovak: The variables are guaranteed to be const monomial when they are created by the `OpenMCCellAverageProblem`. In order to protect ourselves from later on when we might change those types (and forget to update here), could you add a check which will error out if the variable types don't match our expectations?

src/base/CardinalEnums.C
nuclearkevin: The only class which does a length check of one on an enum derived from `getTallyScoreEnum()` is `TallyRelativeError`. Should I swap this to a `getSingleTallyScoreEnum()` and remove the length check + associated test?

aprilnovak: Yes, that would be great! The more we can rely on MOOSE to handle the input checking, the better.

src/indicators/OpenMCIndicator.C
```diff
@@ -0,0 +1,90 @@
...
+#include "OpenMCIndicator.h"
+
+#include "libmesh/threads.h"
+
+InputParameters
+OpenMCIndicator::validParams()
+{
+  auto params = Indicator::validParams();
+  std::vector<SubdomainName> everywhere(1, "ANY_BLOCK_ID");
+  params.addParam<std::vector<SubdomainName>>(
```
aprilnovak, line 30: Can you use the `BlockRestrictable` interface instead?

nuclearkevin: The base indicator in MOOSE (`Indicator`) includes the `BlockRestrictable` interface, so `OpenMCIndicator` already uses it. 

The intermediate derived indicators in MOOSE (such as [ElementIndicator](https://github.com/idaholab/moose/blob/next/framework/src/indicators/ElementIndicator.C#L26)) include the specification of `blocks` in their input parameters. I erred on the side of caution and included it here, though I haven't tested to see if its necessary or not. I'll try removing it to see what happens.

nuclearkevin: After checking it appears that the extra `block` input parameter is redundant. The adaptivity system in MOOSE likely needs a code review at some point...

aprilnovak: Good to know! Can you make a note of that as something to see if it can be pushed upstream to MOOSE (to remove redundant input parameters).

nuclearkevin: Will do!

src/indicators/OpenMCIndicator.C
```diff
@@ -0,0 +1,86 @@
...
+  return params;
+}
+
+OpenMCIndicator::OpenMCIndicator(const InputParameters & parameters)
+  : Indicator(parameters),
+    _openmc_problem(dynamic_cast<OpenMCCellAverageProblem *>(&_subproblem)),
+    _field_var(_subproblem.getStandardVariable(_tid, name())),
+    _current_elem(_field_var.currentElem())
+{
+  if (!_openmc_problem)
```
aprilnovak, line 38: Just an idea -- maybe it's time we made an `OpenMCObject` base class which all of these different `OpenMCPostprocessor`, `OpenMCIndicator`, `OpenMCUserObject`, etc. inherit from. What do you think? (something similar would also make sense for the various `Nek` base classes which do similar checks for `NekRSProblemBase`.

nuclearkevin: Agreed, it would decrease code duplication quite a bit and make maintenance easier. Should I get that implemented in here or open a new issue and handle it later?

aprilnovak: I think we can do it in a separate PR. Thank you for implementing it -- we'll want to be sure we keep one of the problem-type-checking tests around, but we should be able to delete all other duplicates. If you could also do the change to NekRS, that'd be great.

I'm ready to merge this PR, once an issue is created to track this side conversation. Thanks!

PR #1033, opened 2025-01-18
Approximating tally gradients with finite differences

### LINKED ISSUE #1031: Add an auxkernel to calculate tally gradients with finite differences
## Reason

AMR is often driven by gradients in field variables. However, traditional mesh tallies score constants within each element which don't have well-defined gradients. This necessitates the use of a gradient approximation, either with finite differences or a functional expansion/projection. At present MOOSE doesn't contain methods to approximate the gradients of CONSTANT MONOMIAL field variables, and so we need to roll our own to feed into gradient jump / magnitude indicators.

## Design

The finite difference gradient estimate can be performed using the approach outlined in  [K. N. Stolte and P. V. Tsvetkov, X-MeRA: Computationally efficient adaptive mesh refinement of Monte Carlo mesh based tallies, Annals of Nuclear Energy, Volume 182, 2023](https://doi.org/10.1016/j.anucene.2022.109617). The gradient approximation should be done in an auxkernel to maximize flexibility.

## Impact

The addition of a `FDTallyGradAux` allows for the approximation of CONSTANT MONOMIAL tally gradients with finite differences, which can then be fed into indicators used for AMR. This adds another capability to Cardinal for performing AMR on OpenMC unstructured mesh tallies.

nuclearkevin:
Closes #1031


doc/content/source/auxkernels/FDTallyGradAux.md
```diff
@@ -0,0 +1,27 @@
+# FDTallyGradAux
+
+## Description
+
+`FDTallyGradAux` approximates tally gradients using forward finite differences between the current element and it's neighbors.
```
aprilnovak, line 5:
```suggestion
`FDTallyGradAux` approximates tally gradients using forward finite differences between the current element and its neighbors.
```

src/auxkernels/FDTallyGradAux.C
```diff
@@ -0,0 +1,147 @@
...
+{
+  auto params = OpenMCVectorAuxKernel::validParams();
+  params.addClassDescription(
+      "An auxkernel which approximates tally gradients at element centroids using "
+      "forward finite differences.");
+  params.addRequiredParam<MooseEnum>(
+      "score",
+      getSingleTallyScoreEnum(),
+      "The tally score this auxkernel should approximate the gradient of.");
+  params.addParam<unsigned int>(
```
aprilnovak, line 38:
```suggestion
  params.addRangeCheckedParam<unsigned int>(
```

src/auxkernels/FDTallyGradAux.C
```diff
@@ -0,0 +1,147 @@
...
+      "ext_filter_bin",
+      0,
+      "The non-spatial filter bin for the tally score (with bin indices starting at 0).");
```
aprilnovak, line 41: Feel free to disregard the suggestions here if they don't make sense -- I seem to recall that the unsigned int type does not catch when users enter negative numbers (they get immediately case to unsigned int). Not sure if MOOSE has changed that behavior, it was a while ago

src/auxkernels/FDTallyGradAux.C
```diff
@@ -0,0 +1,147 @@
...
+
+  params.addRelationshipManager("ElementSideNeighborLayers",
+                                Moose::RelationshipManagerType::ALGEBRAIC |
+                                    Moose::RelationshipManagerType::GEOMETRIC,
+                                [](const InputParameters &, InputParameters & rm_params)
+                                { rm_params.set<unsigned short>("layers") = 2; });
```
aprilnovak, line 47: Can you please explain what this line is doing? What is `layers`?

src/auxkernels/FDTallyGradAux.C
```diff
@@ -0,0 +1,147 @@
...
+FDTallyGradAux::FDTallyGradAux(const InputParameters & parameters)
+  : OpenMCVectorAuxKernel(parameters),
+    _bin_index(getParam<unsigned int>("ext_filter_bin")),
+    _sum_y_y_t(RealEigenMatrix::Zero(3, 3)),
+    _sum_y_du_dy(RealEigenVector::Zero(3))
+{
+  if (_var.feType() != FEType(CONSTANT, MONOMIAL_VEC))
+    paramError("variable",
+               "FDTallyGradAux only supports CONSTANT MONOMIAL_VEC shape functions. Please "
+               "ensure that 'variable' is of type CONSTANT MONOMIAL_VEC.");
```
aprilnovak, line 61:
```suggestion
               "ensure that 'variable' is of type MONOMIAL_VEC and order CONSTANT.");
```

src/auxkernels/FDTallyGradAux.C
```diff
@@ -0,0 +1,147 @@
...
+
+  std::string score = getParam<MooseEnum>("score");
+  std::replace(score.begin(), score.end(), '_', '-');
+
+  // Error check and fetch the tally score.
+  const auto & all_scores = _openmc_problem->getTallyScores();
```
aprilnovak, line 67: Maybe we can bundle this code into `OpenMCProblemBase`? I think the `TallyRelativeError` probably has identical code, which we can reduce.

src/auxkernels/OpenMCAuxKernel.C
```diff
@@ -20,27 +20,107 @@
...
+
+  if (score_vars.size() == 0)
+    mooseError("No tallies contain the requested score " + score + "!");
+
+  return score_vars;
+}
+
+template <typename ComputeValueType>
+std::vector<const VariableValue *>
+OpenMCAuxKernelTempl<ComputeValueType>::getTallyScoreVariableValues(const std::string & score)
```
aprilnovak, line 74: This (and some other functions added here) feels kind of specific to have in `OpenMCAuxKernel`, since not all derived classes deal with tallies. Would it be more technical debt to move this somewhere a bit closer to the tallies?

src/auxkernels/FDTallyGradAux.C
nuclearkevin: `ElementOpticalDepthIndicator` also uses similar logic for error checking so it makes sense to merge all of the score-checking logic together. Since `OpenMCProblemBase` doesn't handle tallies, I think this check should be in `OpenMCCellAverageProblem` (unless we want to add a pure virtual function to `OpenMCProblemBase` which `OpenMCCellAverageProblem`).

src/auxkernels/FDTallyGradAux.C
nuclearkevin: This line ensures we add proper element ghosting to `FDTallyGradAux`. MOOSE partitions fields such that processors all own a different component of a field. The same is also true for element information (such as centroids, nodes, volume, etc.). Flagging `Moose::RelationshipManagerType::ALGEBRAIC` indicates that you want algebraic ghosting; processors which "own" field information for an element will also allocate resources to store field information for that element's neighbourhood of size `layers-1`. Flagging `Moose::RelationshipManagerType::GEOMETRIC` indicates that you want geometric ghosting; processors which "own" an element will also allocate resources to store that element's neighbourhood of size `layers-1`. 

Basically, this line tells MOOSE that we want to be able to access the values of variables and element information in the neighbourhood of the current element we're operating on.

src/auxkernels/OpenMCAuxKernel.C
nuclearkevin: I can add this to `OpenMCCellAverageProblem` - these functions are essentially duplicated in `ElementOpticalDepthIndicator` so this will decrease the overall technical debt.

src/auxkernels/FDTallyGradAux.C
aprilnovak: ah yes, then I think `OpenMCCellAverageProblem` makes sense as the place to put it!

src/auxkernels/FDTallyGradAux.C
aprilnovak: sweet!

aprilnovak: Do you mind checking if we had any unecessary range-checked unsigned int elsewhere in the code base, and add here?

nuclearkevin: It looks like [we didn't have any unnecessary range checks](https://github.com/search?q=repo%3Aneams-th-coe%2Fcardinal%20addRangeCheckedParam&type=code) on `unsigned int` types.

aprilnovak: they're not exactly range checks on unsigned ints, but maybe we can swap over some of these range checks on ints to be unsigned ints without the range check? Like this one?

```
  params.addRequiredRangeCheckedParam<int>(
      "filter_id",
      "filter_id >= 0",
      "The id of the OpenMC filter that this class should provide to Cardinal tallies.");
```

nuclearkevin: After searching all of our input syntax I only found two cases where a range checked `int` was being used inappropriately in input parameters. They've been swapped out for `unsigned int`s now.

src/base/OpenMCProblemBase.C
```diff
@@ -56,10 +56,10 @@ OpenMCProblemBase::validParams()
       "inactive_batches",
       "inactive_batches >= 0",
       "Number of inactive batches to run in OpenMC; this overrides the setting in the XML files.");
-  params.addRangeCheckedParam<int>("particles",
-                                   "particles > 0 ",
-                                   "Number of particles to run in each OpenMC batch; this "
-                                   "overrides the setting in the XML files.");
+  params.addRangeCheckedParam<unsigned int>("particles",
+                                            "particles > 0 ",
```
aprilnovak, line 60: we shouldn't need this line now, correct?

src/tallies/TallyBase.C
```diff
@@ -407,6 +407,15 @@ TallyBase::getTallyID() const
   return getWrappedTally()->id();
 }
 
+int
+TallyBase::scoreIndex(const std::string & score) const
+{
+  if (!hasScore(score))
```
aprilnovak, line 413: In Cardinal, we don't typically use coding of return values to indicate something is not found. Would it make more sense to throw an error here instead? If not, can we make the `-1` into a const expr like `INVALID_SCORE` so that it's clearer what the number signifies?

src/base/OpenMCProblemBase.C
nuclearkevin: We still need the range check here to stop the user from specifying zero particles per batch (which would result in an OpenMC error).

PR #1038, opened 2025-01-22
Add base classes for OpenMC and Nek objects

### LINKED ISSUE #1036: Combine OpenMC/Nek base classes into two single OpenMC/Nek base classes
## Reason
At present Cardinal uses a series of OpenMC / Nek base classes for each MOOSE system (e.g. `OpenMCPostProcessor`, `OpenMCAuxKernel`, etc.). For the sake of long-term maintainability, these should be merged together into a single base class for OpenMC objects and a single base class for Nek objects.

## Design
Merge all of the OpenMC object base classes into a single class and merge all Nek base classes into a single class.

## Impact
Decreases code duplication and improves maintainability of OpenMC / Nek objects.

nuclearkevin:
This is still a WIP. I've been having issues getting Nek running on my machine, so I'm using CIVET to run Nek tests.

Closes #1036


nuclearkevin:
@aprilnovak I've updated this PR to account for the changes made in #923. Once tests pass it should be ready for review.

include/base/OpenMCBase.h
```diff
@@ -18,38 +18,54 @@
...
-  OpenMCIndicator(const InputParameters & parameters);
+  OpenMCBase(const MooseObject * moose_object, const InputParameters & parameters);
 
 protected:
+  /**
+   * Determine whether the MOOSE element maps to an OpenMC cell to make sure we don't call
+   * accessor methods that rely on valid values for the cell instance and index
+   * @return whether element maps to OpenMC
+   */
+  bool mappedElement(const Elem * elem) const;
```
aprilnovak, line 43: I'm not sure if this belongs in the base class -- you can envision a scenario later where we might have delta tracking in OpenMC which query mesh data without needing to set up this elem <--> cell mapping. Same comment for some of the other methods in this class -- I thought that `OpenMCBase` would basically only have the error check that a compatible problem class is being used?

nuclearkevin: Agreed when it comes to `mappedElement(...)`, I'll move that back into the auxkernels. In my opinion I think the other methods (`getTallyScoreVariables(...)`, `getTallyScoreVariableValues(...)`, `getTallyScoreNeighborVariableValues(...)`) have broad applicability to OpenMC objects as there is no restriction on the type of`MooseVariable` fetched; the variable could implement element-averaged tally results (like what we have now) or per-element/global FETs. I also found myself needing these methods for tally gradient computation in #1033 and wanted to avoid adding duplicate code. 

That was my motivation for implementing these in `OpenMCBase`, though If you'd prefer me to revert the change I can revert the removal of `OpenMCIndicator` to add these back to that class.

aprilnovak: I think for now let's revert those changes back -- I don't want to expose all classes in Cardinal being able to access those tally values directly. Let's put a pin in it, if the design gets cumbersome later we can always re-add it.

aprilnovak:
Ok, while helping debug #907 , I've encountered some weird new behavior we need to solve. I have an example test case,  and if we have this in the GlobalParams block, 

```
[GlobalParams]
  displacements = 'd_x d_y d_z'
[]
```

then this line will fail in the problem check

```
  _openmc_problem = dynamic_cast<OpenMCCellAverageProblem *>(&_subproblem);

  if (!_openmc_problem)
    mooseError("This auxkernel can only be used with problems of type 'OpenMCCellAverageProblem'!");
```

It doesn't matter the identity of `d_x`, `d_y`, `d_z`, you can put any dummy variable name in there and this check will fail. I'm really scratching my head with this one.

src/userobjects/OpenMCVolumeCalculation.C
```diff
@@ -61,8 +63,7 @@ OpenMCVolumeCalculation::OpenMCVolumeCalculation(const InputParameters & paramet
   else
     mooseError("Unhandled trigger enum in OpenMCCellVolumeCalculation!");
 
-  auto openmc_problem = openmcProblem();
```
aprilnovak, line 64: Can we eliminate `openmcProblem()` as a function?

nuclearkevin: Looks like I removed it already (this is the removal part of the diff).

aprilnovak: Oh whoops, sorry I got turned around a bit. 

This looks good to me -- LGTM!

PR #1039, opened 2025-01-25
Adding HPC instructions for ANL's Polaris

### LINKED ISSUE #757: HPC instructions for Polaris
## Reason
We will start using Cardinal on Polaris. It will help users to identify which HPC modules they can use.

## Design
Add instructions for Polaris Cardinal builds.

## Impact
New documentation.

nuclearkevin:
The job script successfully runs the pincell multiphysics tutorial case in CPU mode and the instructions build NekRS with CUDA support. I don't have any experience submitting jobs that run on GPUs; feel free to suggest modifications to enable NekRS mixed runs!

Closes #757

### DIFF
--- a/doc/content/hpc.md
+++ b/doc/content/hpc.md
@@ -84,6 +84,73 @@ export OPENMC_CROSS_SECTIONS=$HOME_DIRECTORY_SYM_LINK/cross_sections/endfb-vii.1
 
 !listing scripts/job_bebop language=bash caption=Sample job script for Bebop with the `startup` project code id=bb2
 
+## Polaris
+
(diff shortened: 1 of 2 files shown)


aprilnovak:
@nuclearkevin if you are up for it, what we'd need to run on GPU is a modified version of this script, just with NekRS executables replaced by Cardinal executables and such: https://github.com/Nek5000/nekRS/blob/next/scripts/nrsqsub_polaris

nuclearkevin:
Thanks @aprilnovak! I've modified the job script to take those changes into account - the pincell multiphysics case successfully ran with NekRS's CUDA backend.

aprilnovak:
Thanks @nuclearkevin!

PR #1044, opened 2025-02-01
Nek function norm

### LINKED ISSUE #1043: Allow computing L2 norms on NekRS mesh
## Reason
To port NekRS's CI suite into Cardinal, we need to be able to compute L2 norms of the NekRS solution fields relative to functions (which provide an analytic solution/some other quantity desired to compare against).

## Design
Add a `NekFunctionL2Norm` object to compute the integrated L2 norm of the NekRS solution field relative to a MOOSE function.

## Impact
New capability.

aprilnovak:
closes #1043


aprilnovak:
Closing because this was addressed in #1431

Issue #932, opened 2024-07-15
Add build instructions for Bitterroot


aprilnovak:
## Reason
New machine coming to INL, let's make sure we know how to build Cardinal on it.

## Design
Add Bitterroot as a system to Cardinal's HPC documents.

## Impact
Better user experience.

lewisgross1296:
I was able to ssh into Bitterrot, but upon opening my terminal, the `~/.bashrc` I have for Sawtooth complained
```
Lmod has detected the following error:  The following module(s) are unknown: "openmpi/4.1.6-gcc-12.3.0-panw"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "openmpi/4.1.6-gcc-12.3.0-panw"

Also make sure that all modulefiles written in TCL start with the string #%Module

Lmod has detected the following error:  The following module(s) are unknown: "cmake/3.27.7-gcc-12.3.0-5cfk"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "cmake/3.27.7-gcc-12.3.0-5cfk"

Also make sure that all modulefiles written in TCL start with the string #%Module

Lmod has detected the following error:  The following module(s) are unknown: "gcc/12.3.0-gcc-10.5.0-vx2f"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "gcc/12.3.0-gcc-10.5.0-vx2f"

Also make sure that all modulefiles written in TCL start with the string #%Module
```
seems like those modules exist when I `module avail`, but perhaps the syntax is causing an issue. Maybe I should remove this from my `~/.bashrc`
```
###################### CARDINAL ENVIRONMENT ######################
module purge
module load use.moose
module load moose-tools
module load openmpi/4.1.6-gcc-12.3.0-panw
module load cmake/3.27.7-gcc-12.3.0-5cfk
module load gcc/12.3.0-gcc-10.5.0-vx2f # needed for NekRS
```
Might just be better to load them in the terminal when building Cardinal? (I do get complaints that these modules don't exist when I `scp` or log onto `inlhpclogin`, but I just ignore the messages.

aprilnovak:
[...]

PR #1047, opened 2025-02-04
Adding build instructions for bitterroot

meltawila:
(Ref. #932)


aprilnovak:
Thanks @meltawila! Just confirming, you checked the `logfile`s for the two runs to be sure they ran to completion ok? (It's sometimes possible for Cardinal to compile ok, but for NekRS to throw runtime errors related to build system settings, so just want to confirm we checked).

meltawila:
I checked and NekRS tests seem to be running ok as well as the pin cell multiphysics tutorial

PR #1050, opened 2025-02-11
Adding a reactivity postprocessor.

MendesMichael:
Adding a reactivity postprocessor using the combined k-effective (k_eff - 1)/k_eff.


src/postprocessors/Reactivity.C
```diff
@@ -0,0 +1,52 @@
...
+
+#include "Reactivity.h"
+#include "openmc/eigenvalue.h"
+
+registerMooseObject("CardinalApp", Reactivity);
+
+InputParameters
+Reactivity::validParams()
+{
+  InputParameters params = OpenMCPostprocessor::validParams();
```
aprilnovak, line 27: Great start! I think you can reduce some code duplication if you have your `Reactivity` postprocessor inherit from `KEigenvalue` -- this would also allow you to just call `KEigenvalue::getValue()` to fetch k (and also allow users to choose which estimator for k they would like).

aprilnovak:
To fully add an object, please also add:

- Documentation (`Reactivity.md`) with how to use the object
- You also need to create an issue to link against this PR. 
- Then, in the `tests` file, make sure to use `design` and `issues` to link the test against the code requirements. This is necessary for SQA (Software Quality Assurance) -- you can grep for examples elsewhere in Cardinal.

MendesMichael:
Closes #1051

aprilnovak:
@MendesMichael, you'll need to apply another round of clang format to get the tests to run.

src/postprocessors/Reactivity.C
```diff
@@ -0,0 +1,41 @@
...
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
```
nuclearkevin, line 18:
```suggestion

#ifdef ENABLE_OPENMC_COUPLING

```

src/postprocessors/Reactivity.C
```diff
@@ -0,0 +1,41 @@
...
+Real
+Reactivity::getValue() const
+{
+  // Fetch the k-effective value using the parent class's getValue() method.
+  Real k_effective = KEigenvalue::getValue();
+
+  // Calculate and return the reactivity.
+  return (k_effective - 1.0) / k_effective;
+}
```
nuclearkevin, line 41:
```suggestion
}

#endif
```

nuclearkevin (commented):
@MendesMichael When writing new OpenMC classes you need to include a pair of matching `#ifdef` statements in the `.C` file to ensure that it doesn't get compiled when the user specifies `export ENABLE_OPENMC=false`. This is whats causing `Test nekRS` to fail; that part of the regression suite isn't compiled with OpenMC and so it doesn't know what to do with the reactivity post processor.

src/postprocessors/Reactivity.C
MendesMichael: @nuclearkevin  Thanks!

doc/content/source/postprocessors/Reactivity.md
```diff
@@ -0,0 +1,23 @@
...
+
+!syntax description /Postprocessors/Reactivity
+
+## Description
+
+This postprocessor extracts the combined $k$ eigenvalue from the latest OpenMC eigenvalue
+calculation and calculates reactivity as:
+\begin{equation}
+\label{line-eq2}
+reactivity = \frac{k_eff-1}{k_eff}
```
aprilnovak, line 11:
```suggestion
reactivity = \frac{k-1}{k}
```

aprilnovak: When you do a multi-character subscript in latex, you need to put all characters inside curly braces or else it will only put the first character in the subscript. I also prefer to just write `k` instead of k-effective.

PR #1055, opened 2025-02-14
Skip spurrious OpenMC runs when the mesh isn't changed in adaptivity cycles

### LINKED ISSUE #1053: Skip spurrious OpenMC runs when the mesh isn't changed in adaptivity cycles
## Reason
At present `OpenMCProblemBase`/`OpenMCCellAverageProblem` continue to run OpenMC on every adaptivity cycle, even when adaptivity doesn't modify the mesh. These extra runs should be disabled as they increase simulation times for no good reason.

## Design
Running OpenMC should be skipped if the mesh is unchanged by adaptivity.

## Impact
Improved simulation times in OpenMC problems that contain h-adaptivity on the unstructured mesh tallies.

nuclearkevin:
Closes #1053


src/base/OpenMCCellAverageProblem.C
```diff
@@ -2442,6 +2442,18 @@ OpenMCCellAverageProblem::checkNormalization(const Real & sum, unsigned int glob
 void
 OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
 {
+  // Always run OpenMC on the first timestep in a steady solve. We have to set the flag here
+  // (as opposed to OpenMCProblemBase::externalSolve()) to ensure feedback fields are transferred
+  // to OpenMC.
+  if (timeStep() == 1 && !isTransient())
```
aprilnovak, line 2448: I'm a little concerned we might be missing the general behavior if we do this. MOOSE supports things like Picard iterations (for the other physics, not OpenMC), which maybe could do unexpected things here? 

Can we accomplish the same goals as this PR, but have `_should_run_openmc` always be true unless `_adaptivity.isOn()`, in which case then we enter the more complicated logic?

nuclearkevin: `_should_run_openmc` is set to true by default (in the `OpenMCProblemBase` initializer list);  the only time it's set to false is when the mesh is adapted (and so `_adaptivity.isOn()`). The logic here resets the flag to true in-between `Steady` executioner Picard iterations, and shouldn't be impacted by other physics which are running in Picard loops (unless they're defined in the same input file as the OpenMC simulation, but that will result in an error).

test/tests/neutronics/adaptivity/tests
```diff
@@ -37,3 +37,26 @@
...
   []
 []
+
+[Tests]
+  issues = '#1053'
+  design = 'OpenMCCellAverageProblem.md'
+
+  [skip_openmc_unchanged]
+    type = Exodiff
+    input = unchanged_mesh.i
```
aprilnovak, line 47: So this test is checking that the tally values remain unchanged, and therefore showing that OpenMC is not re-running?

nuclearkevin: Yep, that's correct.

PR #1058, opened 2025-02-18
Refactor to remove duplicate mapping generation in syncSolutions()

### LINKED ISSUE #1057: Refactor OpenMCCellAverageProblem to improve performance when using a skinner
## Reason
When using a skinner `OpenMCCellAverageProblem` doesn't set `_need_to_reinit_coupling` to true. Instead, an extra call to `resetTallies()` and `setupProblem()` (akin to regenerating the mapping twice) was used. Generating the cell to element map is an expensive procedure, and so this resulted in added simulation time on each Picard iteration.

## Design
`OpenMCCellAverageProblem::syncSolutions()` should be refactored such that `_need_to_reinit_coupling` is set to true when using a skinner, and the skinner acts before the cell to element map is regenerated on each Picard iteration. This allows for the removal of the duplicate `resetTallies()` and `setupProblem()` calls.

## Impact
Performance improvements.

nuclearkevin:
Closes #1057

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -192,7 +192,8 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
                              ? false
                              : getParam<bool>("normalize_by_global_tally")),
     _has_adaptivity(getMooseApp().actionWarehouse().hasActions("set_adaptivity_options")),
-    _need_to_reinit_coupling(_has_adaptivity),
+    _using_skinner(isParamValid("skinner")),
+    _need_to_reinit_coupling(_has_adaptivity || _using_skinner),
     _check_tally_sum(
         isParamValid("check_tally_sum")
             ? getParam<bool>("check_tally_sum")
@@ -208,8 +209,7 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
     _needs_global_tally(_check_tally_sum || _normalize_by_global),
     _volume_calc(nullptr),
     _symmetry(nullptr),
-    _initial_num_openmc_surfaces(openmc::model::surfaces.size()),
-    _using_skinner(isParamValid("skinner"))
+    _initial_num_openmc_surfaces(openmc::model::surfaces.size())
 {
   // Check to see if a displaced problem is being initialized.
   // TODO: this also needs to include a "use_displaced_mesh" parameter, alongside the ability to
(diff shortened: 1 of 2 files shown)


src/base/OpenMCCellAverageProblem.C
```diff
@@ -2448,8 +2448,22 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
...
+        // Update the OpenMC materials (creating new ones as-needed to support the density binning)
+        updateMaterials();
+
+        // regenerate the DAGMC geometry
+        reloadDAGMC();
+      }
+#endif
       // re-establish the mapping from the OpenMC cells to the [Mesh], if needed
-      if (!_first_transfer && _need_to_reinit_coupling)
+      if ((!_first_transfer || _using_skinner) && _need_to_reinit_coupling)
```
aprilnovak, line 2466: isn't `_using_skinner` already accounted for in `_need_to_reinit_coupling`?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2448,8 +2448,22 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
   {
     case ExternalProblem::Direction::TO_EXTERNAL_APP:
     {
+#ifdef ENABLE_DAGMC
+      if (_skinner)
```
aprilnovak, line 2452: Does this do anything weird, if we call this on `_first_transfer`, but then just below here we do NOT call `setupProblem` unless `!_first_transfer`?

src/base/OpenMCCellAverageProblem.C
nuclearkevin: Yes, but if we don't account for it here we won't regenerate the mapping on the first Picard iteration after the skinner acts (due to the and statement in the condition).

src/base/OpenMCCellAverageProblem.C
nuclearkevin: The or statement in `if ((!_first_transfer || _using_skinner) && _need_to_reinit_coupling)` avoids this problem.

src/base/OpenMCCellAverageProblem.C
aprilnovak: Ah, gotcha. 

I know the original code had the logic this way, but does anything break if we uniformly only enter this if `!_first_transfer && _need_to_reinit_coupling`? In other words, the behavior for `_first_transfer` is not different when we do/do not have the skinner?

I'm struggling to remember if we had a design reason for the original code, or just hacky code.

nuclearkevin: I believe the intent is for the case where OpenMC is a sub-application which is executed after the parent. If that happens, then there's a temperature / density field available for the skinner on the first transfer (and so the geometry should be updated).

aprilnovak: That makes sense. While we have this PR open, do you mind adding some comments to the code to explain this structure? Otherwise I might forget again :)

PR #1064, opened 2025-02-22
Dimensionalize interface

aprilnovak:
In order to support additional coupling fields (such as tritium transport), we need to improve the interface for dimensionalizing the nek solution. This moves these settings into a `[Dimensionalize]` sub-block.


aprilnovak:
Scratching my head with the doc error here -- @nuclearkevin, since you recently added an action, does anything obvious jump out at you for why the `Problem/Dimensionalize` syntax is not recognized? I've only used actions a handful of times, so I must be missing something obvious.

nuclearkevin:
I honestly have no idea. I've gone over a bunch of examples in the MOOSE framework and can't find any issues with how you've set up the `DimensionalizeAction`. You're missing `syntax/Problem/Dimensionalize/index.md` (which is causing some of the other errors while building docs) but adding that doesn't fix the missing syntax error, unfortunately.

PR #1074, opened 2025-02-26
Add tips and resources for HPC builds.

### LINKED ISSUE #1073: Tips and resources for HPC builds
## Reason
At present Cardinal contains build instructions which make the implicit assumption that you'll be building Cardinal on a non-HPC system or one of the systems we support. We should include a section to provides tips and resources for getting Cardinal built on HPCs which aren't included in the supported HPC section.

## Design
A new section of documentation should be added to introduce Lmod (the most common environment manager for HPC systems), provide links to minimum requirements for Cardinal's dependencies, provide a framework for building Cardinal on login nodes, discuss building with GPU support, and provide a brief introduction to building on compute nodes.

## Impact
Enhances the new user experience by providing a list of resources for building on new HPC systems or systems we don't have instructions for.

nuclearkevin:
Closes #1073

### DIFF
--- a/doc/content/hpc_build_tips.md
+++ b/doc/content/hpc_build_tips.md
@@ -45,8 +45,12 @@ DIRECTORY_WHERE_YOU_HAVE_CARDINAL=$HOME/projects
 HOME_DIRECTORY_SYM_LINK=$(realpath -P $DIRECTORY_WHERE_YOU_HAVE_CARDINAL)
 export NEKRS_HOME=$HOME_DIRECTORY_SYM_LINK/cardinal/install
 
-# These environment variables indicate how many processes you want to build with. Be careful with the number
-# of jobs specified, most HPC systems limit the number of processes you can run with on login nodes.
+# These environment variables indicate how many processes you want to build with. Keep in mind that the
+# more processes you specify the more memory is consumed during a build. Be careful with the number
+# of jobs, most HPC systems limit the number of processes and memory you can consume when
+# on a login node. If you find that your builds are crashing with an exit code of 1 while building,
+# you've likely exceeded either the CPU or memory limit on the machine - reducing the number of processes
+# should mitigate those crashes.
 export JOBS=8
 export LIBMESH_JOBS=8
 export MOOSE_JOBS=8
@@ -126,9 +130,9 @@ HOME_DIRECTORY_SYM_LINK=$(realpath -P $DIRECTORY_WHERE_YOU_HAVE_CARDINAL)
 export NEKRS_HOME=$HOME_DIRECTORY_SYM_LINK/cardinal/install
 
 # Change these to the number of processes you want to build with on a compute node.
-export JOBS=100
-export LIBMESH_JOBS=100
-export MOOSE_JOBS=100
+export JOBS=32
+export LIBMESH_JOBS=32
+export MOOSE_JOBS=32
 
 # You might need to change directories into the location where the job was launched,
 # which should be the Cardinal repo directory.
@@ -139,7 +143,7 @@ cd $PBS_O_WORKDIR
 ./contrib/moose/scripts/update_and_rebuild_wasp.sh > wasp_log.txt
 
 # Change this to the number of processes you want to build with on a compute node.
-make -j100 MAKEFLAGS=-j100 > cardinal_log.txt
+make -32 MAKEFLAGS=-32 > cardinal_log.txt
 ```
 
 You'd then submit your build job script to the HPC queue (for PBS the command is `qsub job_script_name`). We dump the output of each component of the build to different logfiles; this allows us to follow the build in real-time and makes it easier to parse for errors.
--- a/doc/content/hpc_build_tips.md
+++ b/doc/content/hpc_build_tips.md
@@ -63,7 +63,7 @@ The best place to put these exports is your `.bashrc` to avoid needing to export
 
 If critical errors are reported at this stage, it is likely because your current module set does not meet MOOSE's requirements. If you receive an error stating that `jinja2` is missing, you can safely move on.
 
-**Note:** If you want to build NekRS with GPU support, you need GPU-enabled compilers (CUDA compilers, OpenCL compilers, etc.) on the login node. Some HPC systems only allow users to load those modules on nodes which contain GPUs. If that is the case for you, you'll need to build with a job script (see the next session). To enable a GPU build, set one of the following environment variables in Cardinal's makefile to `1` (depending on the specific GPUs on your system): `OCCA_CUDA_ENABLED`, `OCCA_HIP_ENABLED`, or `OCCA_OPENCL_ENABLED`. You'll also need to make sure you load modules with GPU compilers.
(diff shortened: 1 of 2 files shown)


doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+Before you start loading modules, it's best to consult the minimum requirements of MOOSE, NekRS, OpenMC, and DAGMC:
+
+1. MOOSE: [https://mooseframework.inl.gov/getting_started/installation/hpc_install_moose.html](https://mooseframework.inl.gov/getting_started/installation/hpc_install_moose.html)
+2. NekRS: [https://github.com/Nek5000/nekRS?tab=readme-ov-file#build-instructions](https://github.com/Nek5000/nekRS?tab=readme-ov-file#build-instructions)
+3. OpenMC: [https://docs.openmc.org/en/stable/usersguide/install.html#prerequisites](https://docs.openmc.org/en/stable/usersguide/install.html#prerequisites)
+4. DAGMC: [https://svalinn.github.io/DAGMC/install/dependencies.html](https://svalinn.github.io/DAGMC/install/dependencies.html)
+
+Out of the above requirements, MOOSE's are the strictest. At present the MOOSE minimum requirements are:
+
+- A "modern" C/C++ compiler: GCC 8.5.0 - 12.2.1 or LLVM/Clang 10.0.1 - 16.0.6;
```
aprilnovak, line 24: Can we do an `!include` to automatically populate this? That will protect this from going out of date. You should we able to link to any of the markdown files in MOOSE from Cardinal's docs

doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+
+# This points to the actual location where your Cardinal directory is stored on the HPC's file system.
+# Most HPC systems use symbolic links for directories and files to hide their true paths (to make life
+# easier for users), which is why this is necessary.
+DIRECTORY_WHERE_YOU_HAVE_CARDINAL=$HOME/projects
+HOME_DIRECTORY_SYM_LINK=$(realpath -P $DIRECTORY_WHERE_YOU_HAVE_CARDINAL)
+export NEKRS_HOME=$HOME_DIRECTORY_SYM_LINK/cardinal/install
+
+# These environment variables indicate how many processes you want to build with. Be careful with the number
+# of jobs specified, most HPC systems limit the number of processes you can run with on login nodes.
```
aprilnovak, line 52: Can we add a comment about memory (during compilation) increasing when you build in parallel? I think some people have run into this in the past, and were able to fix it by reducing these numbers.

doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+export JOBS=8
+export LIBMESH_JOBS=8
+export MOOSE_JOBS=8
+```
+
+The best place to put these exports is your `.bashrc` to avoid needing to export these environment variables every time you spin up a new shell. Once these variables have been exported and you've loaded a set of modules that you believe will compile Cardinal, you can run MOOSE diagnostics to make sure you haven't missed anything:
+
+```bash
+./contrib/moose/scripts/diagnostics.sh
```
aprilnovak, line 61: wow I can't believe I didn't know about this! Thanks @nuclearkevin !

Not necessary for this PR, but do you think we could have similar scripts for OpenMC/NekRS/DAGMC? Just a thought; not sure if it'd be a pain to maintain those scripts

doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+If critical errors are reported at this stage, it is likely because your current module set does not meet MOOSE's requirements. If you receive an error stating that `jinja2` is missing, you can safely move on.
+
+**Note:** If you want to build NekRS with GPU support, you need GPU-enabled compilers (CUDA compilers, OpenCL compilers, etc.) on the login node. Some HPC systems only allow users to load those modules on nodes which contain GPUs. If that is the case for you, you'll need to build with a job script (see the next session). To enable a GPU build, set one of the following environment variables in Cardinal's makefile to `1` (depending on the specific GPUs on your system): `OCCA_CUDA_ENABLED`, `OCCA_HIP_ENABLED`, or `OCCA_OPENCL_ENABLED`. You'll also need to make sure you load modules with GPU compilers.
```
aprilnovak, line 66:
```suggestion
**Note:** If you want to build NekRS with GPU support, you need GPU-enabled compilers (CUDA compilers, OpenCL compilers, etc.) on the login node. Some HPC systems only allow users to load those modules on nodes which contain GPUs. If that is the case for you, you'll need to build with a job script (see the next section). To enable a GPU build, set one of the following environment variables in Cardinal's makefile to `1` (depending on the specific GPUs on your system): `OCCA_CUDA_ENABLED`, `OCCA_HIP_ENABLED`, or `OCCA_OPENCL_ENABLED`. You'll also need to make sure you load modules with GPU compilers.
```

doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+```
+nohup ./contrib/moose/scripts/update_and_rebuild_wasp.sh &
+```
+
+Building libMesh may fail due to a lack of libraries that are normally included by default on Linux machines. The most common culprit is `libxdr`, which is required for XDR binary file output in libMesh/MOOSE. The vast majority of the time you'll use Exodus binary output when running Cardinal, and so XDR can be disabled by adding two flags to `update_and_rebuild_libmesh.sh`.
+
+```
+./contrib/moose/scripts/update_and_rebuild_libmesh.sh --disable-xdr-required --disable-xdr
+```
+
```
aprilnovak, line 87:
```suggestion
There are many similar flags like this, that can disable parts of libMesh or ask libMesh to install dependencies if it couldn't find them itself. If you encounter persistent errors, ask on the [MOOSE discussion page](https://github.com/idaholab/moose/discussions) for help, and they may be able to recommend a flag for you to set.
```

doc/content/hpc_build_tips.md
```diff
@@ -0,0 +1,156 @@
...
+If you didn't get any build errors when building PETSc/libMesh/WASP, you can run
+
+```
+nohup make -j8 MAKEFLAGS=-j8 &
+```
+
+to build MOOSE, Cardinal, and all of Cardinal's dependencies. Occasionally the MOOSE Solid Mechanics module will fail to build due to missing `F77` compilers. If that is the case, you can either find mpif77 compilers in a different module set or disable the Solid Mechanics module by setting `SOLID_MECHANICS := no` in Cardinal's makefile.
+
```
aprilnovak, line 95:
```suggestion
The only MOOSE module you absolutely must have enabled is the `REACTOR` module (some of Cardinal's source files link against utilities in that module). Otherwise, feel free to disable all others.
```

aprilnovak:
Maybe it doesn't belong in this guide, but it'd be nice if we had some rough recommendations on ranks/threads per node and/or just a basic discussion on the consequences on going to higher ranks (less memory per rank).

doc/content/hpc_build_tips.md
nuclearkevin: We can do a `!include sqa/minimum_requirements.md`.

doc/content/hpc_build_tips.md
nuclearkevin: We can definitely implement similar scripts for the other Cardinal dependencies. I think it would be rather annoying to maintain them as we'd need to check and see if dependencies have changed do a submodule update, and if they have we'd need to update the script(s). In my own (limited) experience making sure the system meets the MOOSE requirements means we're already most of the way there as libMesh + PETSc pull in all of the dependencies that OpenMC/DAGMC need. 

NekRS is the only picky dependency and we update it's submodule far less often then the others, so it's probably worth a diagnostics script. Might be something for Manavi to look into?

aprilnovak: Great idea! I'll make an issue to keep track of it

PR #1087, opened 2025-03-14
Remove unused params from mesh only tests

shikhar413:
I've added a feature in MOOSE that causes --mesh-only mode to error out if an unused / misspelled parameter exists in the input - https://github.com/idaholab/moose/pull/29824

This causes some Cardinal tests to fail (https://civet.inl.gov/job/2748234/), mainly because background_block_name should be used instead of background_block_names for PatternedHexMeshGenerator.

@aprilnovak this will likely cause some re-goldings, let me know if this is the preferred option here or to remove the background_block_names parameter entirely so that re-golding is not needed.

### DIFF
--- a/test/tests/neutronics/symmetry/rotational/solid_mesh.i
+++ b/test/tests/neutronics/symmetry/rotational/solid_mesh.i
@@ -71,7 +71,7 @@ height = 6.343                           # height of the full core (m)
     rotate_angle = 0
 
     background_block_id = '1'
-    background_block_names = 'graphite'
+    background_block_name = 'graphite'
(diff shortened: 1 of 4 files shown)


aprilnovak:
Thanks @shikhar413! This is really useful, I know I've been burned before on input error checking with `--mesh-only` before. Can you go accept the invitation to join `neams-th-coe` organization, and then re-trigger the tests? This is needed for the test suite to launch.

aprilnovak:
Thanks @shikhar413! Looks like this didn't cause any failures.

PR #1088, opened 2025-03-15
Update moose submodule refs #0

loganharbour:


src/auxkernels/FDTallyGradAux.C
```diff
@@ -84,7 +84,7 @@ FDTallyGradAux::FDTallyGradAux(const InputParameters & parameters)
                    "applied to " +
                    std::string(getParam<MooseEnum>("score")) + "!");
 
-  if (score_vars[_bin_index]->feType() != FEType(CONSTANT, MONOMIAL))
+  if (score_vars[_bin_index]->feType() != FEType(libMesh::CONSTANT, libMesh::MONOMIAL_VEC))
```
nuclearkevin, line 87:
```suggestion
  if (score_vars[_bin_index]->feType() != FEType(libMesh::CONSTANT, libMesh::MONOMIAL))
```

nuclearkevin (changes requested):

src/auxkernels/FDTallyGradAux.C
nuclearkevin: Should be `MONOMIAL` here, not `MONOMIAL_VEC`.

aprilnovak: Did the enumeration get changed in libMesh/MOOSE with this new MOOSE submodule? Just wondering why we're seeing this now

nuclearkevin: The thing that changed between the different MOOSE submodules is the need for the explicit libMesh namespace on the enums. @loganharbour made a small error when adding in this namespace where he swapped out a `MONOMIAL` with a `MONOMIAL_VEC`.

aprilnovak: Ah I see, I didn't catch that in the original diff.

loganharbour: Oops. working-at-a-coffee-shop fail

loganharbour:
VTB failure will be resolved on the VTB. This will get stuck on next.

PR #1090, opened 2025-03-16
Add scores necessary for MGXS generation

### LINKED ISSUE #1089: Add `nu_scatter`, `nu_fission`, and `inverse_velocity` to tallies
## Reason
There are several reaction rate scores that are necessary for computing multi-group cross sections in Cardinal. We should add these to facilitate multi-physics informed MGXS workflows where Cardinal generates cross sections for a deterministic code.

## Design
These scores should be added to `getTallyScoreEnum(...)`, `getSingleTallyScoreEnum(...)`, and `OpenMCProblemBase::isReactionRateScore(...)`. 

## Impact
This supports #1020 while adding new score options in Cardinal's mapped tallies.

nuclearkevin:
Closes #1089


aprilnovak:
The OpenMC documentation seems to list the units for `inverse-velocity` as cm/s. Does anything special need to be done with normalization here?

nuclearkevin:
The OpenMC docs list the units of velocity ($v$) in OpenMC as cm/s. The `inverse-velocity` score itself is flux-weighted, and so it has units of:

```math
\frac{\phi}{v} \equiv  \Bigg[\frac{\text{particle-cm}}{\text{src particle}}\Bigg]\Bigg[\frac{\text{s}}{\text{cm}}\Bigg]\equiv\Bigg[ \frac{\text{particle-s}}{\text{src particle}}\Bigg]
```

(OpenMC). When normalized in the same manner as a volumetric score in Cardinal, this yields:

```math
\frac{\phi}{v} \equiv \frac{\text{particle}}{\text{cm}^{3}}
```

(Cardinal). So, we shouldn't need to do anything special with the inverse velocity thanks to the flux-weighting.

aprilnovak:
And I guess we're sure OpenMC isn't internally also dividing by another tally of the flux, to get just 1/v? As in, the output of this tally is phi/v?

nuclearkevin:
I'm pretty sure that isn't happening since OpenMC's MGXS Python code performs a division by a flux tally to compute the group-wise inverse velocity, and I don't see anywhere in the tally code where a division by flux is happening.

src/tallies/TallyBase.C
```diff
@@ -114,15 +116,30 @@ TallyBase::TallyBase(const InputParameters & parameters)
                  "Tracklength estimators are currently incompatible with photon transport and "
                  "heating scores! For more information: https://tinyurl.com/3wre3kwt");
 
+    if (estimator != tally::analog && nu_scatter)
+      paramError("estimator", "Non-analog estimators are not supported for nu_scatter scores!");
```
aprilnovak, line 120: is there a similar restriction for nu-fission?

nuclearkevin: Surprisingly, no. Tracklength and collision estimators are both valid for `nu-fission`.

PR #1096, opened 2025-04-03
Add remaining filters necessary for MGXS generation

### LINKED ISSUE #1094: Add filters necessary for multi-group cross section generation.
## Reason
Multi-group cross section generation requires several added filters, notably: Legendre functional expansion filters in the particle scattering angle, energy filters that bin particles by their outgoing energy after a reaction, and particle filters to discriminate against neutrons/photons. We should add `[Filters]` which make these OpenMC filters available to Cardinal's mapped tallies, enabling the generation of multi-group scattering matrices and discrete chi spectra.

## Design
OpenMC's `LegendreFilter`, `EnergyoutFilter`, and `ParticleFilter` would need to be wrapped by `FilterBase`-derived objects in Cardinal. The changes made in #1092 allow for non-spatial functional expansion filters, and so additional changes outside of these filters shouldn't be necessary.

## Impact
This enhancement supports MGXS generation in Cardinal (#1020) on the mesh mirror for both photon and neutron multi-group cross sections.

nuclearkevin:
Closes #1094


src/filters/EnergyOutFilter.C
```diff
@@ -0,0 +1,96 @@
...
+    _energy_bnds = EnergyFilter::getGroupBoundaries(
+        getParam<MooseEnum>("group_structure").getEnum<energyfilter::GroupStructureEnum>(), this);
+
+  // Two boundaries are required at minimum to form energy bins.
+  if (_energy_bnds.size() < 2)
+    paramError("energy_boundaries",
+               "At least two energy values are required to create energy bins!");
+
+  // Sort the boundaries so they're monotonically decreasing.
+  std::sort(_energy_bnds.begin(),
```
aprilnovak, line 76: Do we also want to have a `std::unique` in case the user accidentally enters the same energy 2 times? e.g. `energy_boundaries = '0.06 0.06'`. Or, does OpenMC catch that error for us?

src/tallies/TallyBase.C
```diff
@@ -202,6 +205,21 @@ TallyBase::TallyBase(const InputParameters & parameters)
     }
   }
 
+  // Check the estimator to make sure it doesn't conflict with certain filters.
+  for (auto & f : _ext_filters)
```
aprilnovak, line 209: I think I noticed from the CIVET test history that this was added to get parallel reproducibility with distributed mesh. Is that correct? Are these constraints that also exist on the OpenMC side and we just forgot to include them here, or have we identified some strange behavior with parallel runs?

src/tallies/TallyBase.C
```diff
@@ -202,6 +205,21 @@ TallyBase::TallyBase(const InputParameters & parameters)
...
+  {
+    if (dynamic_cast<AngularLegendreFilter *>(f.get()) &&
```
aprilnovak, line 211: These two paramErrors can be collapsed into one.

src/tallies/TallyBase.C
nuclearkevin: I think it's a mixture of both. OpenMC requires EnergyOut and Legendre filters to use analog estimators, and so it resets the estimator to analog when creating a filter _from an XML node_. When filters are added on the C++ side of things OpenMC doesn't do this, which resulted in strange parallel behaviour when the tests accidentally used a tracklength estimator with EnergyOut / Legendre filters.

src/filters/EnergyOutFilter.C
nuclearkevin: OpenMC catches neither of these cases (they appear to defer to the Python API to prevent the user from adding duplicate / negative bins). I'll add checks to both the `EnergyOutFilter` and the `EnergyFilter` to make sure we catch those.

src/tallies/TallyBase.C
aprilnovak: got it, thanks!

src/filters/EnergyOutFilter.C
```diff
@@ -72,11 +72,22 @@ EnergyOutFilter::EnergyOutFilter(const InputParameters & parameters)
     paramError("energy_boundaries",
                "At least two energy values are required to create energy bins!");
 
+  // Check to make sure none of the boundaries are negative.
```
aprilnovak, line 75: Maybe these checks could go into a base class which owns those two parameters?

src/filters/EnergyFilter.C
```diff
@@ -69,11 +69,20 @@ EnergyFilter::EnergyFilter(const InputParameters & parameters)
...
+  for (const auto & bnd : _energy_bnds)
+    if (bnd < 0.0)
+      paramError("energy_boundaries", "Energy group boundaries must be positive to create energy bins!");
+
   // Sort the boundaries so they're monotonically decreasing.
   std::sort(_energy_bnds.begin(),
             _energy_bnds.end(),
             [](const Real & a, const Real & b) { return a < b; });
 
+  // Check for duplicate entries.
```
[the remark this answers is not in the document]
nuclearkevin: This case is handled by the use of `std::sort` followed by `std::adjacent_find`. The sort will place the same items next to eachother (`0.06 1.0 0.06` becomes `0.06 0.06 1.0`). `std::adjacent_find` then detects the `0.06 0.06` and triggers the if statement to throw a `paramError`.

src/filters/EnergyOutFilter.C
nuclearkevin: I'm normally not a fan of creating base classes for only two derived classes as  it makes the the code more difficult to understand for new developers. I think decreasing the number of regression tests justifies it in this case though.

test/tests/neutronics/filters/energy_out/tests
```diff
@@ -18,41 +18,6 @@
     requirement = "The system shall be capable of adding an EnergyOutFilter (with a group structure)."
     required_objects = 'OpenMCCellAverageProblem'
   []
-  [not_enough_bnds]
-    type = RunException
-    input = cell.i
-    cli_args = 'Problem/Filters/EnergyOut/energy_boundaries="1.0"'
-    expect_err = "At least two energy values are required to create energy bins!"
-    requirement = "The system shall error if there aren't enough energy boundaries to form an EnergyOutFilter."
```
aprilnovak: Am I missing these somewhere? I think we do want to retain these tests, but now with the base class we just don't need to have them duplicated for `EnergyFilter` and `EnergyOutFilter`.

nuclearkevin: The tests that were removed for `EnergyOutFilter` are duplicated in the `EnergyFilter` tests. See [test/tests/neutronics/filters/energy](https://github.com/nuclearkevin/cardinal/blob/mgxs_filters/test/tests/neutronics/filters/energy/tests). They're untouched, and so don't show up in the diff.

PR #1099, opened 2025-04-07
Swap the units of `damage_energy` scores to eV

### LINKED ISSUE #1098: Store damage energy with energy units of eV instead of J
## Reason
At present the `damage_energy` score is stored with the units of `W / cm3`. This doesn't make a whole lot of sense as damage energy isn't a true heating score, and is only tallied as an intermediate step to facilitate the calculation of DPA. We should store damage energy with the units `eV / (cm3 - s)` instead.

## Design
Minor changes to tally normalization would be required and some tests which use damage energy would need to be regolded to account for the different units.

## Impact
Improved user experience. Reduces the number of unit conversions required to compute DPA as a post-processing step.

nuclearkevin:
Closes #1098


aprilnovak:
Thanks @nuclearkevin !

PR #1100, opened 2025-04-08
Update dagmc

aprilnovak:


nuclearkevin:
I believe you need to add `-DENABLE_MPI=ON` to the CMake arguments for MOAB to get this to build. They changed up parts of the build system, and so auto-detection of MPI in some places is not supported (for some reason?).

PR #1105, opened 2025-04-22
Add kinetics parameters

### LINKED ISSUE #1104: Add kinetic parameter (Beta / Lambda effective) tallies
## Reason
OpenMC recently merged the capability to calculate kinetics parameters ($\Lambda_{eff}$ / $\beta_{eff}$) in https://github.com/openmc-dev/openmc/pull/3133. We should expose this capability for k-eigenvalue calculations to enable transient simulations where point kinetics parameters are updated on-the-fly to account for multi-physics feedback.

## Design
Global tallies with `ifp-time-numerator`, `ifp-beta-numerator`, and `ifp-denominator` scores should be added by `OpenMCCellAverageProblem` or `OpenMCProblemBase` when requested by the user. This will require some additional work to set up the IFP calculation given that the global tallies wouldn't be added from an OpenMC xml file. Two post-processors should also be added to take the tallied scores and convert them into $\Lambda_{eff}$ / $\beta_{eff}$ for point kinetics calculations after the k-eigenvalue calculation is complete.

## Impact
This would remove the dependence Cardinal currently has on kinetics parameters calculated with other codes for transient neutronics simulations, and would enhance verification efforts which require these parameters (i.e. KRUSTY).

nuclearkevin:
Closes #1104


nuclearkevin:
The test problem (`both.i`) has been compared against a raw OpenMC calculation with some post-processing in Python. 

Cardinal results:
```
+----------------+----------------+----------------+
| time           | lambda     | beta           |
+----------------+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   1.767542e-05 |   8.623088e-03 |
+----------------+----------------+----------------+
```

OpenMC results:
```
[[[1.46028854e-05 7.75000000e-03 8.98750000e-01]]]
Lambda: 1.767546543727036e-05
Beta: 0.008623087621696798
```

doc/content/source/postprocessors/BetaEffective.md
```diff
@@ -0,0 +1,17 @@
+# BetaEffective
+
+## Description
+
+`BetaEffective` computes $\beta_{eff}$ given that the [OpenMC problem](OpenMCCellAverageProblem.md)
```
aprilnovak, line 5: For folks who click on this page perhaps without knowing what beta_eff and Lambda_eff are, it would be good to add a 1-2 sentence description of what these parameters mean/represent.

doc/content/source/problems/OpenMCCellAverageProblem.md
```diff
@@ -404,6 +404,29 @@ can be used to create tallies and/or filters as well. In either case, these
 objects do not interact with tallies and filters used to map data to the mesh
 mirror ([#mapped]).
 
+## Neutron Kinetics
```
aprilnovak, line 407:
```suggestion
## Neutron Kinetics Parameters
```

doc/content/source/problems/OpenMCCellAverageProblem.md
```diff
@@ -404,6 +404,29 @@ can be used to create tallies and/or filters as well. In either case, these
...
+
+Cardinal exposes the calculation of neutron kinetics parameters in OpenMC with iterated
+fission probabilities. $\Lambda_{eff}$ and $\beta_{eff}$ can be obtained by first setting
+`calc_kinetics_params = true`. The parameters can then be accessed with a
```
aprilnovak, line 411:
```suggestion
`calc_kinetics_params = true`. The parameters can then be accessed with
```

src/base/OpenMCProblemBase.C
```diff
@@ -74,6 +74,17 @@ OpenMCProblemBase::validParams()
       false,
       "Whether to skip writing any statepoint files from OpenMC; this is a performance "
       "optimization for scenarios where you may not want the statepoint files anyways");
+
+  // Kinetics parameters.
+  params.addParam<bool>("calc_kinetics_params",
+                        false,
+                        "Whether or not Cardinal should enable the calculation of kinetics "
+                        "parameters (Lambda effective / beta effective).");
```
aprilnovak, line 82:
```suggestion
                        "parameters (Lambda effective and beta effective).");
```

src/base/OpenMCProblemBase.C
```diff
@@ -326,6 +354,21 @@ OpenMCProblemBase::externalSolve()
         std::make_unique<openmc::FileSource>(sourceBankFileName()));
   }
 
+  // Reinitialize the IFP parameters tally.
+  if (_calc_kinetics_params)
+  {
+    if (_ifp_tally_index > -1 || _ifp_tally)
```
aprilnovak, line 360: Can we use a constexpr instead of `-1` so that it's clear this value means something special? Helps with readability a bit

src/base/OpenMCProblemBase.C
```diff
@@ -725,6 +768,15 @@ OpenMCProblemBase::numCells() const
   return n_openmc_cells;
 }
 
+const openmc::Tally &
+OpenMCProblemBase::getKineticsParamTally()
+{
+  if (!_calc_kinetics_params || !_ifp_tally)
```
aprilnovak, line 774: Can this if-statement be equivalently written as `if(!_ifp_tally)`? I guess I'm not clear on when that tally would exist by calc_kinetics_params would be false.

src/postprocessors/BetaEffective.C
```diff
@@ -0,0 +1,64 @@
...
+  params.addClassDescription(
+      "A post-processor which computes and returns the kinetics parameter $\\beta_{eff}$.");
+
+  return params;
+}
+
+BetaEffective::BetaEffective(const InputParameters & parameters)
+  : GeneralPostprocessor(parameters), OpenMCBase(this, parameters)
+{
+  if (openmc::settings::run_mode != openmc::RunMode::EIGENVALUE)
```
aprilnovak, line 41: Can this error check be removed? If on line 44 we are requiring OpenMC to have kinetics params computed, then I think we can rely on the eigenvalue mode check in `OpenMCCellAverageProblem` to reduce a bit of code duplication? (and test coverage)

include/base/OpenMCProblemBase.h
```diff
@@ -492,6 +504,15 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
    */
   bool _run_on_adaptivity_cycle;
 
+  /// A flag to indicate if OpenMC is calculating kinetics parameters.
+  const bool _calc_kinetics_params;
```
aprilnovak, line 508: At this point, the class hierarchy in the OpenMC classes doesn't really get exercised, since we never made another lowest-level daughter class. But assuming one day that we did, would these IFP parameters better belong in the base class? They don't have any notion of spatial mapping.

I don't think this is mandatory in this PR, but just wanted to mention it in case you felt it belonged in the base class that I'd also support that.

src/postprocessors/BetaEffective.C
```diff
@@ -0,0 +1,64 @@
...
+    mooseError("Beta effective is only computed when running OpenMC in eigenvalue mode!");
+
+  if (!_openmc_problem->computeKineticsParams())
+    mooseError(
+        "BetaEffective can only be used if the OpenMC problem is computing kinetics parameters!");
+}
+
+Real
+BetaEffective::getValue() const
```
aprilnovak, line 50: Does IFP give us the beta effective for each delayed group? Or only total?

test/tests/neutronics/kinetics/both.i
```diff
@@ -0,0 +1,71 @@
...
+    type = CombinerGenerator
+    inputs = 'solid_ids fluid_ids'
+  []
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  verbose = true
+  power = 1e4
+  temperature_blocks = '100 200'
```
aprilnovak, line 37: same comment in this file -- for code newcomers, simplest is best

src/base/OpenMCProblemBase.C
nuclearkevin: `_ifp_tally` will also work here since the pointer is initialized with a `nullptr`. I'll remove the index checks (which also removes the need for a `constexpr` for a special value).

include/base/OpenMCProblemBase.h
nuclearkevin: These IFP parameters are in `OpenMCProblemBase` already. I agree, they shouldn't be owned by `OpenMCCellAverageProblem`.

aprilnovak: Whoops, sorry I misread the class name!

include/base/OpenMCProblemBase.h
```diff
@@ -492,6 +504,15 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
...
+
+  /// The index of a global tally to accumulate the scores required for kinetics parameters.
+  int _ifp_tally_index = -1;
```
aprilnovak, line 511: Does this need to be initialized?

src/base/OpenMCProblemBase.C
```diff
@@ -198,6 +210,22 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
   // The OpenMC wrapping doesn't require material properties itself, but we might
   // define them on some blocks of the domain for other auxiliary kernel purposes
   setMaterialCoverageCheck(false);
+
+  // If the user requests kinetics parameters, make sure it's enabled in OpenMC.
+  if (_calc_kinetics_params)
+  {
+    if (_run_mode != openmc::RunMode::EIGENVALUE)
+      paramError("calc_kinetics_params",
```
aprilnovak, line 218: I didn't see a test for this error - should we add one, or will OpenMC already error for us?

src/postprocessors/BetaEffective.C
aprilnovak: Can you add an issue to keep track of that development? It would be nice for us to support this if it gets added in the future.

src/postprocessors/LambdaEffective.C
```diff
@@ -0,0 +1,59 @@
...
+    mooseError(
+        "LambdaEffective can only be used if the OpenMC problem is computing kinetics parameters!");
+}
+
+Real
+LambdaEffective::getValue() const
+{
+  const auto & ifp_tally = _openmc_problem->getKineticsParamTally();
+
+  const Real num =
```
aprilnovak, line 49: Can we expose the standard deviation and relative error in these two kinetics parameters through this postprocessor? Perhaps with a `value_type` enumeration?

src/postprocessors/BetaEffective.C
nuclearkevin: #1107

src/base/OpenMCProblemBase.C
nuclearkevin: Whoops, good catch! OpenMC should error for this, but I think it's more intuitive for the user if we manage the error and let the user know it was caused by them setting `calc_kinetics_params = true`. OpenMC could error if IFP is enable in `settings.xml`, so the distinction helps with debugging.

include/base/CardinalEnums.h
```diff
@@ -250,6 +251,16 @@ enum RelaxationEnum
 };
 } // namespace relaxation
 
+namespace kinetics
+{
+  enum KineticsOutputEnum
+  {
+    Value,
```
aprilnovak: Maybe we can also add this to the k-eigenvalue postprocessor, for uniformity... could be done in a separate PR after this one, then we could delete `KStandardDeviation`

src/postprocessors/LambdaEffective.C
nuclearkevin: Added the capability, and modified the test to include relative error and standard deviation output.

include/base/CardinalEnums.h
nuclearkevin: Agreed. We'd need to change a lot of tests and tutorials to make that work, so doing it in a separate PR makes the most sense IMO. I'll create an issue to track that for the future.

src/base/OpenMCBase.C
```diff
@@ -38,4 +38,18 @@ OpenMCBase::OpenMCBase(const MooseObject * moose_object, const InputParameters &
...
+Real
+OpenMCBase::stdev(const double & mean, const double & sum_sq, unsigned int realizations) const
+{
+  return realizations > 1
+             ? std::sqrt(std::max(0.0, (sum_sq / realizations - mean * mean) / (realizations - 1)))
+             : 0.0;
+}
+
+Real
+OpenMCBase::relerr(const double & mean, const double & sum_sq, unsigned int realizations) const
```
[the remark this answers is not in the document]
nuclearkevin: We could use `OpenMCProblemBase::relativeError(...)` instead. The more I think about it, the more I realize it would make more sense for `OpenMCProblemBase::relativeError(...)` to live inside `OpenMCBase` as a `static` function instead (there's no reason for it to be a member function). The same goes for `OpenMCProblemBase::tallySum(...)`, `OpenMCProblemBase::tallySumAcrossBins(...)`, and `OpenMCProblemBase::tallyMeanAcrossBins(...)`.

I'll save those for a different PR as it's very out of scope for the IFP changes implemented here.

PR #1109, opened 2025-04-27
Merge k Postprocessors

### LINKED ISSUE #1108: Merge `KEigenvalue` and `KStandardDeviation` into a single post-processor
## Reason
#1105 added two new post-processors which give the user the option of fetching kinetics parameters, their standard deviations, or their relative errors. To maintain consistency between different post-processors (and to reduce the number of added MOOSE objects) `KEigenvalue` and `KStandardDeviation` should be merged together into a single post-processor, with an enum input parameter to determine which value is calculated.

## Design
`KStandardDeviation` should be removed. All of it's capabilities will be merged into `KEigenvalue` with the enum parameter `output` that decides if k, k standard deviation, or k relative error is returned by the post-processor.

## Impact
This change will simplify input files (one fewer post-processor to keep track of). It will also break all existing input files which previously used a `KStandardDeviation` post-processor.

nuclearkevin:
Closes #1108


src/postprocessors/BetaEffective.C
```diff
@@ -74,17 +74,18 @@ BetaEffective::getValue() const
 
   switch (_output)
   {
-    case kinetics::KineticsOutputEnum::Value:
+    case eigenvalue::EigenKineticsOutputEnum::Value:
       return beta_eff;
 
-    case kinetics::KineticsOutputEnum::StDev:
+    case eigenvalue::EigenKineticsOutputEnum::StDev:
```
aprilnovak, line 80: The use of `eigenvalue` as the enumeration is kind of confusing to someone reading the code - can be name it something like `statistics::OutputEnum`?

src/postprocessors/KEigenvalue.C
```diff
@@ -28,24 +30,50 @@ KEigenvalue::validParams()
...
   params += OpenMCBase::validParams();
+  params.addClassDescription("k eigenvalue computed by OpenMC");
   params.addParam<MooseEnum>("value_type",
                              getEigenvalueEnum(),
                              "Type of eigenvalue global tally to report");
-  params.addClassDescription("k eigenvalue computed by OpenMC");
+  params.addParam<MooseEnum>(
+      "output",
+      getEigenKineticsOutputEnum(),
+      "The value to output. Options are $k_{eff}$ (val), the standard deviation "
```
aprilnovak, line 40: Instead of `val`, how about `mean`?

PR #1115, opened 2025-05-07
Add Windriver to the HPC list

### LINKED ISSUE #1112: Add build instructions for Windriver
## Reason
Windriver is a new HPC system at INL. Cardinal does not have any build instructions for it, we should add some to the website.

## Design
Addition of build instructions to the website. The module list and submission script should be similar to Bitterroot as the machines are almost identical. 

## Impact
New build instructions yielding an improved user/developer experience.

nuclearkevin:
Works on both Windriver and Bitterroot.

Closes #1112


tjlaboss:
I volunteer to review this one.

nuclearkevin:
@aprilnovak @tjlaboss this should be ready to go now. I found a combination of modules that works for NekRS + OpenMC + DAGMC on both Bitterroot and Windriver. This was confirmed by running the pincell multiphysics problem.

nuclearkevin:
Fixed the module list (again). This version uses an OpenMPI version which doesn't cause NekRS to hang on loading meshes with > 1000 MPI ranks. Verified by running the 67 pebble CHT case across 10 nodes (1100 ranks). Also ran the pincell multiphysics case to ensure that both OpenMC and NekRS run as expected.

Surprisingly, Cardinal compiles without `use.moose` + `moose-tools` (indicating the mandatory python packages are no longer required).

tjlaboss:
Great. I will give this a review with the new module list.

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
-module load cmake/3.30.1-gcc-13.3.0-6mtw
-# Revise for your repository location
-export NEKRS_HOME=$HOME/cardinal/install
-export OPENMC_CROSS_SECTIONS=$HOME/cross_sections/endfb-vii.1-hdf5/cross_sections.xml
+# Bitterroot modules
+if [ "$SHORTHOST" = "bitterroot1" ] || [ "$SHORTHOST" = "bitterroot2" ]; then
+  echo "Loading Bitterroot modules..."
+  module purge
+  module load openmpi/4.1.5_ucx1.14.1 cmake/3.29.3
+fi
```
tjlaboss: I suggest merging this `if` with the WindRiver instructions

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
-module purge
-module load use.moose
-module load moose-tools
-module load openmpi/4.1.7-gcc-13.3.0-xpfl
-module load cmake/3.30.1-gcc-13.3.0-6mtw
-# Revise for your repository location
-export NEKRS_HOME=$HOME/cardinal/install
-export OPENMC_CROSS_SECTIONS=$HOME/cross_sections/endfb-vii.1-hdf5/cross_sections.xml
+# Bitterroot modules
+if [ "$SHORTHOST" = "bitterroot1" ] || [ "$SHORTHOST" = "bitterroot2" ]; then
```
tjlaboss: Bitterroot compute nodes are `br` followed by the node number. May I suggest:

```bash
if [[ "$SHORTHOST" ~= br[0-9]+ ]] || [[ "$SHORTHOST" =~ bitterroot* ]]
```

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
+if [ -f  ~/.bashrc_local ]; then
+       . ~/.bashrc_local
+fi
+
+# Windriver modules
+if [ "$SHORTHOST" = "windriver1" ] || [ "$SHORTHOST" = "windriver2" ]; then
+  echo "Loading Windriver modules..."
+  module purge
+  module load openmpi/4.1.5_ucx1.14.1 cmake/3.29.3
+fi
```
tjlaboss: See above

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
+
+# Source local definitions
+
+if [ -f  ~/.bashrc_local ]; then
+       . ~/.bashrc_local
+fi
+
+# Windriver modules
+if [ "$SHORTHOST" = "windriver1" ] || [ "$SHORTHOST" = "windriver2" ]; then
```
tjlaboss: WindRiver compute nodes are `wr` followed by the row number, then `n` followed by the node number in that row. May I suggest:

```bash
if [[ "$SHORTHOST" ~=wr[0-9]+n[0-9]+ ]] || [[ "$SHORTHOST" =~ windriver* ]]
```

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
+
+!listing scripts/job_bitterriver language=bash caption=Sample job script with the `moose` project code id=bt2
+
+This submission script optimizes performance for OpenMC on Bitterroot using shared memory parallelism and NUMBA bindings on a compute node:
+
+!listing scripts/job_bitterriver_openmc language=bash caption=Sample OpenMC job script with the `moose` project code id=bt3
+
+## Windriver
+
+[Windriver](https://inl.gov/hpc/about/) is an [!ac](HPC) system at [!ac](INL). It has 843 compute nodes, each with
```
tjlaboss: Update link to https://hpcweb.hpc.inl.gov/hardware/windriver/

doc/content/hpc.md
```diff
@@ -338,18 +338,73 @@ if [ -f  ~/.bashrc_local ]; then
...
+
+This submission script is not optimized for any particular problem run with Cardinal:
+
+!listing scripts/job_bitterriver language=bash caption=Sample job script with the `moose` project code id=bt2
+
+This submission script optimizes performance for OpenMC on Bitterroot using shared memory parallelism and NUMBA bindings on a compute node:
+
+!listing scripts/job_bitterriver_openmc language=bash caption=Sample OpenMC job script with the `moose` project code id=bt3
+
+## Windriver
```
tjlaboss: The R in WindRiver is capitalized (replace throughout)

doc/content/hpc.md
tjlaboss: Also do that for Bitterroot

tjlaboss:
> Surprisingly, Cardinal compiles without use.moose + moose-tools (indicating the mandatory python packages are no longer required).

I'm not sure about this. The MOOSE scripts do have some dependencies, such as pyyaml and jinja2, which are not Python builtins.

nuclearkevin:
> I'm not sure about this. The MOOSE scripts do have some dependencies, such as pyyaml and jinja2, which are not Python builtins.

Even without having `packaging` and `pyyaml` loaded (from my experience these are the only python modules that are actually required when building) `CardinalRevisions.h` and `MooseRevisions.h` were generated correctly (and so the build didn't fail). Running tests and other components of the python infrastructure would definitely not work, but Cardinal users don't normally do this on HPC systems so not loading those modules shouldn't cause any problems.

doc/content/hpc.md
nuclearkevin: I would prefer to avoid merging these instructions just to maintain the structure of the HPC instructions (one subheading per machine).

doc/content/hpc.md
nuclearkevin: See my comment above.

doc/content/hpc.md
nuclearkevin: That link is internal to INL - it cannot be accessed unless you're connected to the lab via a VPN. We could do https://hpcweb.hpcondemand.inl.gov/hardware/windriver/ and https://hpcweb.hpcondemand.inl.gov/hardware/bitterroot/ instead.

Not a huge fan of linking to access controlled sites in an open-source code, but it can't be helped in this case...

doc/content/hpc.md
tjlaboss: Sounds good

doc/content/hpc.md
tjlaboss: I always forget which links are and aren't accessible outside the lab's network!

Let's go for the second (the OnDemand links). If users are looking at WindRiver/Bitterroot build instructions, they probably have the necessary access.

aprilnovak:
@tjlaboss is this ready for merge?

nuclearkevin:
Can we wait a bit before merging this? These instructions work for me on WindRiver/Bitterroot, but a user (@luizaldeia) has been trying to use these modules and isn't having much success. I'd like to get to the bottom of that first if possible.

nuclearkevin:
We fixed the issue @luizaldeia was having on Bitterroot / WindRiver (forgot the `export CC=mpicc` etc.). If @tjlaboss is good with these, the build instructions should be ready to merge @aprilnovak.

I've also taken the liberty of adding a "Last updated on" section for each machine so users have an idea of how out of date the build instructions are.

tjlaboss (approved):

PR #1117, opened 2025-05-09
Swap `blocks` for `block` to match MOOSE convention

### LINKED ISSUE #1116: Rename `blocks` to `block` to be consistent with the rest of MOOSE
## Reason
There are several classes in Cardinal (the tally system classes come to mind) where `blocks` are used to denote a block restriction instead of `block`. This is inconsistent with MOOSE convention, and should be changed.

## Design
`blocks` shoukd be changed to `block` in the input parameters of relevant objects. We'd then need to modify all input files included in Cardinal to take the change into account.

## Impact
Improved consistency with the rest of MOOSE.

nuclearkevin:
Closes #1116

### DIFF
--- a/doc/content/source/filters/EnergyOutFilter.md
+++ b/doc/content/source/filters/EnergyOutFilter.md
@@ -32,12 +32,12 @@ scattered particles into two groups using the CASMO-2 group structure.
     [Heating]
       type = CellTally
       score = 'kappa_fission'
-      blocks = '100 200'
+      block = '100 200'
     []
     [Scattering]
       type = CellTally
       score = 'scatter'
-      blocks = '100 200'
+      block = '100 200'
       filters = 'EnergyOut'
     []
   []
--- a/doc/content/source/filters/EnergyOutFilter.md
+++ b/doc/content/source/filters/EnergyOutFilter.md
@@ -17,39 +17,8 @@ by setting `reverse_bins = true` if you wish to use the conventional multi-group
 As an example, a [CellTally](CellTally.md) named `Scattering` applies an `EnergyOutFilter` named `EnergyOut` to bin the outgoing energies from
 scattered particles into two groups using the CASMO-2 group structure.
 
-```
-[Problem]
-  type = OpenMCCellAverageProblem
-  verbose = true
-  power = 1e4
-  temperature_blocks = '100'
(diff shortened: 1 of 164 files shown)


aprilnovak:
Since this is a change likely to affect a lot of users, do you mind adding in a `mooseError` to check if the old `blocks` parameter was used, and to tell the user they need to rename it to `block`?

nuclearkevin:
@aprilnovak I was perusing through the documentation for `InputParameters`, and apparently there's an option to flag that a parameter has been renamed. I can try that out to see if it works for `blocks` to `block` instead of throwing an error (if that's preferred).

nuclearkevin:
That works for me, I'll add an error message for blocks.

nuclearkevin:
I'll open a PR in the VTB to fix the test failures after this is merged.

doc/content/source/filters/EnergyOutFilter.md
```diff
@@ -32,12 +32,12 @@ scattered particles into two groups using the CASMO-2 group structure.
     [Heating]
       type = CellTally
       score = 'kappa_fission'
-      blocks = '100 200'
```
aprilnovak, line 35: Can we use a file link instead of duplicating input here? This will insulate us from other syntax changes.

src/tallies/MeshTally.C
```diff
@@ -37,9 +37,11 @@ MeshTally::validParams()
                          "Coordinate to which this mesh should be "
                          "translated. Units must match those used to define the [Mesh].");
   params.addParam<std::vector<SubdomainName>>(
```
aprilnovak, line 39: is this parameter actually used in MeshTally?

aprilnovak: whoops nvm, forgot we've added features that use it

aprilnovak:
I think there are a few minor places where `blocks` still shows up in our docs:

```
(moose) {21:03}~/projects/cardinal/doc:blocks_to_block ✗ ➭ grep -rI "\`blocks\`" content
content/tutorials/pincell1.md:[CellTally](CellTally.md) in `[Tallies]`. The `blocks` are
content/tutorials/pincell1.md:`blocks` to blocks 2 and 3.
```

PR #1128, opened 2025-05-28
sync clustering brach with devel

magnoxemo:
In order to update clustering with new amalgamation changes, it needs to be up to date with devel. 

We should discuss also how we are going to keep this branch up to date with devel. @gonuke Do you have any thoughts on this?


lewisgross1296:
I think the three options are

1. Merge Walid's work from `magnoxemo/cardinal:feature` into `neams-th-coe:cardinal:devel` as we go and don't have the features turned on for users until ready. 
2. Keep `neams-th-coe/cardinal:clustering` up to date with `neams-th-coe/cardinal:devel` every time we want to merge a new PR from `magnoxemo/cardinal:feature` into `neams-th-coe/cardinal:clustering`.
3. Merge Walid's work from `magnoxemo/cardinal:feature` into `magnoxemo/cardinal:clustering` as we go and then one day merge `magnoxemo/cardinal:clustering` into  `neams-th-coe/cardinal:clustering` (or  `neams-th-coe/cardinal:devel`)

At this point, I think we could defer to the Cardinal dev team's preferences. I think that they all have pros and cons. (1) probably is a bit awkward to have unfinished features be merged, but if no one can use them until they're mature, that might be okay. (2) is slightly annoying to require two PRs every time we want to update Walid's work, but maybe is a good practice? (3) would involve a **big** PR one day which we might want to avoid

magnoxemo:
closing it as not necessary right now

PR #1130, opened 2025-05-30
Add Double-Down and Embree to improve raytracing performance on CAD geometries

### LINKED ISSUE #1129: Add Double-Down and Embree as submodules
## Reason
Particle transport performance on triangulated CAD geometry in Cardinal is currently limited by the performance of MOAB's ray tracer. We should move away from ray tracing in MOAB and enable mixed precision ray tracing in DAGMC via Double-Down + Embree. 

## Design
Embree and Double-Down should be added as dependencies that are required for DAGMC builds. The build system would need to be modified to include these dependencies and enable the use of Double-Down in DAGMC.

## Impact
Large performance and scalability improvements. The CIVET recipe would need to be updated to account for the new submodules.

nuclearkevin:
Closes #1129

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,12 @@ endif
 DAGMC_BUILDDIR := $(CARDINAL_DIR)/build/DAGMC
 DAGMC_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
 
+DOUBLEDOWN_BUILDDIR := $(CARDINAL_DIR)/build/double-down
+DOUBLEDOWN_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
+
+EMBREE_BUILDDIR := $(CARDINAL_DIR)/build/embree
+EMBREE_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
+
 MOAB_BUILDDIR := $(CARDINAL_DIR)/build/moab
 MOAB_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
 
@@ -234,11 +240,19 @@ INSTALLABLE_DIRS   := test/tests->tests tutorials
 ifeq ($(ENABLE_DAGMC), yes)
   ENABLE_DAGMC     := ON
   include          $(CARDINAL_DIR)/config/moab.mk
+	include          $(CARDINAL_DIR)/config/embree.mk
+	include          $(CARDINAL_DIR)/config/double_down.mk
   include          $(CARDINAL_DIR)/config/dagmc.mk
 else
 
 build_dagmc:
 	$(info Skipping DagMC build because ENABLE_DAGMC is not set to 'yes')
+
+build_doubledown:
+	$(info Skipping Double-Down build because ENABLE_DAGMC is not set to 'yes')
+
+build_embree:
+	$(info Skipping Embree build because ENABLE_DAGMC is not set to 'yes')
 
 build_moab:
 	$(info Skipping MOAB build because ENABLE_DAGMC is not set to 'yes')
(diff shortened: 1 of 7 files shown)


nuclearkevin:
@pshriwise pinging you to make you aware of the change, just in case I missed something when modifying Cardinal's buildsystem to add Double-Down and Embree.

pshriwise:
> @pshriwise pinging you to make you aware of the change, just in case I missed something when modifying Cardinal's buildsystem to add Double-Down and Embree.

Thanks for the ping! Looks nice and I hope it helps performance going forward!

nuclearkevin:
I can't seem to get the two MOAB skinner scaling tests too pass with Embree + Double-Down no matter what I do. So, I've switched them over to optional dependencies which are disabled by default until we figure this out. If users wish to speed up ray tracing operations, they can `export ENABLE_DOUBLE_DOWN = yes`.

Makefile
```diff
@@ -289,7 +322,10 @@ endif
 ifeq ($(ENABLE_OPENMC), yes)
   ADDITIONAL_LIBS += -L$(OPENMC_LIBDIR) -lopenmc -lhdf5_hl
   ifeq ($(ENABLE_DAGMC), ON)
-    ADDITIONAL_LIBS += -ldagmc -lMOAB
+    ADDITIONAL_LIBS += -ldagmc -lMOAB -lembree4 -ldd
```
aprilnovak: should `-lembree4 -ldd` be added 2x here? I feel like line 325 should not have them

doc/content/cardinal_prereqs.md
```diff
@@ -4,3 +4,5 @@ whether each library is enabled:
 - [OpenMC](https://docs.openmc.org/en/stable/usersguide/install.html#prerequisites)
 - [NekRS](https://github.com/Nek5000/nekRS/tree/next)
 - [DAGMC](https://svalinn.github.io/DAGMC/install/dependencies.html)
+- [Double-Down](https://double-down.readthedocs.io/en/latest/usersguide/installation.html)
```
aprilnovak: This link doesn't look to point at the prereqs for double down. Is there a different URL we can link to here?

unit/Makefile
```diff
@@ -284,7 +315,10 @@ endif
 ifeq ($(ENABLE_OPENMC), yes)
   ADDITIONAL_LIBS += -L$(OPENMC_LIBDIR) -lopenmc -lhdf5_hl
   ifeq ($(ENABLE_DAGMC), ON)
-    ADDITIONAL_LIBS += -ldagmc -lMOAB
+    ADDITIONAL_LIBS += -ldagmc -lMOAB -lembree4 -ldd
```
aprilnovak: same comment here about libraries

doc/content/cardinal_prereqs.md
nuclearkevin: We could point to the pre-requisites header on the DD Github: https://github.com/pshriwise/double-down?tab=readme-ov-file#prerequisites

aprilnovak: yes, that's better. although is this a fork?

nuclearkevin: It is not. Double-Down only exists on Patrick's Github

Makefile
nuclearkevin: You're 100% right - thanks for catching this!

nuclearkevin:
> Is there anything to worry about with all the intermittent failures? Looks like there might be a problem when pulling the new submodules added? Most of the test recipes now fail that step on the first attempt.

I don't know to be honest, but it doesn't seem to prevent tests from passing. Did you see something similar when you added Nek CI as a submodule?

aprilnovak:
No, this seems specific to this PR. I'll merge this now, but let's keep an eye on it. We could always have CIVET be building Cardinal without embree/double-down if it's a persistent issue.

PR #1133, opened 2025-06-05
Changes to use displaced mesh with OpenMC

meltawila:
These are the changes from #907 after rebasing and resolving conflicts. 

(Ref #906)


include/base/OpenMCCellAverageProblem.h
```diff
@@ -20,6 +20,8 @@
 
 #include "OpenMCProblemBase.h"
 #include "SymmetryPointGenerator.h"
+#include "OpenMCVolumeCalculation.h"
```
aprilnovak: why is this added?

include/tallies/CellTally.h
```diff
@@ -20,9 +20,13 @@
 
 #include "TallyBase.h"
 #include "OpenMCCellAverageProblem.h"
+#include "MooseMesh.h"
```
aprilnovak: You don't need this - you are already #including the base class, `TallyBase`. TallyBase has `#include MooseMesh.h`. YOu can think of `#include` as literally copy-paste of that file's contents into the location where the include command is.

include/tallies/MeshTally.h
```diff
@@ -70,7 +74,7 @@ class MeshTally : public TallyBase
    * each element. This function performs as many checks as possible to ensure that the meshes
    * are indeed identical.
    */
-  void checkMeshTemplateAndTranslations() const;
```
aprilnovak, line 73: why?

include/tallies/MeshTally.h
```diff
@@ -98,6 +102,12 @@ class MeshTally : public TallyBase
   /// OpenMC unstructured mesh instance for use with mesh tallies
   openmc::LibMesh * _mesh_template;
 
+  /// Whether the skinned mesh should be generated from a displaced mesh
+  const bool & _use_displaced;
+
+  /// Moose mesh
+  MooseMesh & getMooseMesh();
```
aprilnovak: This code is object-oriented. You automatically inherit everything (except `private` methods) from the base class. This may do weird things. Remove from this class and `CellTally`

include/tallies/CellTally.h
```diff
@@ -20,9 +20,13 @@
...
 
 #include "openmc/tallies/filter_cell.h"
 
+class OpenMCCellAverageProblem;
+class MooseMesh;
```
aprilnovak: why are these here? when you `#include` the header files, you don't need these. These are forward declarations, which may be something used when you DON'T include a header file. Remove

include/tallies/TallyBase.h
```diff
@@ -340,6 +341,9 @@ class TallyBase : public MooseObject
   /// Suffixes to apply to 'tally_name' in order to name the fields in the 'output'.
   std::vector<std::string> _output_name;
 
+  /// Whether the skinned mesh should be generated from a displaced mesh
+  const bool & _use_displaced;
```
aprilnovak: same comment here -  you are already inheriting these in the derived classes. you are being redundant in a lot of places

src/tallies/TallyBase.C
```diff
@@ -80,6 +81,9 @@ TallyBase::validParams()
 
   params.registerBase("Tally");
   params.registerSystemAttributeName("Tally");
+  params.addParam<bool>("use_displaced_mesh",
```
aprilnovak: you automatically inherit this in the daughter classes. Apply same feedback here

src/tallies/CellTally.C
```diff
@@ -182,9 +183,24 @@ CellTally::checkCellMappedSubdomains()
   }
 }
 
+MooseMesh &
+CellTally::getMooseMesh()
```
aprilnovak: Instead of duplicating this code here, can we instead call `getMooseMesh()` from the `OpenMCCellAverageProblem`? You already have access to `_openmc_problem` in this object

src/tallies/MeshTally.C
```diff
@@ -250,19 +251,34 @@ MeshTally::storeResultsInner(const std::vector<unsigned int> & var_numbers,
   return total;
 }
 
+MooseMesh &
+MeshTally::getMooseMesh()
```
aprilnovak: never need to duplicate code like this, it would be ok to put in the base class and then automatically inherit. but see my other comment instead which is better solution, to call the method in `OpenMCCellAverageProblem`

src/userobjects/OpenMCVolumeCalculation.C
```diff
@@ -65,35 +66,25 @@ OpenMCVolumeCalculation::OpenMCVolumeCalculation(const InputParameters & paramet
...
 }
-
 openmc::Position
 OpenMCVolumeCalculation::position(const Point & pt) const
 {
   openmc::Position p {pt(0), pt(1), pt(2)};
   return p;
 }
 
+MooseMesh &
```
aprilnovak: Get rid of this function - you can instead call the equivalent function in `_openmc_problem` since you have access to it in this class

test/tests/neutronics/dagmc/cell_tallies/tests
```diff
@@ -56,4 +56,14 @@
     requirement = "The system shall warn the user when there is a mismatch between the mesh mirror and the initial OpenMC DAGMC geometry."
     required_objects = 'MoabSkinner'
   []
+  [with_movement]
+    type = Exodiff
+    input = with_movement.i
+    exodiff = with_movement_out.e
+    prereq = with_scaling
+    cli_args = 'Outputs/hide="cell_instance"'
+    requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when the mesh is moving."
```
aprilnovak: Can you please make the `requirement` section of your tests more descriptive? Describe what the test is doing and how you know that the answer is meaningful. For example, this is from one of the other test cases in Cardinal - it explains how we know that the test is giving good results which are worth golding against for all time.

```
  [first_order_temperature]
    type = Exodiff
    input = nek.i
    exodiff = nek_out.e
    requirement = "The nekRS temperature solution shall be accurately reconstructed on "
                  "the nekRSMesh with a first-order surface transfer. By setting an intial condition "
                  "for temperature on the nekRS side and then setting 'solver = none', we show "
                  "that the max/min error in that reconstructed temperature compared to a MOOSE "
                  "function of the same form is O(1e-16)."
    required_objects = 'NekRSProblem'
  []
```

aprilnovak: what is "identical" relative to, in this case, for instance?

test/tests/neutronics/dagmc/mesh_tallies/tests
```diff
@@ -72,7 +72,8 @@
     exodiff = with_movement_out.e
     prereq = move_mesh
     cli_args = 'Outputs/hide="cell_instance"'
-    requirement = "The system shall give identical Monte Carlo solution (direct mesh tallies and k) when the mesh is moving."
+    requirement = "The OpenMC solution (mesh tallies and k-effective) shall remain within statistical uncertainty when the mesh is displaced,"
```
aprilnovak: what does the "within statistical ucnertainty" part specifically mean? These tests don't check that `k`, etc. are within 3\sigma of the mean, they just check the actual mean.

meltawila: I just mean that a difference like 1.33 vs 1.32 in k_eff between the 2 iterations in the test isn't related to mesh movement when the uncertainty is on the order of 500 pcm. I can leave that part out if it's considered trivial. The focus of the test is not to compare the numerical results before/after movement since we are running a low number of particles, but to make sure that all mesh elements are mapped correctly to the corresponding OpenMC cells during the movement and that there are no tally events that get lost due to movement

src/userobjects/OpenMCVolumeCalculation.C
```diff
@@ -104,6 +88,32 @@ OpenMCVolumeCalculation::resetVolumeCalculation()
 void
 OpenMCVolumeCalculation::initializeVolumeCalculation()
 {
+
+  BoundingBox box = MeshTools::create_bounding_box(_openmc_problem->getMooseMesh().getMesh());
+  if (_fe_problem.getDisplacedProblem() != nullptr)
```
aprilnovak: add comment if needed to explain

src/base/OpenMCCellAverageProblem.C
```diff
@@ -442,6 +448,32 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
   }
 }
 
+const MooseMesh &
+OpenMCCellAverageProblem::getMooseMesh() const
+{
+  if (_use_displaced && !_displaced_problem && !_first_transfer)
+    mooseWarning("Displaced mesh was requested but the displaced problem does not exist. "
```
aprilnovak: Please add a test that triggers this warning, so we're sure it does get displayed to users as expected. Your test specs will look like following. The `cli_args` allows you to write which command-line parameters you want to run with your test. `--error` will cause all warnings to become errors so that you can catch them.

```
[test]
  type = RunException
  input = my_input.i
  cli_args = '--error'
  expect_err = "Displaced mesh was requested but the displaced problem does not exist."
  required_objects = 'OpenMCCellAverageProblem'
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -442,6 +448,32 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
...
+                 "Regular mesh will be returned");
```
```suggestion
                 "Un-displaced mesh will be returned");
```

aprilnovak: I think this is a bit more descriptive

src/base/OpenMCCellAverageProblem.C
```diff
@@ -442,6 +448,32 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
...
+                                                                  : mesh());
+  return m;
+}
+
+MooseMesh &
+OpenMCCellAverageProblem::getMooseMesh()
+{
+  if (_use_displaced && !_displaced_problem && !_first_transfer)
+    mooseWarning("Displaced mesh was requested but the displaced problem does not exist. "
+                 "Regular mesh will be returned");
```
aprilnovak, line 468:
```suggestion
                 "Un-displaced mesh will be returned");
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2596,7 +2629,10 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
       {
         if (_volume_calc)
           _volume_calc->resetVolumeCalculation();
-
+        if (_use_displaced)
+        {
```
```suggestion
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2596,7 +2629,10 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
...
+          _displaced_problem->updateMesh();
+        }
```
```suggestion
```

src/tallies/CellTally.C
```diff
@@ -191,6 +192,10 @@ CellTally::checkCellMappedSubdomains()
 std::vector<OpenMCCellAverageProblem::cellInfo>
 CellTally::getTallyCells() const
 {
+  if (_use_displaced)
+  {
```
aprilnovak:
```suggestion
```

src/tallies/CellTally.C
```diff
@@ -191,6 +192,10 @@ CellTally::checkCellMappedSubdomains()
...
+    _openmc_problem.getDisplacedProblem()->updateMesh();
+  }
```
aprilnovak:
```suggestion
```

test/tests/neutronics/dagmc/cell_tallies/with_movement.i
```diff
@@ -0,0 +1,153 @@
...
+  type = OpenMCCellAverageProblem
+  verbose = true
+  scaling = ${scale}
+
+  temperature_blocks = '0 1'
+  cell_level = 0
+  power = 16.0
+
+  skinner = moab
+  fixed_mesh = false
```
aprilnovak: I think when you run this file you should be seeing a warning that this parameter is unused, correct? let's remove it

src/tallies/TallyBase.C
```diff
@@ -337,6 +342,10 @@ TallyBase::storeResults(const std::vector<unsigned int> & var_numbers,
                         unsigned int global_score,
                         const std::string & output_type)
 {
+  if (_use_displaced)
```
aprilnovak: This currently would require the user to have their input like

```
[Problem]
  use_displaced_mesh = true

  [Tallies]
    [c]
      type = CellTally
      use_displaced_mesh = true
    []
  []
[]
```

duplicate input syntax can be error-prone and tedious. Recommend to add an accessor method to `OpenMCCellAverageProblem` to return to you `_use_displaced` and then use it instead of adding another input parameter.

include/base/OpenMCCellAverageProblem.h
```diff
@@ -413,6 +413,11 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
   /// Spatial dimension of the Monte Carlo problem
   static constexpr int DIMENSION{3};
 
+  /// Get a modifyable reference to the Moose mesh
```
aprilnovak: for the doxygen documentation, follow the patterns and make sure you write out what the return types are, if any

include/base/OpenMCCellAverageProblem.h
```diff
@@ -413,6 +413,11 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
...
+  virtual MooseMesh & getMooseMesh();
+
+  virtual const MooseMesh & getMooseMesh() const;
```
aprilnovak, line 420: add doxygen docs

test/tests/neutronics/dagmc/displaced_problems/gold/openmc_out.csv
```diff
@@ -0,0 +1,4 @@
+time,mapped_vol_left,mapped_vol_right,moose_vol_left,moose_vol_right,openmc_vol_left,openmc_vol_right
```
aprilnovak: this file does not seem to be actually used anywhere

test/tests/neutronics/dagmc/displaced_problems/move.i
```diff
@@ -0,0 +1,75 @@
+scale = 100.0
```
aprilnovak: this file does not seem to be used anywhere

src/userobjects/MoabSkinner.C
```diff
@@ -280,15 +280,12 @@ void
...
 {
   _console << "Skinning geometry into " << _n_temperature_bins << " temperature bins, "
-           << _n_density_bins << " density bins, and " << _n_block_bins << " block bins... ";
+           << _n_density_bins << " density bins, and " << _n_block_bins << " block bins... "
+           << std::endl;
 
   // Clear MOAB mesh data from last timestep
   reset();
 
-  if (_use_displaced)
```
aprilnovak, line 288: ok, this one does have to be here -- because we do technically allow `MoabSkinner` to run without an `OpenMCCellAverageProblem`. But we do not need it in the tally objects because those cannot exist without an `OpenMCCellAverageProblem`. Just noting as FYI

PR #1142, opened 2025-06-15
Disable Double-Down and Embree by default.

nuclearkevin:
Disabling Embree and Double-Down by default due to some build errors. Currently working on resolving these.

Ref. #1129


aprilnovak:
You'll need to apply the same changes to the unit makefile

PR #1144, opened 2025-06-20
Add the Embree+Double-Down change to June 2025's news

nuclearkevin:
Ref. #1129


doc/content/news/june2025.md
```diff
@@ -189,3 +191,12 @@ to the new syntax:
...
 []
 ```
+
+## Performance Improvements for Particle Transport on CAD Geometry
+
+We added two optional dependencies to Cardinal: [Double-Down](https://github.com/pshriwise/double-down) and [Embree](https://github.com/RenderKit/embree).
+Embree is a high-performance single-precision ray tracing library and Double-Down is a wrapper library which enables double-precision ray-triangle
+intersection tests in Embree. This addition results in substantial gains in single-thread performance and threaded scalability for particle transport
+on triangulated CAD geometries compared to native MOAB raytracing. Cardinal can be built with both of these libraries by exporting `ENABLE_DOUBLE_DOWN=yes`.
+At present, we are still ironing out some compilation issues - if you experience difficulties building Cardinal with Double-Down and Embree please post on
```
aprilnovak:
```suggestion
At present, these capabilities are disabled by default due to compilation issues on some (but not all) platforms - if you experience difficulties building Cardinal with Double-Down and Embree please post on
```

PR #1147, opened 2025-06-21
Add a lookahead relative error marker

### LINKED ISSUE #1146: Add an `ErrorFractionMarker` equivalent which mitigates cyclic refinement
## Reason
Refinement algorithms and spatial indicators investigated for Monte Carlo AMR are often prone to cyclic refinement behaviour due to the integration of statistical relative error metrics late in the algorithm. We should add an equivalent to `ErrorFractionMarker` which estimates the statistical error after an element is split, and only refines if this post-split relative error is below the user-specified refinement tolerance. 

## Design
This addition should be split into two MOOSE objects. The first would be a `StatRelErrorIndicator`, which makes the relative error of a tally score available to the indicator system. The second addition would be a `ErrorFractionLookAheadMarker`, which marks elements for refinement in a similar manner as the `ErrorFractionMarker` when it comes to a spatial error estimate. It would incorporate statistical error in two places: the first being the determination of the upper spatial error bound, and the second is during the marking process. 

1. Elements only contribute to the upper spatial error bound if their statistical error is below the user-defined threshold. 

2. Elements are only marked for refinement if their estimated statistical relative error (post-refinement) is below the user-provided threshold. The statistical relative error post refinement is calculated with $e_{i+1}=\sqrt{N_{c}}e_{i}$, where $e_{i+1}$ is the error post-refinement, $N_{c}$ is the number of child elements, and $e_{i}$ is the current statistical error.

## Impact
These additions would improve the suite of capabilities for using adaptivity with OpenMC unstructured mesh tallies, and entirely remove cyclic refinement due to statistical relative error.

nuclearkevin:
This also adds the capability to use statistical relative error of a tally field as an indicator, and the inverse optical depth.

Closes #1146

### DIFF
--- a/src/indicators/StatRelErrorIndicator.C
+++ b/src/indicators/StatRelErrorIndicator.C
@@ -33,13 +33,12 @@ StatRelErrorIndicator::validParams()
                              "associated with a tally variable.");
   params.addRequiredParam<MooseEnum>(
       "score", getSingleTallyScoreEnum(), "The tally score used for the relative error.");
-  params.addParam<unsigned int>(
-      "ext_filter_bin",
-      0,
-      "The filter bin for the case where any filters are added to this "
-      "tally with [Filters] (bin indices start at 0). This parameter "
-      "should be specified if you wish to extract the relative error "
-      "of a different non-spatial tally bin.");
+  params.addParam<unsigned int>("ext_filter_bin",
+                                0,
+                                "The filter bin for the case where any filters are added to this "
+                                "tally with [Filters] (bin indices start at 0). This parameter "
+                                "should be specified if you wish to extract the relative error "
+                                "of a different non-spatial tally bin.");
 
   return params;
 }
@@ -60,7 +59,8 @@ StatRelErrorIndicator::StatRelErrorIndicator(const InputParameters & parameters)
   if (!_openmc_problem->hasOutput(score, "rel_error"))
     mooseError(
         "The problem does not contain any tallies that output the relative error for the score " +
-        std::string(getParam<MooseEnum>("score")) + "! Please ensure that your problem "
+        std::string(getParam<MooseEnum>("score")) +
+        "! Please ensure that your problem "
         "includes 'output = unrelaxed_tally_rel_err'");
 
   // Check to ensure the reaction rate / flux variables are CONSTANT MONOMIALS.
--- a/src/auxkernels/FDTallyGradAux.C
+++ b/src/auxkernels/FDTallyGradAux.C
@@ -35,13 +35,12 @@ FDTallyGradAux::validParams()
       "score",
       getSingleTallyScoreEnum(),
       "The tally score this auxkernel should approximate the gradient of.");
-  params.addParam<unsigned int>(
-      "ext_filter_bin",
-      0,
-      "The filter bin for the case where any filters are added to this "
-      "tally with [Filters] (bin indices start at 0). This parameter "
-      "should be specified if you wish to extract the relative error "
-      "of a different non-spatial tally bin.");
+  params.addParam<unsigned int>("ext_filter_bin",
+                                0,
+                                "The filter bin for the case where any filters are added to this "
+                                "tally with [Filters] (bin indices start at 0). This parameter "
+                                "should be specified if you wish to extract the relative error "
+                                "of a different non-spatial tally bin.");
 
   params.addRelationshipManager("ElementSideNeighborLayers",
                                 Moose::RelationshipManagerType::ALGEBRAIC |


doc/content/source/indicators/StatRelErrorIndicator.md
```diff
@@ -0,0 +1,17 @@
+# StatRelErrorIndicator
+
+## Description
+
+The `StatRelErrorIndicator` provides the relative error associated with the tally score to the indicator
+system for use in adaptivity. A tally `score` must be specified, and if any external filter bins have been
+added through the use of the [filter system](AddFilterAction.md) `ext_filter_bin` index should be provided.
+Additionally, the tally scoring `score` must be outputting the relative error of the tally.
```
aprilnovak: can you add a sentence describing how to do this, in case someone doesn't know?

doc/content/source/markers/ErrorFractionLookAheadMarker.md
```diff
@@ -0,0 +1,23 @@
+# ErrorFractionLookAheadMarker
+
+## Description
+
+`ErrorFractionLookAheadMarker` extends the [ErrorFractionMarker](ErrorFractionMarker.md) from MOOSE
+to include the statistical elative error from fields in stochastic solvers (i.e OpenMC). If the
```
aprilnovak:
```suggestion
to include the statistical relative error from fields in stochastic solvers (i.e OpenMC). If the
```

doc/content/source/markers/ErrorFractionLookAheadMarker.md
```diff
@@ -0,0 +1,23 @@
+# ErrorFractionLookAheadMarker
+
+## Description
+
+`ErrorFractionLookAheadMarker` extends the [ErrorFractionMarker](ErrorFractionMarker.md) from MOOSE
+to include the statistical relative error from fields in stochastic solvers (i.e OpenMC). If the
```
aprilnovak:
```suggestion
to include the statistical relative error from fields in stochastic solvers (i.e. OpenMC). If the
```

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -43,12 +43,15 @@ ElementOpticalDepthIndicator::validParams()
       "the "
       "minimum vertex separation (min), the maximum vertex separation (max), and the cube root of "
       "the element volume (cube_root).");
+  params.addParam<bool>("invert", false, "Whether the optical depth is stored as 1 / OD or OD.");
```
aprilnovak:
```suggestion
  params.addParam<bool>("invert", false, "Whether the optical depth is computed as the optical depth (false) or the inverse of the optical depth (true).");
```

aprilnovak: OD is outer diameter for thermal-fluids people, so this might be confusing

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -101,7 +104,7 @@ ElementOpticalDepthIndicator::computeIndicator()
   for (const auto & var : _scalar_fluxes)
     scalar_flux += (*var)[0];
 
-  auto od = scalar_flux < libMesh::TOLERANCE ? 0.0 : rxn_rate / scalar_flux;
+  auto od = scalar_flux < (libMesh::TOLERANCE * libMesh::TOLERANCE) ? 0.0 : rxn_rate / scalar_flux;
```
aprilnovak: why this change?

src/indicators/ElementOpticalDepthIndicator.C
```diff
@@ -119,7 +122,10 @@ ElementOpticalDepthIndicator::computeIndicator()
       break;
   }
 
-  _field_var.setNodalValue(od);
+  if (_invert && scalar_flux > libMesh::TOLERANCE * libMesh::TOLERANCE)
```
aprilnovak: To me, this would read more naturally if the check were done on `od`. Because in principle, the mesh spacing could be what causes very small values, not necessarily always the flux.

src/indicators/StatRelErrorIndicator.C
```diff
@@ -0,0 +1,92 @@
...
+    paramError("score",
+               "The problem does not contain any score named " +
+                   std::string(getParam<MooseEnum>("score")) +
+                   "! Please "
+                   "ensure that one of your [Tallies] is scoring the requested reaction rate.");
+
+  if (!_openmc_problem->hasOutput(score, "rel_error"))
+    mooseError(
+        "The problem does not contain any tallies that output the relative error for the score " +
+        std::string(getParam<MooseEnum>("score")) + "!");
```
aprilnovak:
```suggestion
        std::string(getParam<MooseEnum>("score")) + "! Please ensure that your problem includes 'output = unrelaxed_tally_rel_err'");
```

src/indicators/StatRelErrorIndicator.C
```diff
@@ -0,0 +1,92 @@
...
+InputParameters
+StatRelErrorIndicator::validParams()
+{
+  InputParameters params = OpenMCIndicator::validParams();
+  params.addClassDescription("An Indicator which reports the maximum relative error of all bins "
+                             "associated with a tally variable.");
+  params.addRequiredParam<MooseEnum>(
+      "score", getSingleTallyScoreEnum(), "The tally score used for the relative error.");
+  params.addParam<unsigned int>(
+      "ext_filter_bin",
```
aprilnovak: Is this for energy/angle filters? Can the string description be expanded to elaborate more on this parameter?

test/tests/neutronics/indicators/rel_err/openmc.i
```diff
@@ -0,0 +1,51 @@
...
+[Adaptivity]
+  [Indicators/stat_err]
+    type = StatRelErrorIndicator
+    score = 'kappa_fission'
+  []
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  temperature_blocks = '100'
```
aprilnovak: we can remove this line and then also remove `temp` from `hide` in the outputs

test/tests/neutronics/indicators/rel_err/tests
```diff
@@ -0,0 +1,40 @@
...
+    requirement = 'The system shall error if no tallies are storing the relative error of the provided score.'
+    expect_err = 'The problem does not contain any tallies that output the relative error for the score kappa_fission!'
+    mesh_mode = 'replicated'
+    capabilities = 'openmc'
+  []
+  [invalid_bin]
+    type = RunException
+    input = 'openmc.i'
+    cli_args = 'Adaptivity/Indicators/stat_err/ext_filter_bin=1'
+    requirement = 'The system shall error if the provided tally external filter bin is greater than the total number of external filter bins..'
```
aprilnovak:
```suggestion
    requirement = 'The system shall error if the provided tally external filter bin is greater than the total number of external filter bins.'
```

src/indicators/ElementOpticalDepthIndicator.C
nuclearkevin: Whoops, that shouldn't be in here. I was doing some debugging on an older branch and changed this. Performing the check on `libMesh::TOLERANCE` is conservative in catching divide by zeros.

src/indicators/ElementOpticalDepthIndicator.C
nuclearkevin: Agreed, I'll change this.

src/indicators/StatRelErrorIndicator.C
nuclearkevin: Yeah, this is for any filters added with the `[Filters]` system that are applied to the tally scoring `score`. As we support functional expansion filters, there isn't an easy way to return a "bulk" relative error for the entire tally score so the user needs to pick a filter bin.

nuclearkevin: I also took the liberty of adding the extended description to `FDTallyGradAux`, which had the same parameter and docstring.

PR #1151, opened 2025-06-26
Update check_modules.mk

jfspecht:
Fixing typo in Geochemistry
Adding Solid Mechanics

### DIFF
--- a/config/check_modules.mk
+++ b/config/check_modules.mk
@@ -10,6 +10,9 @@ $(info $nCardinal is compiled with the following MOOSE modules)
 ifeq ($(CHEMICAL_REACTIONS), yes)
   $(info $(SPACE) CHEMICAL_REACTIONS)
 endif
+ifeq ($(INCLUDE_COMBINED), yes)
+  $(info $(SPACE) INCLUDE_COMBINED)
+endif
 ifeq ($(CONTACT), yes)
   $(info $(SPACE) CONTACT)
 endif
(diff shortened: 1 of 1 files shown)


aprilnovak:
Since we're here, would you also please check if there's any other newer ones which we forgot to add from `moose/modules/modules.mk`? The full list is in that file

jfspecht:
Deleted those line and I do not believe there are any more modules that were added that I did not include. Feel free to double check (https://mooseframework.inl.gov/modules/index.html), but I am pretty sure I got all of them.

aprilnovak:
Did you check the list here? https://github.com/idaholab/moose/blob/next/modules/modules.mk

config/check_modules.mk
```diff
@@ -10,15 +10,18 @@ $(info $nCardinal is compiled with the following MOOSE modules)
 ifeq ($(CHEMICAL_REACTIONS), yes)
   $(info $(SPACE) CHEMICAL_REACTIONS)
 endif
-ifeq ($(COMBINED), yes)
-  $(info $(SPACE) COMBINED)
+ifeq ($(INCLUDE_COMBINED), yes)
```
aprilnovak: can you explain this for me?

jfspecht: The only instance of `COMBINED` in `https://github.com/idaholab/moose/blob/next/modules/modules.mk` is in `INCLUDE_COMBINED`. It seems like `COMBINED` follows a different format

aprilnovak: I actually think we can remove the lines having to do with combined altogether. This module is just the sum of all the other ones, which we'd have enabled with `ALL_MODULES := yes` in the makefile as opposed to turning on combined. Would you please remove these lines?

aprilnovak:
For the test suite to start, you'll need to accept the invitation to join `cardinal-devs`. After that, can you please force-push your branch so that the tests will be triggered?

PR #1174, opened 2025-07-19
Update MOOSE and Double-Down

nuclearkevin:


config/double_down.mk
```diff
@@ -12,7 +12,6 @@ $(DOUBLEDOWN_BUILDDIR)/Makefile: build_moab build_embree | $(DOUBLEDOWN_DIR)/CMa
 	-DCMAKE_INSTALL_PREFIX=$(DOUBLEDOWN_INSTALL_DIR) \
 	-DCMAKE_INSTALL_LIBDIR=lib \
 	-DCMAKE_INSTALL_MESSAGE=LAZY \
-	-DCMAKE_CXX_FLAGS="-Wno-error=reorder" \
```
aprilnovak, line 15: this is a useful compile flag that can detect sneaky bugs. Is it a big undertaking to fix any conflicts in double_down? We have made similar PRs into OpenMC in the past, for instance, to fix reordering errors.

nuclearkevin: I had to _disable_ this error to get Double-Down to build initially as it did not obey constructor initialization rules. I poked Patrick to go and fix the reordering issues in Double-Down at the fusion neutronics workshop, which just got merged into Double-Down last week.

This change re-enables throwing the error (removing `Wno-error`), and we can go and fix any reordering errors in Double-Down moving forwards.

aprilnovak: ah whoops - sorry. Not enough coffee yet

PR #1178, opened 2025-07-26
Avoid the creation of a graveyard for mixed CSG+DAGMC models & tag DAGMC+CSG tests

### LINKED ISSUE #1176: Disable forced skinner graveyard creation when the DAGMC geometry is embedded in a CSG cell
## Reason
At present, we always create a graveyard universe surrounding the DAGMC geometry when the skinner is used. This may cause problems if the user embeds their DAGMC geometry inside a CSG universe (i.e. to leave room for thermal expansion or to use the implicit compliment to fill the gap between DAG cells and CSG cells) as the graveyard could kill particles in the CSG region. A current workaround is to use `graveyard_scale_inner` and `graveyard_scale_outer` to place the graveyard outside of the maximum bounds of the CSG region, but this is rather clunky and may take some trial and error from the user to find a bounding box that works (for complicated CSG regions).

## Design
We should only force the addition of a graveyard in the skinner if the geometry is pure DAGMC. If the OpenMCCellAverageProblem detects that the DAGMC universe is embedded in a CSG cell (which it already does when setting up the skinner), it should avoid calling `_skinner->setGraveyard(...)` and respect the value selected by the user in the `MOABSkinner` input file block.

## Impact
Better experience for users using hybrid CSG/DAGMC geometry, more control over skinning behaviour.

### LINKED ISSUE #1177: Allow mixed CSG-DAGMC models when using the MOABSkinner
## Reason
It would be nice if the `MOABSkinner` would support mixed CSG-DAGMC models. This would allow for the specification of a surface mesh where the geometry should be re-generated (for multiphysics feedback), while the rest of the problem is set up as CSG (for accuracy / performance). This would also allow for the use of custom boundary conditions on a CSG cell that contains the DAGMC universe.

## Design
After the DAGMC geometry is regenerated from the mesh mirror, it should be placed in the CSG cell that contains it. The methods used to reset the OpenMC geometry should be changed so they avoid deleting / scrambling the CSG cells in the process. There would be some limitations, namely:

- CSG cells cannot be mapped to the MOOSE mesh mirror when using a skinner. This is to prevent the generation of erroneous DAGMC geometry which may overlap with pre-existing CSG geometry in the OpenMC model. This restriction can be relaxed in a future PR by modifying the `MOABSkinner`.
- The OpenMC geometry can only contain a single DAGMC universe, otherwise the universe to overwrite with skinned geometry cannot be automatically detected. This can also be relaxed in a future PR by allowing the user to specify a DAGMC universe for skinning.
- The DAGMC universe can only be contained within a single CSG cell, and that cell must exist exclusively in the root universe. This avoids the duplication of DAGMC geometry as this duplicated geometry may not be skinned. Similarly, the DAGMC geometry is not allowed in lattices when using a `MOABSkinner`.

## Impact
New feature, added flexibility for multiphysics with adaptive temperature feedback.

**Additional context**
This issue is being created after-the-fact for SQA purposes.

nuclearkevin:
Closes #1176 
Closes #1177


nuclearkevin:
I've re-enabled the non-deterministic DAGMC+CSG test as it's the only good way to test #1176. I'll try to get it passing on CIVET.

nuclearkevin:
Thinking about it some more, worst-case scenario we can just use a `RunApp` test to verify that this capability works. Without the generation of a graveyard, there will be no boundary conditions. The problem will fail catastrophically from particles not undergoing collisions at all.

test/tests/neutronics/dagmc/with_csg/geometry.xml
```diff
@@ -3,9 +3,9 @@
   <cell id="24" fill="2" region="20 -21 22 -23 24 -25" universe="1" />
   <surface boundary="reflective" coeffs="-12.5" id="20" type="x-plane" />
   <surface boundary="reflective" coeffs="87.5" id="21" type="x-plane" />
-  <surface boundary="reflective" coeffs="-12.5" id="22" type="y-plane" />
-  <surface boundary="reflective" coeffs="37.5" id="23" type="y-plane" />
-  <surface boundary="reflective" coeffs="-12.5" id="24" type="z-plane" />
-  <surface boundary="reflective" coeffs="12.5" id="25" type="z-plane" />
+  <surface boundary="vacuum" coeffs="-12.5" id="22" type="y-plane" />
```
aprilnovak, line 6: Can you explain what the motivation is for changing this boundary condition?

nuclearkevin: Initially, I swapped the boundary condition out in an attempt to make the test deterministic (which didn't work - removing the graveyard was the fix to that problem). I left the change in to speed up the test - it took around 60 seconds to run previously due to the high particle count and the reflective boundary conditions.

PR #1184, opened 2025-08-01
WIP: Criticality search

aprilnovak:
This PR adds an example of performing a criticality search using Cardinal, starting by the easier form of criticality maneuver (changing a material density).

### DIFF
--- a/include/criticality/CriticalitySearchBase.h
+++ b/include/criticality/CriticalitySearchBase.h
@@ -1,0 +1,43 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
+/*                  (c) 2021 UChicago Argonne, LLC                  */
+/*                        ALL RIGHTS RESERVED                       */
+/*                                                                  */
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#pragma once
+
+#include "MooseObject.h"
+#include "OpenMCCellAverageProblem.h"
+
+class CriticalitySearchBase : public MooseObject
+{
+public:
+  static InputParameters validParams();
+
+  CriticalitySearchBase(const InputParameters & parameters);
+
+  virtual void updateOpenMCModel(const Real & input) = 0;
+
+  virtual void searchForCriticality();
+
+protected:
+  virtual std::string quantity() const = 0;
+
+  const Real & _maximum;
+  const Real & _minimum;
+  const Real & _tolerance;
+
+  OpenMCCellAverageProblem & _openmc_problem;
+};
--- a/src/actions/AddCriticalitySearchAction.C
+++ b/src/actions/AddCriticalitySearchAction.C
@@ -1,0 +1,59 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
+/*                  (c) 2021 UChicago Argonne, LLC                  */
+/*                        ALL RIGHTS RESERVED                       */
+/*                                                                  */
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "AddCriticalitySearchAction.h"
+#include "OpenMCCellAverageProblem.h"
+#include "CriticalitySearchBase.h"
+
+registerMooseAction("CardinalApp", AddCriticalitySearchAction, "add_criticality_search");
+
+InputParameters
+AddCriticalitySearchAction::validParams()
+{
+  auto params = MooseObjectAction::validParams();
+  params.addClassDescription("Adds a criticality search for OpenMC");
+  return params;
+}
+
+AddCriticalitySearchAction::AddCriticalitySearchAction(const InputParameters & parameters)
+  : MooseObjectAction(parameters)
+{
+}
+
+void
+AddCriticalitySearchAction::act()
+{
+  if (_current_task == "add_criticality_search")
+  {
+    auto openmc_problem = dynamic_cast<OpenMCCellAverageProblem *>(_problem.get());
+
+    if (!openmc_problem)
+      mooseError("The [CriticalitySearch] block can only be used with wrapped OpenMC cases! "
+                 "You need to change the [Problem] block to 'OpenMCCellAverageProblem'.");
+
+    if (_type == "OpenMCMaterialDensity")
+    {
+      _moose_object_pars.set<OpenMCCellAverageProblem *>("_openmc_problem") = openmc_problem;
+      auto search =
+          openmc_problem->addObject<CriticalitySearchBase>(_type, _name, _moose_object_pars, false)[0];
+    }
+  }
+}
+#endif
--- a/src/base/CardinalApp.C
+++ b/src/base/CardinalApp.C
@@ -219,6 +219,12 @@ CardinalApp::associateSyntaxInner(Syntax & syntax, ActionFactory & /* action_fac
   // Can only add external auxvars after the tallies have been added.
   addTaskDependency("add_external_aux_variables", "add_tallies");
 
+  // Add the [Problem/CriticalitySearch] block
+  registerSyntax("AddCriticalitySearchAction", "Problem/CriticalitySearch");
+  registerMooseObjectTask("add_criticality_search", CriticalitySearch, false);
+  registerTask("add_criticality_search", false /* is required */);
+  addTaskDependency("add_criticality_search", "init_mesh");
+
   // Register a modify outputs task to enable variable hiding in the MGXS action.
   registerTask("modify_outputs", true /* is required */);
   addTaskDependency("modify_outputs", "common_output");
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -27,6 +27,9 @@
 #include "OpenMCNuclideDensities.h"
 #include "OpenMCDomainFilterEditor.h"
 #include "OpenMCTallyEditor.h"
+#include "CriticalitySearchBase.h"
+
+//#include "BrentsMethod.h"
 
 #include "openmc/random_lcg.h"
 
@@ -372,9 +375,15 @@ OpenMCProblemBase::externalSolve()
     openmc_set_seed(_initial_seed);
   }
 
-  int err = openmc_run();
-  if (err)
-    mooseError(openmc_err_msg);
+  int err;
+  if (_criticality_search)
+   _criticality_search->searchForCriticality();
+  else
+  {
+    err = openmc_run();
+    if (err)
+      mooseError(openmc_err_msg);
+  }
 
   _total_n_particles += nParticles();
 
@@ -387,6 +396,23 @@ OpenMCProblemBase::externalSolve()
   // save the latest fission source for re-use in the next iteration
   if (_reuse_source)
     writeSourceBank(sourceBankFileName());
+}
+
+void
+OpenMCProblemBase::initialSetup()
+{
+  CardinalProblem::initialSetup();
+
+  // Find a criticality search object
+  TheWarehouse::Query query = theWarehouse().query().condition<AttribSystem>("CriticalitySearch");
+  std::vector<CriticalitySearchBase *> objs;
+  query.queryInto(objs);
+
+  if (objs.size() > 1)
+    mooseError("Cannot have more than one CriticalitySearch object");
+
+  if (objs.size())
+    _criticality_search = objs[0];
 }
 
 void
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -233,14 +233,6 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
 OpenMCProblemBase::~OpenMCProblemBase() { openmc_finalize(); }
 
 void
-OpenMCProblemBase::catchOpenMCError(const int & err, const std::string descriptor) const
-{
-  if (err)
-    mooseError("In attempting to ", descriptor, ", OpenMC reported:\n\n",
-      std::string(openmc_err_msg));
-}
-
-void
 OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
                                             const std::vector<unsigned int> & elem_ids,
                                             const Real & value)
--- a/src/criticality/CriticalitySearchBase.C
+++ b/src/criticality/CriticalitySearchBase.C
@@ -1,0 +1,81 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
+/*                  (c) 2021 UChicago Argonne, LLC                  */
+/*                        ALL RIGHTS RESERVED                       */
+/*                                                                  */
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "CriticalitySearchBase.h"
+#include "BrentsMethod.h"
+#include "openmc/eigenvalue.h"
+
+InputParameters
+CriticalitySearchBase::validParams()
+{
+  auto params = MooseObject::validParams();
+  params.addRequiredParam<Real>("minimum", "Minimum for values to search over");
+  params.addRequiredParam<Real>("maximum", "Maximum for values to search over");
+  params.addRangeCheckedParam<Real>("tolerance", 1e-3, "tolerance > 0", "Absolute tolerance to converge multiplication factor");
+  params.addClassDescription("Base class for defining parameters used in a criticality search in OpenMC.");
+  params.registerBase("CriticalitySearch");
+  params.registerSystemAttributeName("CriticalitySearch");
+  params.addPrivateParam<OpenMCCellAverageProblem *>("_openmc_problem");
+  return params;
+}
+
+CriticalitySearchBase::CriticalitySearchBase(const InputParameters & parameters)
+  : MooseObject(parameters),
+    _maximum(getParam<Real>("maximum")),
+    _minimum(getParam<Real>("minimum")),
+    _tolerance(getParam<Real>("tolerance")),
+    _openmc_problem(*getParam<OpenMCCellAverageProblem *>("_openmc_problem"))
+{
+  if (_minimum >= _maximum)
+    paramError("minimum", "The 'minimum' value must be less than the 'maximum' value");
+
+  auto pp_params = _factory.getValidParams("KEigenvalue");
+  _openmc_problem.addPostprocessor("KEigenvalue", "k", pp_params);
+}
+
+void
+CriticalitySearchBase::searchForCriticality()
+{
+  _console << "Running criticality search in OpenMC for " << quantity() << " in range " << std::to_string(_minimum) << " - " << std::to_string(_maximum) << std::endl;
+
+  std::function <Real(Real)> func;
+  func = [this](Real x) {
+    updateOpenMCModel(x);
+    int err = openmc_run();
+    if (err)
+      mooseError(openmc_err_msg);
+
+    int n = openmc::simulation::n_realizations;
+    const auto & gt = openmc::simulation::global_tallies;
+
+    if (n <= 3)
+      mooseError("Cannot compute combined k-effective estimate with fewer than 4 realizations!\n"
+        "Please change the 'value_type' to either 'collision', 'tracklength', or 'absorption'.");
+
+    double k_eff[2];
+    openmc::openmc_get_keff(k_eff);
+    std::cout << k_eff[0] << std::endl;
+    return k_eff[0] - 1.0;
+  };
+
+  BrentsMethod::root(func, _minimum, _maximum, _tolerance);
+}
+
+#endif
--- a/include/actions/AddCriticalitySearchAction.h
+++ b/include/actions/AddCriticalitySearchAction.h
@@ -1,0 +1,31 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
+/*                  (c) 2021 UChicago Argonne, LLC                  */
+/*                        ALL RIGHTS RESERVED                       */
+/*                                                                  */
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#pragma once
+
+#include "MooseObjectAction.h"
+
+class AddCriticalitySearchAction : public MooseObjectAction
+{
+public:
+  static InputParameters validParams();
+
+  AddCriticalitySearchAction(const InputParameters & parameters);
+
+  virtual void act() override;
+};
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -66,13 +66,6 @@ public:
   std::string subdomainName(const SubdomainID & id) const;
 
   /**
-   * Print a full error message when catching errors from OpenMC
-   * @param[in] err OpenMC error code
-   * @param[in] descriptor descriptive message for error
-   */
-  void catchOpenMCError(const int & err, const std::string descriptor) const;
-
-  /**
    * Whether the score is a reaction rate score
    * @return whether the tally from OpenMC has units of 1/src
    */
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -45,6 +45,7 @@
 class OpenMCNuclideDensities;
 class OpenMCDomainFilterEditor;
 class OpenMCTallyEditor;
+class CriticalitySearchBase;
 
 /**
  * Base class for all MOOSE wrappings of OpenMC
@@ -57,6 +58,8 @@ public:
   static InputParameters validParams();
 
   virtual ~OpenMCProblemBase() override;
+
+  virtual void initialSetup() override;
 
   /**
    * Get the subdomain name for a given ID. If not named, we return the ID
@@ -517,4 +520,8 @@ protected:
 
   /// ID used by OpenMC to indicate that a material fill is VOID
   static constexpr int MATERIAL_VOID{-1};
+
+  CriticalitySearchBase * _criticality_search;
+
+  //std::function<Real(Real)> criticalitySearch;
 };
--- a/include/base/UserErrorChecking.h
+++ b/include/base/UserErrorChecking.h
@@ -75,3 +75,10 @@ void checkRequiredParam(const InputParameters & p,
 void checkJointParams(const InputParameters & p,
                       const std::vector<std::string> & name,
                       const std::string & explanation);
+
+/**
+ * Print a full error message when catching errors from OpenMC
+ * @param[in] err OpenMC error code
+ * @param[in] descriptor descriptive message for error
+ */
+void catchOpenMCError(const int & err, const std::string descriptor);
--- a/include/criticality/OpenMCMaterialDensity.h
+++ b/include/criticality/OpenMCMaterialDensity.h
@@ -1,0 +1,37 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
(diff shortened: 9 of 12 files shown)


include/criticality/CriticalitySearchBase.h
```diff
@@ -0,0 +1,43 @@
...
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#pragma once
+
+#include "MooseObject.h"
+#include "OpenMCCellAverageProblem.h"
+
+class CriticalitySearchBase : public MooseObject
```
nuclearkevin: This should inherit from `OpenMCBase` alongside `MooseObject`. This gives it the `OpenMCProblemBase` error check and the pointer to a `OpenMCCellAverageProblem`.

src/base/CardinalApp.C
```diff
@@ -219,6 +219,12 @@ CardinalApp::associateSyntaxInner(Syntax & syntax, ActionFactory & /* action_fac
   // Can only add external auxvars after the tallies have been added.
   addTaskDependency("add_external_aux_variables", "add_tallies");
 
+  // Add the [Problem/CriticalitySearch] block
+  registerSyntax("AddCriticalitySearchAction", "Problem/CriticalitySearch");
+  registerMooseObjectTask("add_criticality_search", CriticalitySearch, false);
+  registerTask("add_criticality_search", false /* is required */);
+  addTaskDependency("add_criticality_search", "init_mesh");
```
nuclearkevin: Setting a task dependency to `init_mesh` will cause problems, as the `Problem` isn't initialized at this point. On some machines this will result in the criticality search getting added before the problem. I'd recommend setting this to `init_problem` instead.

src/actions/AddCriticalitySearchAction.C
```diff
@@ -0,0 +1,59 @@
...
+{
+  if (_current_task == "add_criticality_search")
+  {
+    auto openmc_problem = dynamic_cast<OpenMCCellAverageProblem *>(_problem.get());
+
+    if (!openmc_problem)
+      mooseError("The [CriticalitySearch] block can only be used with wrapped OpenMC cases! "
+                 "You need to change the [Problem] block to 'OpenMCCellAverageProblem'.");
+
+    if (_type == "OpenMCMaterialDensity")
```
nuclearkevin: Is there any reason as to why this if statement is here? It doesn't appear to be necessary as `_openmc_problem` will be common to all `CriticalitySearchBase` derived objects.

src/actions/AddCriticalitySearchAction.C
```diff
@@ -0,0 +1,59 @@
...
+    {
+      _moose_object_pars.set<OpenMCCellAverageProblem *>("_openmc_problem") = openmc_problem;
```
nuclearkevin: You won't need to set this parameter if `CriticalitySearchBase` derives from `OpenMCBase`.

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,81 @@
...
+  params.addPrivateParam<OpenMCCellAverageProblem *>("_openmc_problem");
+  return params;
+}
+
+CriticalitySearchBase::CriticalitySearchBase(const InputParameters & parameters)
+  : MooseObject(parameters),
+    _maximum(getParam<Real>("maximum")),
+    _minimum(getParam<Real>("minimum")),
+    _tolerance(getParam<Real>("tolerance")),
+    _openmc_problem(*getParam<OpenMCCellAverageProblem *>("_openmc_problem"))
```
nuclearkevin: You get this for free if `CriticalitySearchBase` derives from `OpenMCBase`.

include/criticality/CriticalitySearchBase.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual void searchForCriticality();
+
+protected:
+  virtual std::string quantity() const = 0;
+
+  const Real & _maximum;
+  const Real & _minimum;
+  const Real & _tolerance;
+
+  OpenMCCellAverageProblem & _openmc_problem;
```
nuclearkevin: See above.

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,81 @@
...
+{
+  if (_minimum >= _maximum)
+    paramError("minimum", "The 'minimum' value must be less than the 'maximum' value");
+
+  auto pp_params = _factory.getValidParams("KEigenvalue");
+  _openmc_problem.addPostprocessor("KEigenvalue", "k", pp_params);
```
nuclearkevin: Why do you add a `KEigenvalue` post-processor here? It doesn't look like it gets used anywhere. If the intent is to use this to compute and report k during the search loop, I'd recommend doing some error checking to make sure the user hasn't added a `KEigenvalue` named `k` yet. Adding one when there is a collision results in a confusing error message.

nuclearkevin (changes requested):
Looks good, my requested changes mostly involve implementation details. The design looks easy to extend to other search quantities (boron concentration, rod height, drum rotation, etc.)!

src/criticality/CriticalitySearchBase.C
aprilnovak: yeah I was about halfway through the coding - I wanted to pull k from these postprocessors but ultimately the postprocessor needs to execute on a strange frequency not typical for MOOSE postprocessors. So I opted to do some refactoring to simply access the objects directly from OpenMC as opposed to via a postprocessor that I have to time correctly.

src/actions/AddCriticalitySearchAction.C
```diff
@@ -0,0 +1,55 @@
...
+{
+  if (_current_task == "add_criticality_search")
+  {
+    auto openmc_problem = dynamic_cast<OpenMCCellAverageProblem *>(_problem.get());
+
+    if (!openmc_problem)
+      mooseError("The [CriticalitySearch] block can only be used with wrapped OpenMC cases! "
+                 "You need to change the [Problem] block to 'OpenMCCellAverageProblem'.");
+
+    auto search = openmc_problem->addObject<CriticalitySearchBase>(
```
nuclearkevin: We can discard the return value here as it isn't used.

```suggestion
    openmc_problem->addObject<CriticalitySearchBase>(
```

src/actions/AddFieldTransferAction.C
```diff
@@ -48,13 +48,9 @@ AddFieldTransferAction::act()
...
 
-    if (_type == "NekFieldVariable" || _type == "NekVolumetricSource" ||
-        _type == "NekBoundaryFlux" || _type == "NekMeshDeformation")
-    {
-      _moose_object_pars.set<NekRSProblem *>("_nek_problem") = nek_problem;
-      auto transfer =
-          nek_problem->addObject<FieldTransferBase>(_type, _name, _moose_object_pars, false)[0];
-    }
+    _moose_object_pars.set<NekRSProblem *>("_nek_problem") = nek_problem;
+    auto transfer =
```
nuclearkevin:
```suggestion
```

src/actions/AddScalarTransferAction.C
```diff
@@ -49,12 +49,9 @@ AddScalarTransferAction::act()
...
                  "You need to change the [Problem] block to 'NekRSProblem'.");
 
-    if (_type == "NekScalarValue" || _type == "NekPostprocessorValue")
-    {
-      _moose_object_pars.set<NekRSProblem *>("_nek_problem") = nek_problem;
-      auto transfer =
-          nek_problem->addObject<ScalarTransferBase>(_type, _name, _moose_object_pars, false)[0];
-    }
+    _moose_object_pars.set<NekRSProblem *>("_nek_problem") = nek_problem;
+    auto transfer =
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -895,6 +911,10 @@ void
 OpenMCProblemBase::executeFilterEditors()
 {
   executeControls(EXEC_FILTER_EDITORS);
+
+  if (!_filter_editor_uos.size())
```
nuclearkevin: Out of curiosity, why is this (and the addition between lines 927 and 930) necessary?

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,125 @@
...
+    _tolerance(getParam<Real>("tolerance")),
+    _estimator(getParam<MooseEnum>("estimator").getEnum<eigenvalue::EigenvalueEnum>())
+{
+  if (_minimum >= _maximum)
+    paramError("minimum",
+               "The 'minimum' value (" + std::to_string(_minimum) +
+                   ") must be less than the 'maximum' value (" + std::to_string(_maximum) + ").");
+
+  auto pp_params = _factory.getValidParams("Receiver");
+  _openmc_problem->addPostprocessor("Receiver", _pp_name, pp_params);
```
nuclearkevin: I'd recommend checking to see if a post-processor called `_pp_name` already exists before adding one, and throwing an error message if one does. 

If the user attempts to add one in their input file, this will result in an opaque error message from MOOSE that will be difficult for them to debug. I don't foresee this happening all that often but it's better that we handle this ourselves.

nuclearkevin (changes requested):
Just a few minor comments / changes. The only major change I'd recommend would be to swap over to using https://github.com/OTU-Centre-for-SMRs/nuclear_data as a submodule instead of copy-pasting functions in (to avoid adding 3,000 lines of lookup tables). Otherwise, looks good to me!

src/actions/AddFieldTransferAction.C
```diff
@@ -48,13 +48,9 @@ AddFieldTransferAction::act()
       mooseError("The [FieldTransfers] block can only be used with wrapped Nek cases! "
                  "You need to change the [Problem] block to 'NekRSProblem'.");
 
-    if (_type == "NekFieldVariable" || _type == "NekVolumetricSource" ||
-        _type == "NekBoundaryFlux" || _type == "NekMeshDeformation")
```
gonuke, line 52: I know nothing about Nek transfers, but this seems like an unusual change for this PR

src/base/OpenMCProblemBase.C
aprilnovak: they're not relevant to this PR, I was just getting annoyed by the following lines printing out things about editing filters, etc. when there weren't any of these objects. If you like I can move to a separate PR, I just felt it was harmless to sneak this in :)

src/actions/AddScalarTransferAction.C
```diff
@@ -49,12 +49,9 @@ AddScalarTransferAction::act()
       mooseError("The [ScalarTransfers] block can only be used with wrapped Nek cases! "
                  "You need to change the [Problem] block to 'NekRSProblem'.");
 
-    if (_type == "NekScalarValue" || _type == "NekPostprocessorValue")
```
gonuke, line 52: Same comment as above

src/base/OpenMCProblemBase.C
nuclearkevin: It's a small change, so no harm in keeping it in here :)

gonuke (changes requested):
Thanks for the effort on this @aprilnovak - I definitely would not have produced a design quite this sophisticated.

I've taken a Quick Look at the code (i.e. not yet the docs or tests) and it all looks good to me.

I did find two small changes related to Nek that seemed out of place and I wanted to flag to see that it's what you intended.

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,3892 @@
...
+               "The 'minimum' boron ppm (" + std::to_string(_minimum) + ") must be positive!");
+  if (_maximum > 50000)
+    paramError("maximum",
+               "The borated water composition is computed using a dilute species approximation. "
+               "Results will not be accurate if the boron species is no longer dilute, which we "
+               "take as 5\% weight concentration or higher. Please decrease 'maximum' (" +
+                   std::to_string(_maximum) + ") to an upper limit which is in the dilute regime.");
+
+  // We take the provided material, retaining its density, but then overwriting
+  // any nuclides there to be H2O + boron weight ppm and add the S(a,b) tables
```
aprilnovak: note to self: forgot to add S(a,b) tables

src/actions/AddFieldTransferAction.C
aprilnovak: From one of Kevin's earlier comments, I realized that this if statement was actually not necessary to have in Cardinal (I'm adding a new action in this PR for a different purpose). I took the liberty to simplify this code at the same time, even though it is irrelevant to the topic of criticality searches. If you prefer I can certainly move this to a separate PR - I was just lazy :)

aprilnovak:
> The only major change I'd recommend would be to swap over to using https://github.com/OTU-Centre-for-SMRs/nuclear_data as a submodule instead of copy-pasting functions in (to avoid adding 3,000 lines of lookup tables).

Yes, agreed! Thanks for moving that shared code into a submodule @nuclearkevin. I've updated this PR to add the new submodule.

src/actions/AddFieldTransferAction.C
gonuke: No need to remove... I just wanted to make sure it wasn't creeping in accidentally and in some kind of incomplete way

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,223 @@
...
+          idx++;
+        }
+
+        idx = 0;
+        for (const auto & o : _oxygen_natural)
+        {
+          if (o.first == "O18")
+            idx18 = idx;
+          idx++;
+        }
```
gonuke: Can't these be combined?
```suggestion
        for (const auto & o : _oxygen_natural)
        {
          if (o.first == "O16")
            idx16 = idx;
          if (o.first == "O18")
            idx18 = idx;
          idx++;
        }
```

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,223 @@
...
+  std::vector<std::string> allowable;
+  for (const auto & i : _hydrogen_natural)
+    allowable.push_back(i.first);
+  for (const auto & i : _boron_natural)
+    allowable.push_back(i.first);
+  for (const auto & i : _oxygen_natural)
+    allowable.push_back(i.first);
+
+  if (isParamValid("absent_nuclides"))
+  {
```
gonuke: It seems useful to me for this block to be in its own private function.  It can help the overall readability by reducing the cognitive burden of this function.

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,223 @@
...
+    densities.push_back(b.second * frac_B * rho * openmc::N_AVOGADRO);
+  }
+
+  openmc::model::materials[_material_index]->set_densities(names, densities /* atom/b-cm */);
+
+  const int * nucl;
+  const double * dens;
+  int n;
+  err = openmc_material_get_densities(_material_index, &nucl, &dens, &n);
+  catchOpenMCError(err, "get nuclide densities from material " + std::to_string(_material_id));
```
gonuke: Is this just checking that we haven't left the material in some broken state?

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,223 @@
...
+          idx++;
+        }
+
+        _oxygen_natural[idx16].second += _oxygen_natural[idx18].second;
+        _oxygen_natural.erase(_oxygen_natural.begin() + idx18);
+      }
+      else
+        paramError(
+            "absent_nuclides",
+            "Cardinal currently only assumes that O18 may be missing from your cross section "
```
gonuke: All documentation and code comments suggest that H and B are handled also, but this message contradicts this.  Should the docs be consistent with this?

src/criticality/BoratedWater.C
aprilnovak: all of OpenMC's C-API functions have error code return values, so I always capture those just in case something gets thrown. I don't think it's possible for something to be thrown here because we fetch the `_material_index` from another C-API call, but I'd prefer to leave the error catch in case this code does change some way in the future where that's no longer true.

gonuke: I guess my question was larger than that... there is no algorithmic reason to call `openmc_material_get_densities()` here.  We have already set them, and don't do anything with these results.  I therefore assumed that we called this to just test that we set things properly?

aprilnovak: Ah, I understand now - I agree, this is not necessary here (I think I had this as part of debugging while writing the implementation and forgot to remove it later). Thanks for catching this!

src/criticality/CriticalitySearchBase.C
aprilnovak: I'd like to propose that such a change belongs in MOOSE, not necessarily Cardinal. I can open a PR there? fwiw, this is the error message that MOOSE throws already:

```
A UserObject with the name "critical_value" already exists.  You may not add a Postprocessor by the same name.
```

IMO it's relatively clear that the way to fix the problem is to rename the postprocessor the user is adding (although it's certainly not clear on what else is adding a conflicting-named postprocessor). Perhaps this error message from MOOSE can just be revised to something like

```
A UserObject/Postprocessor with the name "critical_value" already exists (possibly added by an action).  You may not add a Postprocessor by the same name.
```

src/criticality/BoratedWater.C
aprilnovak: I see both perspectives on this -- my philosophy has been for the description of input parameters to reflect the ideal, end-code state and to handle any gaps using error messages. For example, there is a lot of this which happens using enumerations for types of data transfers, only some of which are supported by class A, class B, etc. Rather than define separate enumerations for every class having to do with data transfer, I simply re-use the same enumeration and have error messages. 

Otherwise, I think a better name for this parameter would be `O18_missing` (bool), but I prefer to add the parameter `absent_nuclides` to allow easy extension in the future while supporting backwards compatibility.

But it's not a big deal to modify this - let me know if you prefer that.

src/criticality/CriticalitySearchBase.C
nuclearkevin: An improved error message on the MOOSE side works for me.

doc/content/source/criticalitysearch/BoratedWater.md
```diff
@@ -0,0 +1,30 @@
...
+The density of the material remains unchanged (but it may still be being set by a
+coupled thermal-fluid application).
+
+!alert note
+This object does not search in your cross section library to see what nuclides
+are present (some libraries do not include O18, for instance). OpenMC will
+throw an error if your library is missing a nuclide that `BoratedWater` is
+trying to add. To circumvent this, please explicitly note which of the natural
+nuclides from hydrogen (H1, H2), boron (B10, B11), and oxygen (O16, O17, O18)
+are missing from your material with `absent_nuclides`.
```
gonuke: Apropos my earlier comment about the implementation being consistent with the documentation... 

This documentation could be rephrased to focus on the cases that are most likely.  I'd add that given their importance in reactor physics, it's unlikely for either of the H or B nuclides to be missing - there would be a lot of other problems!

aprilnovak: works for me - I updated the documentation. let me know what you think

nuclearkevin (approved):
LGTM! Should be good to go once the CIVET recipe is updated to take into account the new submodule addition and the tests start running + passing.

aprilnovak:
yep, waiting for the PR in civet_recipes to be merged.

src/base/OpenMCBase.C
```diff
@@ -46,4 +47,82 @@ OpenMCBase::stdev(const double & mean, const double & sum_sq, unsigned int reali
...
+    case eigenvalue::combined:
+    {
+      if (n <= 3)
+        mooseError("Cannot compute combined k-effective standard deviation with fewer than 4 "
+                   "realizations!");
+
+      double k_eff[2];
+      openmc::openmc_get_keff(k_eff);
+      return k_eff[1];
+    }
```
lewisgross1296: I believe this is a conversion between batch statistics to population statistics for reporting the k standard deviation. Is there a reason we don't do the same thing for the `combined` as for the others? If so, maybe a comment could be added for posterity

src/criticality/BoratedWater.C
lewisgross1296: Is this something you intend to do in a future PR? (just double checking since I saw this comment)

src/criticality/BoratedWater.C
```diff
@@ -0,0 +1,217 @@
...
+    {
+      // only missing nuclides for H, O, B are meaningful
+      if (std::find(allowable.begin(), allowable.end(), a) == allowable.end())
+        paramWarning("absent_nuclides",
+                     "Only absent isotopes of hydrogen, oxygen, or boron will be used to adjust "
+                     "natural abundances. The entry '" +
+                         a + "' will be unused.");
+
+      // adjust the natural abundances if nuclides are missing from the cross section library
+      // TODO: implement in a general fashion, this only works for O18 which is also the only
```
lewisgross1296: Is this TODO left for a future PR? (confirming, if so maybe we could make an issue to keep track?)

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,126 @@
...
+    _k_std_dev_values.push_back(k_std_dev);
+
+    if (_tolerance < 3 * k_std_dev)
+      mooseDoOnce(mooseWarning(
+          "The 'tolerance' for the criticality search (" + std::to_string(_tolerance) +
+          ") is smaller than 3-sigma standard deviation in k (" + std::to_string(3 * k_std_dev) +
+          "); you may have to run a lot of criticality search points to converge to this "
+          "tolerance. You may want to loosen 'tolerance' or increase the number of particles."));
+
+    return k - 1.0;
```
lewisgross1296: Can you remind me why we return `k-1.0` here?

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,126 @@
...
+    Real k_std_dev = kStandardDeviation(_estimator);
+    _k_values.push_back(k);
+    _k_std_dev_values.push_back(k_std_dev);
+
+    if (_tolerance < 3 * k_std_dev)
+      mooseDoOnce(mooseWarning(
+          "The 'tolerance' for the criticality search (" + std::to_string(_tolerance) +
+          ") is smaller than 3-sigma standard deviation in k (" + std::to_string(3 * k_std_dev) +
+          "); you may have to run a lot of criticality search points to converge to this "
+          "tolerance. You may want to loosen 'tolerance' or increase the number of particles."));
```
lewisgross1296: What if the reason this is true is because of poor statistics? Might be worth considering a comment that they can lower their `k_std_dev` by using more particles

src/criticality/CriticalitySearchBase.C
```diff
@@ -0,0 +1,126 @@
...
+
+    return k - 1.0;
+  };
+
+  BrentsMethod::root(func, _minimum, _maximum, _tolerance);
```
lewisgross1296: Do we plan to support all the different OpenMC methods for search (different scipy root finders)? I think there's a refactor coming soon to allow the user to specify the exact arguments for a search that scipy uses (see [discussion](https://github.com/openmc-dev/openmc/pull/2693)), so once that happens perhaps we can PR here to allow the same specification of arguments in the `CriticalitySearch` block

lewisgross1296 (commented):
Nothing major, just somethings I didn't fully understand and was hoping for clarification / addition of comments to the code.

Thanks for putting this together so rapidly @aprilnovak!

src/criticality/CriticalitySearchBase.C
aprilnovak: We have to call root finding methods from C++ - we can certainly support other options, but they will not be related to any scipy python-based implementation.

src/criticality/CriticalitySearchBase.C
aprilnovak: I'm not sure I understand what is being suggested - the warning message tells the user they could "increase the number of particles" as one possible fix

src/criticality/CriticalitySearchBase.C
aprilnovak: our target k is 1.0, so this number represents a residual

src/criticality/BoratedWater.C
aprilnovak: It could be implemented in a future PR if needed, but I think it would be unlikely. My preference however is to leave a TODO statement anywhere that I know there is missing functionality

src/criticality/CriticalitySearchBase.C
lewisgross1296: My bad I missed the particles comment in the error message :)

src/criticality/CriticalitySearchBase.C
nuclearkevin: We're also constrained by the rootfinding methods that are available in MOOSE / the ones we want to roll ourselves. Going through all of the options in SciPy to add every root finding algorithm to Cardinal / MOOSE would be quite tedious, and not really worth our time unless there's a good reason.

lewisgross1296: Yeah that makes sense. The only non-bracketed method I'd like to formally request is the secant method, since Dean Price studied it in a Journal paper examining drum searches. I've done some secant method testing in my `search_for_keff` work so far and it works pretty well. I can take an attempt at adding this in a future PR

gonuke: Do we know that Brent is much worse than secant? I suggest that we only do secant if/when we know we need it

aprilnovak: Something else to consider -  at the moment, we run an entire OpenMC calculation for each trial. We could save a lot of time if we ran the initial calculations with fewer particles and ramped up particles during the calculations. This (or some variant like running the criticality search during inactive batches) could be a future PR.

lewisgross1296: > Do we know that Brent is much worse than secant? I suggest that we only do secant if/when we know we need it

I can do some testing to confirm, but my current understanding is the Bracketed methods may require more eigenvalue solves to get the same answer. Maybe @pshriwise has some input.

I'll link this [paper](https://doi.org/10.1016/j.pnucene.2023.104731) because it's very solid for understanding eigenvalue searches using Monte Carlo (with drums as the focus). But from the abstract:
> In the end, a single secant-like method is recommended for implementation due to superior performance on these worth curves.

lewisgross1296: > Something else to consider - at the moment, we run an entire OpenMC calculation for each trial. We could save a lot of time if we ran the initial calculations with fewer particles and ramped up particles during the calculations. This (or some variant like running the criticality search during inactive batches) could be a future PR.

I'm not totally sure about the savings of using less particles in the earlier evaluations. The paper I linked above examined an algorithm for changing the number of particles simulated and incorporating uncertainties into predicting the next guess.

From my (recent) experience with the secant solves -- neutronics only, I think there's an advantage to having very well converged first and second guesses, as it will have a more accurate slope to predict the next evaluation. With two good initial evalutaions, it's possible to get a very good third guess, and so on, which might lead to less evaluations in total. 

This is certainly worth studying because the way to get to a critical configuration with the least number of eigenvalue solves is important/interesting

pshriwise: > Something else to consider - at the moment, we run an entire OpenMC calculation for each trial. We could save a lot of time if we ran the initial calculations with fewer particles and ramped up particles during the calculations. This (or some variant like running the criticality search during inactive batches) could be a future PR.

> Maybe @pshriwise has some input.

I don't have a lot of experience in this area, but I think there could be something to this! My feeling is that there is use in truncating evaluations, but perhaps only in certain situations and the approach may be significantly impacted by the root finding method. If the current parameter value (guess) is far from the root value then, as @lewisgross1296 said, having a smaller uncertainty is useful for a better guess of the derivative in our current methods. If we allow ourselves to consider changing the root finding approach to take into account uncertainty when updating the parameter value (i.e. making a conservative change to the parameter), we might be able to find an advantage. 

Brain storming here, so bear w/ me:

It would be entirely feasible to do something like run OpenMC batch-by-batch until we know k-eff is outside the tolerance to 3-sigma, virtually guaranteeing that this evaluation won't be within tolerance. The parameter could then be updated conservatively using a 1-sigma interval so as not to overshoot the root value (tbd what the details of that would look like). As we get closer to the root and the residual becomes smaller, the uncertainty in our estimation will get progressively smaller inherently to satisfy this check until we either complete the simulation batches as-specified or satisfy the condition that the final value gets the residual within N-sigma of zero, completing the search. Ours could be something of a unique situation as a lot of the literature I've seen seems to assume that uncertainty on the function evaluation is assumed ahead of time and can't be measured on-the-fly (to a degree) like we are able to in OpenMC. Using the tolerance to influence the number of batches/particles would also remove the need for users to be concerned about setting a tolerance that's far smaller than the simulation std. dev. based on the OpenMC settings.

aprilnovak: Yeah, I like this idea! I think it belongs as a separate PR, since there's some investigation to be done.

gonuke: I think most of the topics discussed here are fodder for separate PRs.  This method should be robust and any of the alternatives are easy to add as extensions in the future - and many require some research to be sure what the right/best/reasonable answer is.

gonuke:
> Passing thought: might be nice to allow a target k-eff to be specified (with a default of 1.0).

This might be a straightforward addition to include here, esp if it's optional with a default value of 1.0.

aprilnovak:
I'm going to leave adding the S(a,b) tables to a future PR since there's not a nice API to do it on the OpenMC side. Updated the docs to reflect this.

src/criticality/BoratedWater.C
aprilnovak: addressed in issue #1194

src/base/OpenMCBase.C
aprilnovak: I'm not sure I follow - these APIs at least give the same values for k that are printed to the console at the end of an OpenMC run (this "new" code is just being moved from `KEigenvalue` to a new place).

test/tests/criticality/borated_water/gold/openmc_out.csv
```diff
@@ -0,0 +1,3 @@
+time,critical_value,k
+0,0,0
+1,522.19791328636,1.0002525559801
```
nuclearkevin: I believe something changed in how the critical boron search is performed between when you generated the gold file for the test and now (which is why the test is failing). It might need a regold.

aprilnovak: it's passing fine on my local machine - I think it's some indeterminisim as the number of batches increases (locally, I'm seeing the same values for k print out up until a point, so I tried decreasing the number of batches in my latest push). Will keep trying!

gonuke:
> Passing thought: might be nice to allow a target k-eff to be specified (with a default of 1.0).

We'd be happy to PR this little tweak to your branch.

We're getting close to having the rotational search ready.  Should we PR that to this branch? or wait for this to be merged first?

aprilnovak:
That would be great if you wanted to add, thank you! Sorry this has slipped a bit on my to-do list (still need to figure out why the test is passing for me locally but not on CIVET). Please feel free to go ahead and add a commit with the k \ne 1 option!

I'd suggest we merge this one first and then a follow-on one with the cell modification options.

gonuke:
> That would be great if you wanted to add, thank you! Sorry this has slipped a bit on my to-do list (still need to figure out why the test is passing for me locally but not on CIVET). Please feel free to go ahead and add a commit with the k \ne 1 option!
> 
> I'd suggest we merge this one first and then a follow-on one with the cell modification options.

I PR'd this into your branch for you to review and merge there.

lewisgross1296:
Is it possible to use `openmc.Settings.seed` in the test? Also are any of the tests using OpenMP parallelism? In my experience to get identical searches, you need to use only MPI parallelism

aprilnovak:
I don't think the seed is relevant (the tests are using the same exact seed). Also, I believe OpenMC should be reproducible even if using OpenMP (but FWIW, the tests on CIVET use MPI only I believe, unless otherwise instructed, which we are not in these new tests being added).

PR #1196, opened 2025-08-23
Block restrict tally variables

### LINKED ISSUE #1195: Block restrict tally variables in OpenMCCellAverageProblem
## Reason
At present, tally variables are added for all blocks in an OpenMC simulation. We support block restriction in `CellTally` & `MeshTally`, so we can enable block restriction for the auxvariables themselves to reduce simulation memory requirements.

## Design
`TallyBase` should be refactored such that it includes the blocks tallies act on and the associated user error handling (to reduce code duplication). This will also allow `OpenMCCellAverageProblem` to fetch the list of blocks and use them when it adds the tally variables.

## Impact
Reduction in memory usage when block restriction is employed for tally variables.

nuclearkevin:
Closes #1195


test/tests/neutronics/fixed_source/overlap_solid.i
```diff
@@ -59,11 +59,7 @@
   [heat_source]
     type = ElementIntegralVariablePostprocessor
     variable = kappa_fission
-  []
-  [heat_source_fluid]
-    type = ElementIntegralVariablePostprocessor
-    variable = kappa_fission
-    block = '200'
+    block = '100'
```
aprilnovak: I'm confused what this is doing - now the same block is being provided to 2 different postprocessors?

nuclearkevin: This was done to minimize the amount of regolding required (the post-processors reported the same results even though one had a different set of blocks). I can get rid of the post-processor entirely and remove it's entry in the gold file if it makes it less confusing. 

The overlap tests don't make a lot of sense now that tallies are properly block restricted.

aprilnovak: I'd just update the test to make sense - otherwise the way the test is written now is confusing and doesn't relate to the test requirements. Especially since this one is just CSV, it's no memory penalty to re-gold.

tutorials/pincell_multiphysics/openmc.i
```diff
@@ -162,6 +162,8 @@ dT = ${fparse power / mdot / Cp}
     source_variable = heat_source
     variable = power
     to_multi_app = bison
+    from_blocks = '2'
```
aprilnovak, line 165: For the tutorials, would you please also update the XML files? We do version control them in case someone doesn't have the Python API installed

aprilnovak: Is the `gas_assembly` tutorial missing here as well?

aprilnovak: sorry, ignore these comments. I thought I was on the density PR

src/tallies/TallyBase.C
```diff
@@ -44,6 +44,13 @@ TallyBase::validParams()
       "Auxiliary variable name(s) to use for OpenMC tallies. "
       "If not specified, defaults to the names of the scores");
 
+  params.addParam<std::vector<SubdomainName>>(
+      "block",
+      "Subdomains for which to add tallies in OpenMC. If not provided, "
+      "tallies will be applied over the entire mesh.");
```
aprilnovak:
```suggestion
      "tallies will be applied over the entire domain corresponding to the [Mesh] block.");
```

aprilnovak: I feel like this is slightly more specific and might be helpful for people using cell tallies (I could see "entire mesh" leading to some confusion with what role the mesh plays for cell tallies). Feel free to disregard, since the phrasing was already there before

test/tests/neutronics/feedback/lattice/openmc.i
```diff
@@ -62,16 +62,7 @@
   [heat_source]
     type = ElementIntegralVariablePostprocessor
     variable = heat_source
-  []
```
aprilnovak, line 65: couldn't a fix to avoid a re-gold be to have the tallies exist on blocks 1, 2, and 3? Sometimes our tests become slightly more verbose/not recommended approach in the interest of avoiding regolds.

aprilnovak: And I would say in this case if we _can_ avoid a re-gold of exodus, to do so.

test/tests/neutronics/heat_source/gold/multi_tally_mapping_solid_out.csv
```diff
@@ -1,3 +1,3 @@
-time,fluid_heating,fluid_kappa_fission,heating,kappa_fission,solid_heating,solid_kappa_fission
-0,0,0,0,0,0,0
-1,0,0,100,100,100,100
+time,solid_heating,solid_kappa_fission
```
aprilnovak, line 1: For these `....overlap` tests, I would prefer if we didn't have to re-gold the CSV files. The tests are checking for correct behavior of zero tallies in some locations but not others. It's highly unlikely this would happen, but I'm envisioning a scenario where the code regresses and we do start putting nonzero tallies in the fluid blocks and I want to be sure we retain that functionality. In other words, I'd just restrict this PR to be 90% the source code changes, and then whatever changes are needed to the `block` in the `.i` files (and try avoiding all regolds unless we have to).

aprilnovak: Sorry for the extended review on this one, I could have been clearer with the comments earlier

nuclearkevin: The issue is that with actual block restriction of the tally variables they aren't defined over the overlapping parts of the domain at all (those components don't exist in the DoF array), and so the post-processors that are testing to ensure that the "old style" of block restriction works (where cell tallies store a 0 in unmapped regions) throw an error. If I add the overlapping blocks to the cell tally's block restriction (to stop the post-processors from throwing an error), the tests also fail due to some heating being reported in the fluid regions.

nuclearkevin: "True" block restriction can only really be tested with an Exodiff as block restricted variables get a (partial) vs unrestricted variables. An alternative would be to use a RunException test to catch a post-processor block restriction error. 

As it stands these tests won't pass with the changes in this PR unless we do something horrible to the test input files, like changing the integral post-processors with constant post-processors purely to get the tests passing (which is not best practice in the slightest).

aprilnovak: > The issue is that with actual block restriction of the tally variables they aren't defined over the overlapping parts of the domain at all

instead of "aren't defined over the overlapping" did you mean "aren't defined over the non-overlapping"?

Also, what is the error which gets thrown?

nuclearkevin: The error that gets thrown is a variant of the following:
```
The 'block' parameter of the object 'heat_source' must be a subset of the 'block' parameter of the variable 'heat_source':
    Object 'heat_source': 100, 200
    Variable 'heat_source': 100
```
This is caused by the use of true block restriction for the tallies. Regions that would previously be filled with zeros are no longer defined in memory, and so post-processors cannot access variables in those blocks (in this case, block 200).

> instead of "aren't defined over the overlapping" did you mean "aren't defined over the non-overlapping"?

Ah yes, my bad. When thinking about these tests I conceptualized the "overlap" to mean the fluid blocks that overlap with the solid CSG.

aprilnovak: > This is caused by the use of true block restriction for the tallies.

What I'm proposing is the following (to avoid a re-gold). Let's suppose we'll only have nonzero tallies over block 200. The original test still created the auxvariable over blocks 100 and 200. Why don't we just still leave the tally defined over blocks 100 AND 200? It would retain the original intention of these tests. We can use some other test to actually utilize the new block restriction.

aprilnovak: If I am still not understanding something, let's talk in person once I'm back.

nuclearkevin: I believe it's best if we discuss this PR in person and go over some examples.

aprilnovak:
@nuclearkevin were you still planning on refactoring the `heat_source` folder tests to a more modern coverage? Just checking if that was still planned or if this PR is waiting on me.

nuclearkevin:
Yep, I'm still planning on refactoring those tests. I just haven't gotten around to it yet.

PR #1198, opened 2025-08-25
Implement distributed cell densities

### LINKED ISSUE #1197: Distributed cell density support
## Reason
Currently, if users wish to take density feedback into account they need to create a new OpenMC material for each distributed cell instance to receive the updated cell average densities. This is an extremely laborious process for most reactor models as one cannot use lattices (as OpenMC lattice objects do not support unique materials per lattice element). To remedy this, a per-cell density or per-cell density multiplier should be added to allow density to be distributed in a similar manner to temperature. Cardinal could then use this distributed density when applying density feedback on a per-cell basis.

## Design
This would almost exclusively entail a change in OpenMC to support distributed densities. See https://github.com/openmc-dev/openmc/issues/2900 and https://github.com/openmc-dev/openmc/pull/3546 for more information. Cardinal would need to set the cell density / cell density multiplier instead of a material density. 

## Impact
Improved model building experience for density feedback and some simplification of density feedback code. An added bonus of this change would be that extra materials won't be required any more for density skinning in the `MoabSkinner` - the model rebuilding code in `OpenMCCellAverageProblem` could be simplified quite a bit.

nuclearkevin:
This PR implements distributed cell densities (using a per-cell density multiplier) based on [this draft PR in OpenMC](https://github.com/openmc-dev/openmc/pull/3546). This shouldn't be merged until the API on the OpenMC side is finalized (as it currently points to a branch in an OpenMC fork), I just wanted to put this up there to get some feedback on the design.

Note:  a few tests have been removed as they test for limitations of density feedback; one test required a regold as the `MoabSkinner` no longer needs new materials (so the reported material IDs on the mesh mirror were different); and one new test was added for distributed cell densities.

Closes #1197


aprilnovak:
I'd be in favor of re-golding for the tutorials, since most people I imagine will generate the XML files themselves as prompted by the tutorial (and it could be confusing to see the diffs). None of the tutorials test any requirements which aren't already covered in the test suite (i.e. the only purpose of the tests on the tutorials is to make sure they don't break).

tutorials/gas_assembly/assembly.py
```diff
@@ -353,7 +352,7 @@ def assembly(n_ax_zones, n_inactive, n_active, add_entropy_mesh=False):
     l = bundle_pitch / math.sqrt(3.0)
     lower_left = (-l, -l, reactor_bottom)
     upper_right = (l, l, reactor_top)
-    source_dist = openmc.stats.Box(lower_left, upper_right, only_fissionable=True)
+    source_dist = openmc.stats.Box(lower_left, upper_right)
```
aprilnovak, line 355:
```suggestion
    source_dist = openmc.stats.Box(lower_left, upper_right, constraints={'fissionable': True})
```

doc/content/tutorials/pincell_multiphysics.md
```diff
@@ -93,9 +93,9 @@ First, we create a number of materials to represent
 the fuel pellet and cladding. Because these materials will only receive temperature updates (i.e.
 we will not change their density, because doing so would require us to move cell boundaries in
 order to preserve mass), we can simply create one material that we will use on all axial layers.
-Next, we create a number of materials to represent the sodium material in each
-axial layer of the model. Because the sodium cells will change both their temperature *and*
-density, we need to create a unique material for each axial layer.
+Similar to temperature, a unique value of density can also be specified for each cell instance,
```
aprilnovak, line 96: `pincell_multiphysics.py` should be updated to remove the loop creating N sodium materials

aprilnovak: whoops, sorry I see it later.

doc/content/tutorials/pincell_multiphysics.md
```diff
@@ -93,9 +93,9 @@ First, we create a number of materials to represent
...
+and so we create a single sodium material for all of the coolant cells that will receive density
+and temperature feedback from NekRS.
```
aprilnovak, line 99: The `gas_compact_multiphysics` tutorial should also be updated -  the `unit_cell.py` creates cloned materials and there might be something in the markdown as well.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1495,14 +1477,6 @@ OpenMCCellAverageProblem::getMaterialFills()
     // didnt set up the materials correctly (if mapping by cell)
```
aprilnovak: Do we need `materials_in_fluid` at all anymore? We use it elsewhere in your PR still to see if a cell's filling material is present in uncoupled regions - so as to throw an error if you were to change material density in a coupled region but that material is also present somewhere else in the model. But now I'm wondering if the density is tied to a cell, if we need this at all anymore?

tutorials/pincell_multiphysics/gold/openmc_out.csv
```diff
@@ -1,3 +1,3 @@
 time,k,k_std_dev,max_tally_rel_err
 0,0,0,0
-0.5,0.2481419481052,0.0028332223175455,0.047053473144978
+0.5,0.24484871719829,0.0030298690436078,0.050045006530896
```
aprilnovak: What is the explanation for these changes?

tutorials/pincell_multiphysics/pincell.py
```diff
@@ -82,7 +74,7 @@
 # Create an initial uniform spatial source distribution over fissionable zones
 lower_left = (-pitch, -pitch, 0.0)
 upper_right = (pitch, pitch, height)
-uniform_dist = openmc.stats.Box(lower_left, upper_right, only_fissionable = True)
+uniform_dist = openmc.stats.Box(lower_left, upper_right)
 
 settings.source = openmc.IndependentSource(space=uniform_dist)
```
aprilnovak:
```suggestion
settings.source = openmc.IndependentSource(space=uniform_dist, constraints={'fissionable' : True})
```

aprilnovak: if you accept this and the other change related to constraints, please re-update the XML files too

aprilnovak:
```
  for (const auto & f : materials_in_fluid)
    if (other_materials.count(f))
      mooseError(
          printMaterial(f) +
          " is present in more than one OpenMC cell with different "
          "density feedback settings!\nIn other words, this material will have its density changed "
          "by Cardinal (because it is\ncontained in cells which map to the 'density_blocks'), but "
          "this material is also present in\nOTHER OpenMC cells, which will give unintended "          "behavior "
          "by changing density in ALL parts of the\ndomain containing this material (some of which "
          "have not been coupled via Cardinal).\n\n"
          "Please change your OpenMC model so that unique materials are used in regions which "
          "receive "
          "density feedback.");
```

I think this check can also be deleted - because density is now tied to a cell, we'd never be in the case where we're touching densities in uncoupled cells on accident.

doc/content/tutorials/openmc_fluid.md
```diff
@@ -160,13 +160,8 @@ In each layer we set up the lattice
 structure by listing the universes in each "ring" of the lattice, with `ring0` being
 the centermost ring and `ring11` being the outermost ring.
```
aprilnovak, line 162: Please also check the `msfr` tutorial, there might be some wording in the markdown file related to creating unique materials by the skinner.

Also, does any documentation in the `MoabSkinner` need to be changed?

tutorials/gas_assembly/assembly.py
nuclearkevin: I would prefer to not add the fissionable constraint here. Restricting the starter source to fissionable material in TRISO problems runs afoul of the "95% of source sites have been rejected" error in OpenMC.

tutorials/pincell_multiphysics/gold/openmc_out.csv
nuclearkevin: Changes to the OpenMC Python API (notably natural abundances) between the last time the xml files were generated and this PR resulted in slightly different material compositions. This, combined with the low number of particle histories used in the test resulted in failures, and so a regold was necessary. I mentioned this in this PR 2 weeks ago, where you were OK with me regolding the tests.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: My OpenMC addition didn't fully add density as a cell property, it added a "density multiplier" which is used to scale the material density to a specified cell density. This requires that cells be filled with materials in order to set the density multiplier (`mult = cell_density / mat_density`), which I error check on the Cardinal side. There's an error on the OpenMC side for this, but we can add more information if we handle it ourselves.

doc/content/tutorials/openmc_fluid.md
nuclearkevin: Good catch, I missed the MSFR tutorial. Surprisingly the `MoabSkinner` documentation doesn't mention adding new materials for each density bin.

tutorials/gas_assembly/assembly.py
aprilnovak: this case initially was restricted, but it was removed - I just wanted to understand why that was done here since it seems unrelated. I'm ok to remove it though, because I agree

tutorials/pincell_multiphysics/gold/openmc_out.csv
nuclearkevin: To elaborate: the changes to natural abundances are small but modify the RNG stream. This is what yields the difference in test results. As does the removal of the fissionable constraint, which likely has a larger impact now that I think about it...

I re-added the fissionable constraint and the test is still failing with the change in isotopic abundances for uranium nuclides. If we **really** want to avoid a regold, I can ditch the change to the fuel composition in the `materials.xml` file and the test will pass. This will result in some differences between the checked-in xml files and the xml files generated with the OpenMC Python API though, so I'm not for this personally.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: Ah wait, you're right; we don't use it at all. I was thinking about a previous version of this PR. I'm on board with removing it.

~~I believe `_cell_to_material` can also be removed.~~ On second thought, it helps with debugging models so it's probably good to keep in.

tutorials/pincell_multiphysics/gold/openmc_out.csv
aprilnovak: I'm ok with the re-gold, I just wanted to confirm it wasn't solely due to the change in the fissionable source sampling.

src/base/OpenMCCellAverageProblem.C
aprilnovak: Ok sounds good - then let's remove `materials_in_fluid` (I agree to keep `_cell_to_material` since it's used to print some helpful tables for debugging). In place of the error check related to `materials_in_fluid`, let's add a test where the same material is in 2 cells, but only 1 cell is having its density changed. We can check that the solution is identical to a standalone OpenMC run with different densities for 2 different (but otherwise identical) materials. Then I think we'll have full coverage of the new behavior.

aprilnovak:
`gas_assembly/geometry.xml` snuck in there, that file is too large for commit

nuclearkevin:
Ak, thanks for catching that!

aprilnovak:
would you please rebase to address the conflict?

nuclearkevin:
@aprilnovak should be good to merge now.

PR #1204, opened 2025-09-02
Refactor build configuration

ahnaf-tahmid-chowdhury:
Update the build configuration to include additional directories and streamline the inclusion of dependencies. Remove outdated checks and ensure proper environment variable usage for NEKRS_HOME. Introduce new build options for related components like MOAB, DAGMC, and OpenMC. This enhances the overall build process and dependency management.


config/dagmc.mk
```diff
@@ -1,3 +1,11 @@
+DAGMC_BUILDDIR := $(CARDINAL_DIR)/build/DAGMC
+DAGMC_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
+
+# By default, build dagmc
+BUILD_DAGMC ?= yes
```
nuclearkevin:
```suggestion
BUILD_DAGMC ?= no
```
Default behaviour in Cardinal is not build with DAGMC unless the user specifies that they want CAD geometry support (due to the additional build burden).

config/double_down.mk
```diff
@@ -1,3 +1,11 @@
+DOUBLEDOWN_BUILDDIR := $(CARDINAL_DIR)/build/double-down
+DOUBLEDOWN_INSTALL_DIR := $(CONTRIB_INSTALL_DIR)
+
+# By default, build doubledown
+BUILD_DOUBLEDOWN ?= yes
```
nuclearkevin, line 5: Double-Down support in Cardinal is still experimental - so the default behaviour is to avoid building with it. This is due to some issues related to double-precision ray tracing in the CSG-to-CAD tutorial - it leaks all particles when running with Double-Down (which does not happen with MOAB).

Additionally, we also had issues building Double-Down on certain systems / platforms, but this PR might've fixed those issues. I'll check this out and try building in the problem environments to see if it works.

nuclearkevin (commented):
Just a few notes about changed default behaviour. April will need to add you to the neams-th-coe organization for tests to run.

config/dagmc.mk
ahnaf-tahmid-chowdhury: I think that behavior is handled by `ENABLE_DAGMC`. The `BUILD_DAGMC` option lets users decide whether to build DAGMC through Cardinal. If it is disabled, Cardinal may use a preinstalled version of DAGMC.

nuclearkevin: Ahh, sorry. Misread the flags. Feel free to disregard!

ahnaf-tahmid-chowdhury:
Not sure how to start the workflows.

aprilnovak:
I am not sure either - I just relaunched the test to see if it is repeatable.

ahnaf-tahmid-chowdhury:
> @ahnaf-tahmid-chowdhury are the recent commits trying to get around the test failure? If so, I can take a closer look to see if I can figure it out

Yes, but I have undone this. Still not sure why the test is failing with the same version of NekRS. Maybe tweaking the iteration number will fix it. I have merged the upstream branch; let’s see what the difference is.

nuclearkevin:
Apologies for the wait - I've been slammed over the last few weeks and I've only just started clawing through my backlog. I'm testing this on a few different machines at Argonne to ensure our build instructions still function on their systems, and building with these changes on my personal laptop. Will drop a review once that's done!

config/check_deps.mk
```diff
@@ -71,53 +71,77 @@ ifneq (,$(findstring +,$(moose_status)))
...
-  ifeq ($(OPENMC_CONTENT),)
-    $(error $n"OpenMC does not seem to be available, but ENABLE_OPENMC is enabled. Make sure that the submodule is checked out.$n$nTo fetch the OpenMC submodule, use ./scripts/get-dependencies.sh")
-  else
-    $(info Cardinal is using OpenMC from          $(OPENMC_DIR))
-  endif
+  ifeq ($(BUILD_OPENMC),yes)
+    ifeq ($(OPENMC_CONTENT),)
+      $(error $n"OpenMC does not seem to be available, but ENABLE_OPENMC is enabled. Make sure that the submodule is checked out.$n$nTo fetch the OpenMC submodule, use ./scripts/get-dependencies.sh")
+    else
+      $(info Cardinal is using OpenMC from          $(OPENMC_DIR))
```
nuclearkevin, line 95: When building with the OpenMC/NekRS/DAGMC/Embree/Double-Down submodules in Cardinal, I've noticed that I don't get this message letting me know that I'm using Cardinal's submodules. Instead, I get the pre-built messages:
```bash
Cardinal is using HDF5 from            /home/kevin/Projects/cardinal/contrib/moose/petsc/arch-moose
Cardinal is using MOOSE from           /home/kevin/Projects/cardinal/contrib/moose
Cardinal is using pre-built OpenMC from /home/kevin/Projects/cardinal/contrib/openmc
Cardinal is using pre-built DAGMC from /home/kevin/Projects/cardinal/contrib/DAGMC
Cardinal is using pre-built Moab from /home/kevin/Projects/cardinal/contrib/moab
```
This is fairly minor as the build proceeds normally, but it would be nice if these messages were consistent with the HDF5/MOOSE messages for the case where only Cardinal submodules are being used.

ahnaf-tahmid-chowdhury: Could you please take another look? This only happens when these are already pre-built.

nuclearkevin: Yep, this works now! I might've forgot to delete the `build` and `install` directories prior to testing your branch, sorry about that.

aprilnovak:
@ahnaf-tahmid-chowdhury actually, would you please add something to the `without_conda.md` and `with_conda.md` pages to document the `BUILD_...` options?

doc/content/with_conda.md
```diff
@@ -39,13 +42,42 @@ If the above produces a `cardinal-opt` Cardinal executable, you can
 jump straight to [#running]. If you want to customize the
 build or were not successful with the above, please consult the detailed instructions
 that follow.
+
+!alert! tip title=Using OpenMC from your Conda Environment
+You can install OpenMC directly into your conda environment. See the "Using OpenMC from your Conda Environment" section for more details.
```
aprilnovak: @ahnaf-tahmid-chowdhury are you using the MOOSE conda environment with OpenMC installed into it?

nuclearkevin (approved):
Everything works as expected on INL HPC (WindRiver/Bitterroot). The 67 Pebble CHT case runs with no issues, and the pincell multiphysics problem also runs fine. Tested with 2 nodes (224 MPI ranks).

doc/content/with_conda.md
ahnaf-tahmid-chowdhury: After looking into it more deeply, I realized that the conda version of the cardinal build isn’t actually being used yet. It only pulls in hdf5 from conda, and Cardinal seems to build all the moose packages by itself anyway. So there’s really no need for a separate moose environment with MOOSE already installed. We can still use openmc from the conda environment, as long as we choose a version that’s compatible with cardinal. In short, I haven’t found any real benefit to having a moose environment, since the current build script doesn’t use any moose deps from it.

aprilnovak: I guess my request would be to clarify the difference between conda environments in the Cardinal documentation. The "with/without" conda is referring to MOOSE's conda environment, and then building OpenMC from source as part of that. But I think you've written the docs also assuming someone is using OpenMC's conda. I'd suggest we leave it as just MOOSE's conda environemnt, then perhaps have a hyperlink to another page which clarifies that you can _also_ use OpenMC's conda environment on top of that?

The MOOSE conda environment allows us to skip building libMesh, PETSc, and wasp. These take significant amount of time.

doc/content/with_conda.md
```diff
@@ -21,16 +21,19 @@ OpenMC separately, or using its conda environment (i.e., do
 
 !alert! note title=tldr
 
-All that you need to compile Cardinal is:
+Once your `cardinal` conda environment is set up, all you need to do to compile Cardinal is:
 
 ```
-conda activate moose
+conda activate cardinal
+conda install -c conda-forge openmc # Install OpenMC from conda-forge
```
aprilnovak: Can you please update this documentation so that it does not rely on the OpenMC conda environment? Let's stick with the original MOOSE conda environment since it includes libMesh & wasp & petsc and simultaneously has all the OpenMC dependencies already

doc/content/with_conda.md
```diff
@@ -21,16 +21,19 @@ OpenMC separately, or using its conda environment (i.e., do
...
 cd $HOME
 git clone https://github.com/neams-th-coe/cardinal.git
 cd cardinal
 ./scripts/get-dependencies.sh
 export ENABLE_NEK=false
 export HDF5_ROOT=$CONDA_PREFIX
+export OPENMC_DIR=$CONDA_PREFIX
```
aprilnovak: suggest to move this to a "tip" box, if anyone wants to use OpenMC conda and MOOSE conda simultaneously

`!alert tip`

doc/content/with_conda.md
```diff
@@ -39,13 +42,42 @@ If the above produces a `cardinal-opt` Cardinal executable, you can
...
+You can install OpenMC directly into your conda environment. See the "Using OpenMC from your Conda Environment" section for more details.
+!alert-end!
 !alert-end!
 
 ## Access
 
 !include access.md
 
-To get MOOSE's conda environment, follow the instructions [here](https://mooseframework.inl.gov/getting_started/installation/index.html).
+To get started, you'll need to create a dedicated conda environment for Cardinal. This ensures that all the necessary dependencies are isolated and managed correctly.
```
aprilnovak: I'm not a conda expert, but I don't think this is truly necessary? I'd prefer to avoid any customized instructions beyond just what MOOSE requires (but please educate me if there is an advantage to doing this).

PR #1205, opened 2025-09-02
Adding element hierarchy aux

### LINKED ISSUE #1202: Create an ElementHierarchyAux kernel
## Reason

To compare the solution between different AMR strategies we can a third mesh where the discretization of the mesh will be the union of final meshes created different AMR strategies. To do that we need to track the element hierarchy in each of those meshes. For which we can create an AuxKernel that will store the element hierarchy (  `element->level()` ) in an aux variable. 

## Design
It will inherit form the `AuxKernel` class and just override the  `computeValue()` method. The `computeValue()` will return the `_current_elem->level()`

## Impact

Will store the element hierarchy as an aux variable which then can be read by a combination of `SolutionUserObject` and `SolutionAux` to create a union mesh

magnoxemo:
closes #1202


nuclearkevin:
Just a thought, but wouldn't this make more sense to add to the MOOSE framework instead of Cardinal? It's generic enough that it can be used for more than just Monte Carlo AMR. This was my thought process for adding `ElementMaxLevelPostProcessor` to MOOSE instead of Cardinal.

PR #1208, opened 2025-09-08
Delayed neutron filters and scores

### LINKED ISSUE #1207: Add scores and filters necessary to compute multigroup delayed properties
## Reason
Presently, Cardinal's mesh-based multi-group cross sections are limited to the group constants necessary for static reactor configuration. To support coupling Cardinal with deterministic codes for on-the-fly group constant generation / updates in transients, we should add the filters and scores necessary for delayed multi-group cross sections.

## Design
Add the scores `prompt-nu-fission`, `delayed-nu-fission`, and `decay-rate`. These are required to calculate $\beta_{i,g}$, $\chi_{d,i,g}$, $\chi_{p,g}$, and $\lambda_{i}$. We also need to add a `DelayedGroupFilter` to Cardinal's filter system to filter by delayed group.

## Impact
Enables the tallying of delayed neutron properties and the discrimination of prompt neutrons from delayed neutrons. This also enables a follow-up feature adding delayed multi-group properties to the MGXS action.

nuclearkevin:
Closes #1207


include/filters/FromXMLFilter.h
```diff
@@ -20,6 +20,9 @@
 
 #include "FilterBase.h"
 
+/// A class which provides a thin wrapper around an arbitrary OpenMC
```
aprilnovak: Does this get rendered appropriately in Doxygen? I always through a multiline comment had to be `/** ... **/` for Doxygen

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2550,13 +2550,19 @@ OpenMCCellAverageProblem::tallyMultiplier(unsigned int global_score) const
     else
       source *= *_source_strength;
 
-    // - Reaction rate scores & 'inverse-velocity' have units of reactions/src (OpenMC) or
-    //   reactions/s (Cardinal).
+    // - Reaction rate scores have units of reactions/src (OpenMC) or reactions/s (Cardinal).
+    // - 'inverse-velocity' has units of particles*s/src (OpenMC) or particles (Cardinal).
```
aprilnovak, line 2554: Does this division happen automatically in the source code, or should it be added to the documentation so it's clear for users?

test/tests/neutronics/filters/delayed/cell.i
```diff
@@ -0,0 +1,138 @@
...
+    positions = '0 0 0
+                 0 0 4
+                 0 0 8'
+  []
+  [solid_ids]
+    type = SubdomainIDGenerator
+    input = solid
+    subdomain_id = '100'
+  []
+  [fluid]
```
aprilnovak, line 18: you don't need to do this, but the mesh generators can be simplified a lot - no need to have the fluid at all, and we could have just 1 pebble. Feel free to ignore

nuclearkevin: Ah, thanks for mentioning this! I've gotten rather lazy and have been copying other test inputs with modifications to test the feature being added. In the future I'll make sure to simplify the input files.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: It doesn't happen automatically as we'd have to add an additional `flux` or `delayed -nu-fission` score (incurring a performance penalty). I'll add a note to the documentation so users know that they need to add those scores to obtain the true inverse velocity or decay constant.

include/filters/FromXMLFilter.h
nuclearkevin: You are totally right - I'll get these fixed.

doc/content/source/tallies/scores_triggers.md
```diff
@@ -21,7 +21,15 @@ a tally. Options include:
 - `nu_scatter`: the scattering reaction rate combined with the total (n, xn) reaction rate
 - `fission`: the fission reaction rate
 - `nu_fission`: the fission neutron production rate
-- `inverse_velocity`: one over the particle velocity
+- `prompt_nu_fission`: the prompt fission neutron production rate
```
aprilnovak, line 24: What do you think about adding the units to each of these in the documentation? Could be helpful for users now that we're adding more of these. Perhaps a table that compares the units from OpenMC vs. Cardinal? (I don't recall if we have that already)

nuclearkevin: We have something like this in the documentation for the [OpenMCCellAverageProblem](https://cardinal.cels.anl.gov/source/problems/OpenMCCellAverageProblem.html). Looks like I forgot to add the units for the new scores to that table.

It's probably worth migrating this table to the tallies as it's a little out of place at present. If you're good with this, I'll shuffle it over.

aprilnovak: works for me, thanks!

doc/content/source/filters/DelayedGroupFilter.md
```diff
@@ -0,0 +1,22 @@
+# DelayedGroupFilter
+
+!alert note
```
aprilnovak: Do we need this alert here? We catch this anyways through errors in the source code. I wonder if we can make the docuemntation a little simpler by removing this throughout.

nuclearkevin: I seem to remember discussing this documentation note in a different PR (trying to find a reference to that discussion but I'm coming up empty-handed at the moment). If I'm remembering correctly, the conclusion was that these warnings were fine to keep as this is a custom system added by Cardinal. If you're of the opinion that we should remove these now, I can go through all of the tally + filter documentation and remove the notes. Just wanted to make sure that's the new norm before doing so.

nuclearkevin: Found it: https://github.com/neams-th-coe/cardinal/pull/1029#discussion_r1920245196

aprilnovak: I'm kind of leaning towards removing them (I guess I was seeing more scenarios where we have overly verbose documentation when the source code implementation prevents the error from happening anyways in Walid's amalgamantion PR). So I thought it might be nice to be consistent in the documentation style guide.

src/base/OpenMCCellAverageProblem.C
aprilnovak, line 2554: Should the units be `particles/cm3` in Cardinal (not `particles`), so that when dividing by flux (which has units of `particles/cm2/s` in Cardinal) returns `s/cm`? As far as I can tell (and I know this code has changed a bit over the last year), all tallies are divided by volume except for the relative errors.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2550,13 +2550,19 @@ OpenMCCellAverageProblem::tallyMultiplier(unsigned int global_score) const
...
+    //   This score is flux-weighted, and must be divided by the flux to recover the true
+    //   inverse velocity, which has units of s/cm.
+    // - 'decay-rate' has units of reactions/src/s (OpenMC) or reactions/s^2 (Cardinal).
```
aprilnovak, line 2557: same general question, should the units in Cardinal be reactions/s^2/vol?

nuclearkevin: After the division by volume, yes. However, this function operates on the scores prior to the volume division and so the units are `reactions/s^2`.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: After the division by volume, yes. However, this function operates on the scores prior to the volume division and so the units are `particles`.

aprilnovak: Ah gotcha -thanks!

nuclearkevin:
Unless you have any other comments / concerns @aprilnovak, I think this is good to go!

PR #1210, opened 2025-09-15
Added xml_directory parameter to OpenMCProblemBase

### LINKED ISSUE #1209: Add input parameter to allow user to set OpenMC xml files directory
## Reason
Presently, Cardinal assumed the xml settings files used by OpenMC are within the present working directory. This causes issues if for example we wish to run a MOOSE multi-app system containing two Cardinal apps, where each Cardinal app needs to use separate xml files. 

## Design
Add an input parameter to `OpenMCBaseProblem` names `xml_directory` to allow users to specify the directory of the xml files they wish for OpenMC to use. Insert this directory into argv for OpenMC's initialiser. 

## Impact
Users can now specify a directory for their xml settings files, which as stated could allow for two Cardinal apps to be used in the same MOOSE multi-app system.

TheBEllis:
Added xml_directory parameter to OpenMCProblemBase to let user specify directory of OpenMC xml settings files. 
closes #1209

### DIFF
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -92,7 +92,8 @@ OpenMCProblemBase::validParams()
       "ifp_generations",
       openmc::DEFAULT_IFP_N_GENERATION,
       "The number of generations to use with the method of iterated fission probabilities.");
-
+  params.addParam<std::string>(
+      "xml_directory", "./", "The directory in which to look for OpenMC XML files.");
   return params;
 }
 
@@ -116,11 +117,23 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
     mooseError("The tally system used by OpenMCProblemBase derived classes has been deprecated. "
                "Please add tallies with the [Tallies] block instead.");
 
-  int argc = 1;
-  char openmc[] = "openmc";
-  char * argv[1] = {openmc};
+  std::vector<std::string> argv_vec = {"openmc"};
 
-  openmc_init(argc, argv, &_communicator.get());
+  if (isParamSetByUser("xml_directory"))
+  {
+    argv_vec.push_back(getParam<std::string>("xml_directory"));
+  }
+
+  std::vector<char *> argv;
+
+  for (const auto & arg : argv_vec)
+  {
+    argv.push_back(const_cast<char *>(arg.data()));
+  }
+  // Add terminating nullptr
+  argv.push_back(nullptr);
+
+  openmc_init(argv.size() - 1, argv.data(), &_communicator.get());
 
   // ensure that any mapped cells have their distribcell indices generated in OpenMC
   if (!openmc::settings::material_cell_offsets)


src/base/OpenMCProblemBase.C
```diff
@@ -116,11 +117,23 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
     mooseError("The tally system used by OpenMCProblemBase derived classes has been deprecated. "
                "Please add tallies with the [Tallies] block instead.");
 
-  int argc = 1;
-  char openmc[] = "openmc";
-  char * argv[1] = {openmc};
+  std::vector<std::string> argv_vec = {"openmc"};
+
+  if (isParamSetByUser("xml_directory"))
```
aprilnovak: MOOSE convention is to fetch this parameter in the member initialization list instead - and, since you are setting a default, you actually don't need to check for the `isParamSetByUser` since presumably the parameter will always have a valid value unless the user decides to set it themselves.

TheBEllis: Thanks for looking at this so quickly @aprilnovak, I will get started on getting a test implemented! And I will clean up the initialisation/ parameter retrieval

TheBEllis: A test has been added, it's in the same directory as the eigenvalue tests, I didn't know if it should reside in it's own dir or if it can live there? I've also added a class member `_xml_directory` to `OpenMCProblemBase` that gets initialised in the init list.

aprilnovak: Thanks @TheBEllis! This looks great. Now that I think of it, I think it would be better for the test case to be entirely its own directory (since there's technically no way right now to detect if the case is actually using the XML files in the sub-folder if they are identical to the XML files in the working directory).

Would you please move the test case into its own folder, like under `neutronics`?

Also, I sent you an invite to join `cardinal-devs` - you'll need to accept this invite from github for the test suite to launch here.

src/base/OpenMCProblemBase.C
```diff
@@ -92,7 +92,8 @@ OpenMCProblemBase::validParams()
       "ifp_generations",
       openmc::DEFAULT_IFP_N_GENERATION,
       "The number of generations to use with the method of iterated fission probabilities.");
-
+  params.addParam<std::string>(
```
loganharbour:
```suggestion
  params.addParam<FileName>(
```

This is a special parameter type (underneath the hood it's just a string) that will resolve the parameter based on where the input file is. That is - if you specify `./` it'll resolve to the directory the input file that sets it is. Additionally, if using the `!include` syntax for input files, it'll then resolve to the directory (by default as you have it) where the parameter lives in the input file.

TheBEllis: Thanks Logan, I didn't know that existed and that's really useful! I have committed the change

src/base/OpenMCProblemBase.C
TheBEllis: I've accepted the `cardinal-devs` invite and it looks like the tests are running, I think I've reorganised the tests like you suggested, but let me know if I've put it in the wrong place! 

I did just want to note though that this doesn't completely enable multiple Cardinal OpenMC apps to run in a multiapp structure. I am running into different issues now, where the first multiapp initialised does so successfully, but the second fails to read in reading `settings.xml`  with `ERROR: Need to specify number of particles.`. 

The number of particles is definitely defined, and if I remove either of the apps from my `MultiApp` block, they independently initialise just fine. I'm not completely clued up on how MultiApps handle MPI communicators, whether they get their own or they borrow from the main app. I have a hunch that two separate openmc instances running on the same communicator may be causing the issue, but I have no concrete info to back that up:)

Either way, hopefully this very very small addition is still helpful:)

loganharbour (approved):
FileName change looks good to me! Test does too.

include/base/OpenMCProblemBase.h
```diff
@@ -522,6 +522,9 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
   /// The global tally used to accumulate the scores required for kinetics parameters.
   openmc::Tally * _ifp_tally = nullptr;
 
+  /// Directory in which OpenMC settings xml files are located
+  const std::string _xml_directory;
```
aprilnovak:
```suggestion
  const std::string & _xml_directory;
```

aprilnovak:
Thanks @TheBEllis , this looks great! Just one small request from me to capture the filename as a const reference in sticking with MOOSE convention.

include/base/OpenMCProblemBase.h
TheBEllis: Just committed this:)

PR #1217, opened 2025-09-21
Update OpenMC

nuclearkevin:
This PR updates OpenMC and performs a wholesale regolding of OpenMC tests. This is necessary as the RNG stream recently changed to accommodate temporal DNP sampling and combing for fission site sampling: https://github.com/openmc-dev/openmc/pull/2992


aprilnovak:
Can you update https://github.com/neams-th-coe/cardinal/pull/1196 first? I think we are going to end up deleting some tests in this folder anyways, and I'd prefer to avoid re-adding binary files if we're just going to delete them anyways.

Also, I want to ask if there's any opportunity for tests you've added over the last year whether they can be adequately converted to CSV type tests instead? In the past,  the first time I had to re-gold the entire OpenMC suite I modified a bunch of tests from Exodus to CSV since we will almost surely need to do this again down the road. Certainly not all tests can be made into CSV-type tests, but the more we can have the better. Sorry for not being more stringent about new tests being CSV type, so I thought I'd just propose it here.

aprilnovak:
> I don't know how many of the ExoDiff tests can be easily converted to CSVDiff as the vast majority of the Exodus tests I added test bin to element mapping (to ensure we don't scramble anything during adaptivity or the use of filters). There are a few others that test markers/indicators added over the last year which could be modified, but they're very small compared to the other Exodiff tests.

Ok, that's fine by me - just wanted to check because we'll certainly have to do this in the future! Would you please re-base?

nuclearkevin:
@aprilnovak there was a recent [bugfix for IFP kinetics parameters in OpenMC](https://github.com/openmc-dev/openmc/pull/3580). This PR also pulls that in and regolds the $\beta_{eff}$ `CSVdiff` test.

Also, multi-group kinetic parameters have been merged into OpenMC as well (as of today)! I'll look to getting them implemented end of October when my schedule isn't as crazy.

PR #1221, opened 2025-10-01
Use std::string instead of FileName for the OpenMC XML directory

nuclearkevin:
Fixes relative path issues with symlinks / input file merging / virtual file paths. 
This PR also adds some output to let users know where OpenMC is looking for XML files.

Ref. #1209


aprilnovak:
@TheBEllis for context, the FileName does not work with symlinks, so we're actually reverting back to a std::string. Didn't realize this at the time it was merged!

nuclearkevin:
We should hold on merging this until Logan gets back to us on some of the errors we've been running into with `FileName`.

TheBEllis (approved):
Looks good to me:) Sorry for the delay in reviewing, got caught up in moving house

PR #1233, opened 2025-10-27
Criticality search

aprilnovak:
See #1184


gonuke:
> I opted to rewrite the tests so that they pass as long as (k - target) < tolerance. This allows some floating point non-determinism, while still adequately covering the new feature (in my opinion).

I think this is probably adequate but less than ideal. I'm concerned about false positive where the final eigenvalue is within the tolerance for some reason other than a successful search.  I'm happy to accept this for now and make an issue to consider if there are better ways to construct this.

It might remove some of the pathologies that could create a false positive by reducing the tolerance, but for this particular problem, we know that it's first guesses are far away from the target, so it might not have any impact here.

I also think it might be valuable to include an OpenMC script that performs the criticality search, but I'm not sure what your typical approach is.  I only note that there is an OpenMC script that generates the model, and for other test problems, it's possible to run the OpenMC model directly as well as via Cardinal and compare results.  In some cases, the addition of multiphysics feedbacks makes that irrelevant, but in other cases, the OpenMC and Cardinal may be expected to match.  Right now, there is a small amount of work needed to perform criticality search using the OpenMC test problem if one wanted to compare manually.  That may be outside the scope of what you expect here, though.

gonuke:
I'm happy with it as is, too!  Thanks for all the effort @aprilnovak

PR #1239, opened 2025-11-12
Allow controlling number of particles with a postprocessor

### LINKED ISSUE #1237: Allow number of particles to be provided by a postprocessor
## Reason
In time-dependent simulations, it'd be nice to allow `particles` to be provided by a postprocessor similar to `power`, where it can be controlled as a function, such that a large number of particles is only used during "the important part" without significantly impacting runtime by keeping the rest of the simulation running with a lower number of particles.

## Design
Allow `particles` to take a postprocessor name similar to `power`.

## Impact
Allows for flexibility where a high number of particles can be used only where needed, without impacting the overall runtime unnecessarily.

meltawila:
closes #1237

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -2268,7 +2268,15 @@ OpenMCCellAverageProblem::externalSolve()
   else
   {
     if (isParamValid("particles"))
-      openmc::settings::n_particles = OpenMCProblemBase::nParticles();
+    {
+      if (*_particles <= 0.0)
+        mooseError(
+            "'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+            "your postprocessor and check that the postprocessor value itself is not less than "
+            "or equal to zero.");
+      int64_t n = std::llround(*_particles);
+      openmc::settings::n_particles = n;
+    }
   }
 
   OpenMCProblemBase::externalSolve();
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -279,12 +279,7 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
 int
 OpenMCProblemBase::nParticles() const
 {
-  if (*_particles <= 0.0)
-    mooseError("'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
-               "your postprocessor and check that the postprocessor value itself is not less than "
-               "or equal to zero.");
-
-  return *_particles;
+  return openmc::settings::n_particles;
 }
 
 std::string
@@ -366,8 +361,7 @@ OpenMCProblemBase::externalSolve()
     return;
   }
 
-  _console << " Running OpenMC with " << openmc::settings::n_particles << " particles per batch..."
-           << std::endl;
+  _console << " Running OpenMC with " << nParticles() << " particles per batch..." << std::endl;
 
   // apply a new starting fission source
   if (_reuse_source && !firstSolve())
@@ -396,7 +390,7 @@ OpenMCProblemBase::externalSolve()
       mooseError(openmc_err_msg);
   }
 
-  _total_n_particles += openmc::settings::n_particles;
+  _total_n_particles += nParticles();
 
   err = openmc_reset_timers();
   if (err)


src/base/OpenMCProblemBase.C
```diff
@@ -274,7 +273,20 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
 int
 OpenMCProblemBase::nParticles() const
 {
-  return openmc::settings::n_particles;
+  if (!isParamValid("particles"))
+    return static_cast<int>(openmc::settings::n_particles);
```
aprilnovak: do you need to have this static cast?

src/base/OpenMCProblemBase.C
```diff
@@ -274,7 +273,20 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
+
+  const double r = std::round(*_particles);
+
+  if (r <= 0.0)
+  {
+    mooseWarning(
```
aprilnovak: IMO I think a mooseError would be more appropriate here - a negative number of particles is definitely wrong

src/base/OpenMCProblemBase.C
```diff
@@ -274,7 +273,20 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
+        "'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in your "
+        "postprocessor. number of particles from settings.xml will be used instead.");
+    return openmc::settings::n_particles;
+  }
+
+  return static_cast<int>(r);
```
aprilnovak: I think you can combine the round line from 279 here - you don't really need to round before you do your check above.

src/base/OpenMCProblemBase.C
```diff
@@ -355,7 +367,7 @@ OpenMCProblemBase::externalSolve()
     _console << " Skipping running OpenMC as the mesh has not changed!" << std::endl;
     return;
   }
-
+  openmc::settings::n_particles = nParticles();
```
aprilnovak: Do you need this line? The code already had support for changing the number of particles each batch (for Dufek-Gudowski relaxation). Please check if the infrastructure can be combined somehow

test/tests/neutronics/heat_source/particles_from_postprocessor.i
```diff
@@ -0,0 +1,90 @@
...
+  []
+[]
+
+[Postprocessors]
+  [p]
+    type = Receiver
+    default = 100.0
+  []
+  [particles]
+    type = ConstantPostprocessor
```
aprilnovak: For this test, I think it would be better if the particles postprocessor changed each picard iteration (that way, we'd also check that we not only change the first setting, but in later Picard iterations too)

test/tests/neutronics/heat_source/particles_from_postprocessor.i
```diff
@@ -0,0 +1,90 @@
...
+  power = p
+  particles = particles
+  temperature_blocks = '100 200'
+  density_blocks = '200'
+  verbose = true
+  cell_level = 0
+
+  initial_properties = xml
+
+  [Tallies]
```
aprilnovak: Please make the test as simple as possible - you can delete all tallies, you can turn off multiphysics feedback, etc. Will make it easier to maintain long-term.

test/tests/neutronics/heat_source/tests
```diff
@@ -35,6 +35,21 @@
     requirement = "The power shall be provided by a postprocessor"
     capabilities = 'openmc'
   []
+  [n_particles_from_postprocessor]
+    type = Exodiff
+    input = particles_from_postprocessor.i
```
aprilnovak, line 48: let's use a CSVDiff type test - we don't care about any of the multiphysics stuff, we just want to see that particles are changing, which we can do via the `OpenMCParticles` postprocessor. Otherwise, if e.g. OpenMC changes their random number stream (which they have some several times in the past), we have to re-gold any of our tests which have to do with tallies.

meltawila: I changed the test to CSVDiff

test/tests/neutronics/heat_source/particles_from_postprocessor.i
```diff
@@ -0,0 +1,73 @@
...
+  []
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  power = 100
+  particles = n_particles
+
+  verbose = true
+  cell_level = 0
```
aprilnovak:
```suggestion
```

test/tests/neutronics/heat_source/particles_from_postprocessor.i
```diff
@@ -0,0 +1,73 @@
+[Mesh]
+  [sphere]
+    type = FileMeshGenerator
+    file = ../meshes/sphere.e
+  []
+  [solid]
```
aprilnovak: we can simplify this mesh stuff - let's delete all the mesh generators from `solid` onwards (will help with maintaining this long term, and its irrelevant for the test)

src/base/OpenMCProblemBase.C
```diff
@@ -276,7 +273,16 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
+  if (!isParamValid("particles"))
+    return openmc::settings::n_particles;
+
+  if (*_particles <= 0.0)
+  {
+    mooseError("'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+               "your postprocessor.");
+  }
+
+  return std::round(*_particles);
```
aprilnovak:
```suggestion
  return *_particles;
```

aprilnovak: this should do a cast automatically, since the return type is already `int`. Slightly different effect, but slightly simpler code

src/base/OpenMCProblemBase.C
```diff
@@ -276,7 +273,16 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
 OpenMCProblemBase::nParticles() const
 {
-  return openmc::settings::n_particles;
+  if (!isParamValid("particles"))
+    return openmc::settings::n_particles;
+
+  if (*_particles <= 0.0)
+  {
+    mooseError("'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+               "your postprocessor.");
```
aprilnovak:
```suggestion
               "your postprocessor and check that the postprocessor value itself is not less than zero.");
```

aprilnovak: please also add a test checking that this error is trigger correctly when particles <= 0

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2302,6 +2299,13 @@ OpenMCCellAverageProblem::externalSolve()
   // doesn't intrude with any other postprocessing routines that happen outside this class's purview
   if (_relaxation == relaxation::dufek_gudowski && !firstSolve())
     dufekGudowskiParticleUpdate();
+  else
+  {
+    if (isParamValid("particles"))
+      openmc::settings::n_particles = OpenMCProblemBase::nParticles();
+  }
+  _console << " Running OpenMC with " << openmc::settings::n_particles << " particles per batch..."
```
aprilnovak: Is there a reason this was moved from the base class to here? Since `OpenMCProblemBase::externalSolve();` is called just after here, I'd leave the console print where it originally was

test/tests/neutronics/heat_source/tests
```diff
@@ -35,6 +35,17 @@
...
   []
+  [n_particles_from_postprocessor]
+    type = CSVDiff
+    input = particles_from_postprocessor.i
+    csvdiff = 'particles_from_postprocessor_out.csv'
+
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
+    # in the fission bank on a process
+    max_parallel = 8
+    requirement = "The number of particles shall be provided by a postprocessor"
```
aprilnovak:
```suggestion
    requirement = "The system shall control the number of particles in OpenMC by a postprocessor"
```

test/tests/neutronics/heat_source/tests
```diff
@@ -35,6 +35,17 @@
     requirement = "The power shall be provided by a postprocessor"
     capabilities = 'openmc'
   []
+  [n_particles_from_postprocessor]
+    type = CSVDiff
+    input = particles_from_postprocessor.i
+    csvdiff = 'particles_from_postprocessor_out.csv'
+
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
```
aprilnovak: Is this comment and restriction here for parallel settings still applicable given the new choices of particles being used?

test/tests/neutronics/heat_source/particles_from_postprocessor.i
```diff
@@ -0,0 +1,73 @@
...
+    type = ParsedFunction
+    value = 100*t
+  []
+[]
+
+[Postprocessors]
+  [n_particles]
+    type = FunctionValuePostprocessor
+    function = nparticles_fn
+    execute_on = 'initial timestep_begin'
```
aprilnovak: Do you need the `initial` here? The simpler the better in terms of a new user copying this as a starting point

include/base/OpenMCProblemBase.h
aprilnovak, line 89: why was this added here? I don't see any corresponding changes to .C files which call `addExternalVariable`

aprilnovak, line 239: also, why was this added? I don't see this signature being called anywhere in the new code added?

meltawila: same as above, both are from commit ab4fc4f to resolve the merge conflict

meltawila: This was to resolve a merge conflict after another PR got merged, not part of my code

aprilnovak: ah, ok then I think this will go away if you just rebase your branch off the latest `devel` and then force push to here.

src/base/OpenMCProblemBase.C
```diff
@@ -280,7 +279,17 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
 int
 OpenMCProblemBase::nParticles() const
 {
-  return openmc::settings::n_particles;
+  if (!isParamValid("particles"))
```
aprilnovak: does this check need to be here? elsewhere, you only call this function `if isParamValid("particles")`, so I think your code is slightly redundant. It'd be simpler to have just one of these checks

meltawila: it's necessary for when Dufek Gudowski relaxation is used and _total_n_particles gets updated with += nParticles(). I actually added this check because it was failing for this reason

meltawila: I think can change `_total_n_particles += nParticles();` to `_total_n_particles += openmc::settings::n_particles;` and remove this check if you prefer that

src/base/OpenMCCellAverageProblem.C
```diff
@@ -305,6 +305,7 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
     checkUnusedParam(params, "particles", "using Dufek-Gudowski relaxation");
     checkRequiredParam(params, "first_iteration_particles", "using Dufek-Gudowski relaxation");
     openmc::settings::n_particles = getParam<int>("first_iteration_particles");
+    _n_particles_1 = getParam<int>("first_iteration_particles");
```
aprilnovak, line 308: Now, `_n_particles_1` is only being filled if using Dufek-Gudowski relaxation (before, it was always being populated). This is fine, but let's change the documentation in `OpenMCCellAverageProblem.h` to be clear that `_n_particles_1` doesn't always contain the first particles

src/base/OpenMCProblemBase.C
aprilnovak: I like that suggestion, that's clearer IMO (and your suggestion would also be consistent, style-wise, with the console print you use in `externalSolve` which prints the number of particles using `openmc::settings::n_particles`. It's best to write code using a consistent accessor/getter across the board.

src/base/OpenMCProblemBase.C
```diff
@@ -280,7 +279,17 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
+    return openmc::settings::n_particles;
+
+  if (*_particles <= 0.0)
+  {
```
aprilnovak: It looks like the clang formatter doesn't care, but the convention we use in Cardinal is to avoid using the braces if there's only a single line inside a conditional

aprilnovak:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -280,7 +279,17 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
+    mooseError(
+        "'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+        "your postprocessor and check that the postprocessor value itself is not less than zero.");
+  }
```
aprilnovak:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -280,7 +279,17 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
...
 {
-  return openmc::settings::n_particles;
+  if (!isParamValid("particles"))
+    return openmc::settings::n_particles;
+
+  if (*_particles <= 0.0)
+  {
+    mooseError(
+        "'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+        "your postprocessor and check that the postprocessor value itself is not less than zero.");
```
aprilnovak:
```suggestion
        "your postprocessor and check that the postprocessor value itself is not less than or equal to zero.");
```

aprilnovak: sorry, missed this specific error the first review.

test/tests/neutronics/heat_source/tests
```diff
@@ -35,6 +35,21 @@
     requirement = "The power shall be provided by a postprocessor"
     capabilities = 'openmc'
   []
+  [zero_particles]
+    type = RunException
+    input = zero_particles_from_postprocessor.i
+    expect_err = "'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
+               "your postprocessor and check that the postprocessor value itself is not less than zero."
```
aprilnovak:
```suggestion
               "your postprocessor and check that the postprocessor value itself is not less than or equal to zero."
```

test/tests/neutronics/heat_source/tests
```diff
@@ -35,6 +35,21 @@
...
+    requirement = "The system shall be capable of block restricting CellTally variables."
```
aprilnovak:
```suggestion
    requirement = "The system shall error if the user specifies an invalid number of particles."
```

aprilnovak:
thanks @meltawila this looks good after these last comments LGTM!

src/base/OpenMCProblemBase.C
```diff
@@ -280,7 +279,12 @@ OpenMCProblemBase::fillElementalAuxVariable(const unsigned int & var_num,
 int
 OpenMCProblemBase::nParticles() const
 {
-  return openmc::settings::n_particles;
+  if (*_particles <= 0.0)
```
aprilnovak: Sorry for flip-flopping on this, but I realized that there are some other places in the code where `nParticles()` is being called (in `TallyBase`) so I think we actually do need `nParticles()` to always return the number of particles being used. 

Can we instead have `nParticles()` just return `openmc::settings::n_particles` (and use `nParticles()` as the accessor) and then put the postprocessor-based particle update here instead?

Instead of:
```
  else
  {
    if (isParamValid("particles"))
      openmc::settings::n_particles = OpenMCProblemBase::nParticles();
  }
```

do:

```
else
{
  if (*_particles <= 0.0)
    mooseError("'particles' must be a positive integer. Try `execute_on = 'timestep_begin'` in "
               "your postprocessor and check that the postprocessor value itself is not less than "
               "or equal to zero.")
  openmc::settings::n_particles = *_particles;
}
```

aprilnovak: or maybe we get rid of `nParticles()` altogether to avoid this confusion :) If you think that's the way to go, I'd be on board and just use `openmc::settings::n_particles` as both the getter and setter.

meltawila: I saw TallyBase uses nTotalParticles() as well so I just kept nParticles() as it was originally to be consistent and took the first approach you suggested

PR #1241, opened 2025-11-19
Add delayed-group values of Beta Effective

### LINKED ISSUE #1107: Support multiple delayed groups for kinetics parameters
## Reason
Support for kinetics parameters was added recently in #1105, however it is limited to computing a value of $\beta_{eff}$ for a single delayed group. Once OpenMC adds support for the tallying of multiple delayed groups for $\beta_{i}$ with a `DelayedFilter` we should expose that support in Cardinal.

## Design
A `DelayedFilter` would need to be added to the IFP tally managed by `OpenMCProblemBase` for a user-specified number of groups. The `BetaEffective` post-processor would also require modification to fetch and output a value of $\beta$ for a specific group.

## Impact
This will allow for the calculation of the adjoint-weighted delayed neutron fraction for multiple delayed groups in Cardinal, enabling more accurate transient multiphysics simulations which use point kinetics.

nuclearkevin:
Based on the design musings in https://github.com/neams-th-coe/cardinal/issues/1107#issuecomment-2825149940.
ccing @meltawila.

Closes #1107

### DIFF
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -530,10 +530,13 @@ protected:
   /// The initial OpenMC seed
   const int64_t _initial_seed;
 
-  /// The index of a global tally to accumulate the common scores required for kinetics parameters.
+  /// The index of a global tally which accumulates the common scores required for kinetics parameters.
   int _ifp_common_tally_index = -1;
 
-  /// The index of a global tally to accumulate the common scores required for kinetics parameters.
+  /**
+   * The index of a global tally which accumulates the score required for the delayed group values
+   * of \beta_eff.
+   */
   int _ifp_mg_beta_tally_index = -1;
 
   /// The global tally used to accumulate the common scores required for kinetics parameters.
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -408,7 +408,9 @@ OpenMCProblemBase::initialSetup()
     _ifp_common_tally->set_scores({"ifp-time-numerator", "ifp-denominator", "ifp-beta-numerator"});
     _ifp_common_tally->estimator_ = openmc::TallyEstimator::COLLISION;
 
-    // For \beta_{eff,i}
+    // For \beta_{eff,i}. A separate tally is required when sieving by delayed group to compute
+    // standard deviations and relative errors correctly for the total \beta_eff (due to covariances
+    // between delayed groups).
     _ifp_mg_beta_tally_index = openmc::model::tallies.size();
     _ifp_mg_beta_tally = openmc::Tally::create();
     _ifp_mg_beta_tally->set_scores({"ifp-beta-numerator"});
@@ -833,9 +835,6 @@ OpenMCProblemBase::getCommonKineticsTally()
 const openmc::Tally &
 OpenMCProblemBase::getMGBetaTally()
 {
-  if (!_ifp_mg_beta_tally)
-    mooseError("Internal error: kinetics parameters have not been enabled.");
-
   return *_ifp_mg_beta_tally;
 }
 
--- a/test/tests/neutronics/kinetics/tests
+++ b/test/tests/neutronics/kinetics/tests
@@ -13,7 +13,7 @@
     type = CSVDiff
     input = six_grp.i
     csvdiff = six_grp_out.csv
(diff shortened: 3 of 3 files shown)


include/base/OpenMCProblemBase.h
```diff
@@ -516,11 +530,17 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
   /// The initial OpenMC seed
   const int64_t _initial_seed;
 
-  /// The index of a global tally to accumulate the scores required for kinetics parameters.
-  int _ifp_tally_index = -1;
+  /// The index of a global tally to accumulate the common scores required for kinetics parameters.
```
aprilnovak: Can different comments be used here for line 533 and line 536? Right now they are the same

src/base/OpenMCProblemBase.C
```diff
@@ -400,10 +402,25 @@ OpenMCProblemBase::initialSetup()
...
+    _ifp_common_tally_index = openmc::model::tallies.size();
+    _ifp_common_tally = openmc::Tally::create();
+    _ifp_common_tally->set_scores({"ifp-time-numerator", "ifp-denominator", "ifp-beta-numerator"});
+    _ifp_common_tally->estimator_ = openmc::TallyEstimator::COLLISION;
+
+    // For \beta_{eff,i}
+    _ifp_mg_beta_tally_index = openmc::model::tallies.size();
+    _ifp_mg_beta_tally = openmc::Tally::create();
+    _ifp_mg_beta_tally->set_scores({"ifp-beta-numerator"});
+    _ifp_mg_beta_tally->estimator_ = openmc::TallyEstimator::COLLISION;
```
aprilnovak, line 417: why is this created as a separate tally here?  The same settings and scores are used in `_ifp_common_tally`. Haven't finished reviewing yet so may remove this comment if it becomes clear later!

src/base/OpenMCProblemBase.C
```diff
@@ -793,12 +822,21 @@ OpenMCProblemBase::numCells() const
 }
 
 const openmc::Tally &
-OpenMCProblemBase::getKineticsParamTally()
+OpenMCProblemBase::getCommonKineticsTally()
+{
+  if (!_ifp_common_tally)
```
aprilnovak, line 829: Do we need this check on both `_ifp_common_tally` and `_ifp_mg_beta_tally`? They are created together, seems like we could get away with just one error check. Of course, not harmful to leave in, but we don't have a ton of this internal error checking in Cardinal if there's not a conceivable way we wouldn't have something initialized.

src/postprocessors/BetaEffective.C
```diff
@@ -53,23 +60,54 @@ BetaEffective::BetaEffective(const InputParameters & parameters)
...
+      num_ss = xt::view(
+          common_tally.results_, xt::all(), 2, static_cast<int>(openmc::TallyResult::SUM_SQ))[0];
+      break;
+    case BetaTypeEnum::D_1:
+    case BetaTypeEnum::D_2:
+    case BetaTypeEnum::D_3:
+    case BetaTypeEnum::D_4:
+    case BetaTypeEnum::D_5:
+    case BetaTypeEnum::D_6:
+      n_num = mg_beta.n_realizations_;
```
aprilnovak, line 84: This can go outside the switch statement I think

test/tests/neutronics/kinetics/tests
```diff
@@ -9,6 +9,13 @@
     requirement = "The system shall be able to calculate neutron kinetics parameters."
     capabilities = 'openmc'
   []
+  [six_group_beta]
+    type = CSVDiff
+    input = six_grp.i
+    csvdiff = six_grp_out.csv
+    requirement = "The system shall be able to calculate separate values of beta_eff for each delayed group in the ENDF delayed group structure.."
```
aprilnovak:
```suggestion
    requirement = "The system shall be able to calculate separate values of beta_eff for each delayed group in the ENDF delayed group structure."
```

src/postprocessors/BetaEffective.C
nuclearkevin: Not in this case, as `n_num = common_tally.n_realizations_` when `_beta_type == BetaTypeEnum::Sum`. It's equal to `n_num = mg_beta.n_realizations_` on the other switch branches.

PR #1243, opened 2025-11-20
Improve `check_equal_mapped_tally_volumes` error message.

### LINKED ISSUE #1242: Improve `check_equal_mapped_tally_volumes` error message
## Reason
Fairly recently, a user was running into some issues where they were computing `heating_local` over two different cell regions (fuel and moderator) with a `CellTally`. For this case, it is expected that the volume mapping to cells will be different as the cells themselves have different volumes. They had `check_equal_mapped_tally_volumes = true`, which resulted in an opaque error message that made this difficult to debug until I took a look at the source code for the volume check and realized it didn't discriminate by cell id.

## Design
The `check_equal_mapped_tally_volumes` error message should let people know that they need to disable the check if the `CellTally` maps to more then a single unique cell.

## Impact
Improved user experience, easier model debugging when using non-localized scores with cell tallies.

nuclearkevin:
Closes #1242

### DIFF
--- a/src/tallies/CellTally.C
+++ b/src/tallies/CellTally.C
@@ -204,9 +204,12 @@ CellTally::getTallyCells() const
                  "do actually have the\n"
                  "same value, the volumetric tally will be different because you'll be "
                  "dividing each tally by a\n"
-                 "different mapped MOOSE volume. This can occur if the CellTally maps to different regions in\n"
-                 "your OpenMC geometry while computing the same score over each region. A common example is\n"
-                 "tallying fluxes over multiple regions (e.g. fuel + moderator) instead of a single region\n"
+                 "different mapped MOOSE volume. This can occur if the CellTally maps to different "
+                 "regions in\n"
+                 "your OpenMC geometry while computing the same score over each region. A common "
+                 "example is\n"
+                 "tallying fluxes over multiple regions (e.g. fuel + moderator) instead of a "
+                 "single region\n"
                  "(e.g. fuel).\n\n";
 
           if (_openmc_problem.hasPointTransformations())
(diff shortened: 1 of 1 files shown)


aprilnovak:
did anything actually change here? it looks like just style changes to me

nuclearkevin:
Yes, the error message did change. Here's the old message:
```
The following occurred in the Tally 'Cell' of type CellTally.

Detected un-equal mapped tally volumes!
 cell id 1, instance 0 (of 1) maps to a volume of 13.2213 (cm3)
 cell id 2, instance 0 (of 1) maps to a volume of 10.6346 (cm3).

If the tallied cells in your OpenMC model are of identical volumes, this means that you can get
distortion of the volumetric tally output. For instance, suppose you have two equal-size OpenMC
cells which have the same volume - but each OpenMC cell maps to a MOOSE region of different volume
just due to the nature of the centroid mapping scheme. Even if those two tallies do actually have the
same value, the volumetric tally will be different because you'll be dividing each tally by a
different mapped MOOSE volume.

We recommend re-creating the mesh mirror to have an equal volume mapping of MOOSE elements to each
OpenMC cell. Or, you can disable this check by setting 'check_equal_mapped_tally_volumes = false'.
```

and here's the new message:
```
*** ERROR ***
The following occurred in the Tally 'Cell' of type CellTally.

Detected un-equal mapped tally volumes!
 cell id 1, instance 0 (of 1) maps to a volume of 13.2213 (cm3)
 cell id 2, instance 0 (of 1) maps to a volume of 10.6346 (cm3).

If the tallied cells in your OpenMC model are of identical volumes, this means that you can get
distortion of the volumetric tally output. For instance, suppose you have two equal-size OpenMC
cells which have the same volume - but each OpenMC cell maps to a MOOSE region of different volume
just due to the nature of the centroid mapping scheme. Even if those two tallies do actually have the
same value, the volumetric tally will be different because you'll be dividing each tally by a
different mapped MOOSE volume. This can occur if the CellTally maps to different regions in
your OpenMC geometry while computing the same score over each region. A common example is
tallying fluxes over multiple regions (e.g. fuel + moderator) instead of a single region
(e.g. fuel).

We recommend re-creating the mesh mirror to have an equal volume mapping of MOOSE elements to each
OpenMC cell instance if this tally is intended to compute scores over a single OpenMC cell. If this
tally computes scores over multiple unique cells, you will need to disable this check by setting
'check_equal_mapped_tally_volumes = false'.
```

The diff is a little misleading here as the number of lines didn't change much.

PR #1247, opened 2025-11-20
Remove the error message when using scaling with a MeshTally on the mesh mirror

### LINKED ISSUE #1246: Remove the error message when using scaling with a `MeshTally` on the mesh mirror
## Reason
We currently thrown an error if the user attempts to use `scaling` when using a mesh tally based on the MOOSE mesh. There is no software reason for this as all `LibMesh` tallies get scaled equally, regardless of whether or not they get initialized from a file or a `libMesh::ReplicatedMesh*`.

## Design
Remove the error message and add a test demonstrating scaling works when acting on the mesh mirror.

## Impact
Improved usability of the `scaling` feature.

nuclearkevin:
Closes #1246


test/tests/neutronics/mesh_tally/tests
```diff
@@ -43,12 +45,17 @@
...
-    type = RunException
-    input = one_mesh_no_input_file.i
-    cli_args = 'Problem/scaling=2.0'
-    expect_err = "Directly tallying on the \[Mesh\] is only supported for 'scaling' of unity."
-    requirement = "The system shall error if attempting to directly tally on a MOOSE mesh that has a scaling not equal to 1.0."
+    type = CSVDiff
+    input = scaling_mm.i
+    csvdiff = scaling_mm_out.csv
+    requirement = "The system shall allow users to use a mesh tally based on the mesh mirror with scaling. "
+                  "The gold file contains relative errors for an unscaled version of the test. The test "
```
aprilnovak, line 52: relative errors wouldn't see any difference from normalization though? the volume would cancel out in the numerator (standard deviation of power density) and denominator (mean of power density)

nuclearkevin: For normalization, 100%. Since the mesh gets scaled down in OpenMC the physical size of the elements in the mesh tally (on the OpenMC side) will be smaller if scaling stops working. This reduces the number of hits per element, and so the relative errors will change.

I can also add the maximum and minimum elemental values to this test if you want to hit volume changes for the mean values (as stored on the mesh).

aprilnovak: Yes, but we don't remember why this restriction exists here - so my thinking is that just looking at relative error would not catch if the _normalization_ on the Cardinal side isn't working (even if the mesh is indeed scaled in OpenMC, I'd want to see the actual means for comparison so we can be sure the division by volume is still working correctly).

Adding max/min works for me!

nuclearkevin: Ahhh, gotcha. That works for me - I'll add the extra components to the test shortly.

meltawila: Just wanted to confirm that this change works as expected for my use case with both tally results and std_dev

PR #1254, opened 2025-11-28
Refactor tally system to support multiple tallies with the same score

### LINKED ISSUE #1106: Allow multiple mapped tallies with the same score
## Reason
At present Cardinal's mapped tallies do not allow users to add multiple tallies with the same score. The reason for this limitation is that it would be difficult to normalize tallies as there could be bin overlap. This becomes problematic when dealing with Multi-Group Cross Section (MGXS) generation as certain MGXS' require analog estimators (such as nu-fission when computing the fission neutron spectra). This bleeds into computing neutron production cross sections with an analog estimator (even though they could be computed with tracklength/collision estimators) as well since there can be only one mapped tally scoring nu-fission.

## Design
Allowing multiple mapped tallies with the same scores would require a complete overhaul of tally normalization. For the case where the user does not wish to normalize by a single global tally, they would need to provide the name of a mapped tally to use for normalization on a per-score basis (to avoid bin overlap). If there's only one tally accumulating a particular score, the user wouldn't need to provide a mapped tally for the score (preserving existing behaviour). 

## Impact
This would allow for the use of tracklength/collision estimators in MGXS generation. Incidentally, this would allow different tally spatial discretizations and external filter bin combinations for duplicate scores in the same input file.

nuclearkevin:
To support this change, tally sum checks and global normalization options have been moved into `TallyBase`. This allows users to specify on a per-tally basis if they want error messages thrown on failed normalization or local tally sums not adding up to the global tally sum.

I had to rework spatially distributed mesh tallies in this PR as the previous approach only worked with one score per tally. A new concept has been added which I've termed "tally linkage". This allows tallies to communicate partial sums for normalization purposes. Only one linked tally needs to add a global tally, the rest access it during normalization and error checking. I made this reasonably generic so linkages could be extended to other tally types. i.e. per-pin Zernike/Legendre FETs that are distributed and linked together in a similar manner to spatially-distributed mesh tallies.

I ultimately decided against adding per-score global normalization / sum checking in this PR due to the present size of the diff. That will come in a follow-on PR once this has been merged.

Closes #1106

### DIFF
--- a/test/tests/conduction/boundary_and_volume/prism/tests
+++ b/test/tests/conduction/boundary_and_volume/prism/tests
@@ -29,7 +29,7 @@
     type = RunException
     input = nek.i
     cli_args = "AuxVariables/heat_source/order=FIRST"
-    expect_err = "Cardinal is trying to add an auxiliary variable named 'heat_source', but you already have a variable by this name."
+    expect_err = "Cardinal is trying to add an auxiliary variable named 'heat_source' through the FieldTransfer system"
     requirement = "The system shall error if the user specifies a duplicate variable with a name overlapping with special names reserved for Cardinal data transfers."
     capabilities = 'nekrs'
   []
--- a/test/tests/nek_errors/invalid_field/tests
+++ b/test/tests/nek_errors/invalid_field/tests
@@ -64,7 +64,7 @@
   [duplicate_auxvar]
     type = RunException
     input = auxvar.i
-    expect_err = "Cardinal is trying to add an auxiliary variable named 'temp', but you already have a variable by this name."
+    expect_err = "Cardinal is trying to add an auxiliary variable named 'temp' through the FieldTransfer system"
     requirement = "The system shall error if the user manually specifies a duplicate name for an output field."
     capabilities = 'nekrs'
   []


nuclearkevin:
I'll PR a syntax fix into VTB when this is merged.

include/base/OpenMCCellAverageProblem.h
```diff
@@ -160,11 +170,11 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
   }
 
   /**
-   * Fetch the tally that contains the requested score.
-   * @param[in] score the tally score
-   * @return a vector of Cardinal tally objects containing the scores
+   * get a tally object by it's name.
```
aprilnovak:
```suggestion
   * Get a tally object by its name.
```

include/base/OpenMCCellAverageProblem.h
aprilnovak, line 184: add tally_name documentation

aprilnovak, line 200: add tally_name documentation

include/base/OpenMCCellAverageProblem.h
```diff
@@ -206,6 +220,7 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
   std::vector<const VariableValue *>
   getTallyScoreNeighborVariableValues(const std::string & score,
                                       THREAD_ID tid,
+                                      const std::string & tally_name = "",
```
aprilnovak: add tally_name documentation for doxygen

include/tallies/TallyBase.h
```diff
@@ -218,22 +241,33 @@ class TallyBase : public MooseObject
...
+   * Whether this tally requires a global tally or not.
+   * @return if the tally adds a global tally in addition to a mapped local tally
+   */
+  virtual bool addingGlobalTally() const { return _needs_global_tally; }
+
+  /**
+   * Get the number of tallies linked to this tally for normalization.
+   * @return the number of linked tallies
+   */
+  unsigned int numLinkedTallies() const { return _linked_tallies.size(); }
```
aprilnovak: IMO it's easier to read if we just replace this function by `linkedTallies.size()` in the two places it's called

include/tallies/TallyBase.h
```diff
@@ -289,12 +353,18 @@ class TallyBase : public MooseObject
   /// Sum value of this tally across all bins. Indexed by score.
   std::vector<Real> _local_sum_tally;
 
+  /// Sum value of the global tally associated with this tally object. Indexed by score.
```
aprilnovak: can you elaborate what "indexed by score" means? scores are enumerations, not ints (or do you mean it is an enumeration converted to ints?)

include/tallies/TallyBase.h
```diff
@@ -289,12 +353,18 @@ class TallyBase : public MooseObject
...
+  std::vector<Real> _global_sum_tally;
+
   /**
    * Mean value of this tally across all bins; only used for fixed source mode.
    * Indexed by score.
    */
   std::vector<Real> _local_mean_tally;
 
+  /// Linked sum value across all bins. Indexed by score.
```
aprilnovak, line 369: same comment here regarding indexing comment

src/actions/AddTallyAction.C
```diff
@@ -89,13 +99,15 @@ AddTallyAction::addMeshTally(unsigned int instance, const Point & translation)
   std::string obj_name = _name;
   if (_mesh_translations.size() > 1)
   {
-    obj_name += "_" + Moose::stringify(instance);
+    if (instance != 0)
```
aprilnovak, line 102: IMO we should still print the zero instance if there is > 1 instance (but I agree, it can be suppressed if there's only ever one instance)

src/actions/SetupMGXSAction.C
```diff
@@ -57,6 +57,14 @@ SetupMGXSAction::validParams()
       "The type of estimator to use with the tallies added for MGXS generation. This is not "
       "applied to scattering / fission"
       " scores as the filters applied to those scores only support analog estimators.");
+  params.addParam<bool>("check_tally_sum",
+                        "Whether to check consistency between the local tallies "
+                        "with a global tally sum");
```
aprilnovak:
```suggestion
                        "and a global tally sum. This will require that the integral of the local tally matches a tally with no filters (defined over the entire phase space).");
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -92,6 +92,9 @@ OpenMCCellAverageProblem::validParams()
       "particle source rate (source/sec) for a certain tallies in "
       "eigenvalue mode. In other words, the "
       "source/sec is computed as (power divided by the global value of this tally)");
+  params.addParam<std::string>(
+      "normalization_tally",
```
aprilnovak, line 98: haven't read remainder of src yet - is this the name of a tally object or the score? I'm not clear from just reading this, could you please add something more specific to the docstring?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -785,28 +775,26 @@ OpenMCCellAverageProblem::getTallyTriggerParameters(const InputParameters & para
...
                                                  THREAD_ID tid,
+                                                 const std::string & tally_name,
                                                  const std::string & output,
                                                  bool skip_func_exp)
 {
   std::vector<const MooseVariableFE<Real> *> score_vars;
   for (const auto & t : _local_tallies)
   {
-    if (t->hasScore(score))
+    if (t->hasScore(score) && (t->name() == tally_name || tally_name == ""))
```
aprilnovak: is tally_name an optional parameter for this function, or are there scenarios where the name is actually empty? is this a moose object name or a name in OpenMC?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2675,88 +2617,39 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
...
-          auto local_score = _local_tally_score_map[i].at(_all_tally_scores[global_score]);
-          _local_sum_tally[global_score] += _local_tallies[i]->getSum(local_score);
-          _local_mean_tally[global_score] += _local_tallies[i]->getMean(local_score);
-        }
-      }
-
-      if (_check_tally_sum)
-        for (unsigned int global_score = 0; global_score < _all_tally_scores.size(); ++global_score)
-          checkTallySum(global_score);
+      // Recompute sums and means for tallies that are linked to othher tallies.
```
```suggestion
      // Recompute sums and means for tallies that are linked to other tallies.
```

include/tallies/TallyBase.h
nuclearkevin: This is a vector where each entry corresponds to the same entry in the tally score enumeration vector. As an example, `_global_sum_tally[0]` corresponds to the score `_tally_score[0]`. I'll make this clearer in the comment above the variable.

include/tallies/TallyBase.h
nuclearkevin: Same response as the comment above.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: This is the name of a tally object. I'll make that more clear in the string associated with the parameter.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: `tally_name` (the name of a tally object added in the `[Tallies]` block) is an optional parameter in this function which defaults to an empty string (for the case where all variables associated with a score need to be fetched). I can't really think of any use cases for bulk fetching all of the variables, so if it's preferred I can remove the default and make the parameter mandatory instead to clean this up.

src/actions/AddTallyAction.C
nuclearkevin: This change makes it so the MOOSE object name can be easily referenced by users which need to pick out linked mesh tallies for normalization (instance zero is referenced, which pulls all of the other tallies in as they share a variable). I'm a little confused about what you mean when you say "print the zero instance"?

aprilnovak: nevermind!

src/base/OpenMCCellAverageProblem.C
aprilnovak: if we can make it simpler and more constrained, IMO that's better if there's not yet a use case for the more general case

test/tests/neutronics/tally_system/tests
```diff
@@ -97,4 +89,35 @@
...
+    expect_err = "Cardinal is trying to add an auxiliary variable named 'kappa_fission', but you already have a variable by this name."
+    requirement = "The system shall error if two tallies scoring the same score add the same variable names."
+    capabilities = 'openmc'
+  []
+  [duplicate_scores_separate_tallies]
+    type = RunException
+    input = same_scores.i
+    cli_args = "Problem/assume_separate_tallies=true Problem/Tallies/Cell/normalize_by_global_tally=false"
+    expect_err = "Cannot assume separate tallies when multiple tallies tally the same score!"
+    requirement = "The system shall error if attempting to use separate tallies when adding two tallies with the same score."
```
aprilnovak, line 120: Do we need to make this stricter, and prevent assume_separate whenever there's two tallies (no matter what they're scoring?) (or maybe this is already covered, in which case is this redundant?)

test/tests/neutronics/tally_grad/finite_diff/tests
```diff
@@ -32,9 +32,43 @@
...
+    requirement = "The system shall error if two tallies accumulate the same scores and the user doesn't "
+                  "select one to use for computing tally gradients."
+    expect_err = "When adding more then one tally with kappa-fission in the \[Tallies\] block, the 'tally' parameter is required!"
+    capabilities = 'openmc'
+  []
+  [duplicate_scores_missing_tally]
+    type = RunException
+    input = openmc_dup.i
+    cli_args = "AuxKernels/comp_grad_kappa_fission/tally='na'"
+    requirement = "The system shall error if the user provides a non-existent tally."
```
aprilnovak: there are a lot of good tests we're adding for these types of error checks - would it be worth adding something like a `TallyInterface` that these various classes (auxkernels, postprocessors, indicators) can direct through to avoid code and test duplication?

nuclearkevin: Definitely on board with the idea of a `TallyInterface`. I'll prototype something and push it, let me know what you think!

test/tests/neutronics/tally_system/tests
nuclearkevin: I looked through OpenMC's source for tally scoring, and it appears as if we do need to make this stricter. My previous impression was that OpenMC would consider tally scores when skipping all other tallies. That appears to not be the case.

test/tests/neutronics/tally_grad/finite_diff/tests
aprilnovak: looks good to me! can we then remove some of the tests which cover these error messages? should be able to keep just one "set"

include/base/OpenMCCellAverageProblem.h
```diff
@@ -160,22 +170,24 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
   }
 
   /**
-   * Fetch the tally that contains the requested score.
-   * @param[in] score the tally score
-   * @return a vector of Cardinal tally objects containing the scores
+   * Get a tally object by it's name.
```
aprilnovak:
```suggestion
   * Get a tally object by its name.
```

test/tests/postprocessors/fission_tally_relative_error/tests
```diff
@@ -43,4 +43,40 @@
...
+    requirement = "The system shall allow for the computation of tally relative errors when multiple "
+                  "tallies accumulating the same scores are added to the problem."
+    capabilities = 'openmc'
+  []
+  [duplicate_scores_tally_not_provided]
+    type = RunException
+    input = openmc_dup.i
+    requirement = "The system shall error if two tallies accumulate the same scores and the user doesn't "
+                  "select one to use for computing a relative error metric."
+    expect_err = "When adding more then one tally with kappa-fission in the \[Tallies\] block, the 'tally' parameter is required!"
```
aprilnovak:
```suggestion
    expect_err = "When adding more than one tally with kappa-fission in the \[Tallies\] block, the 'tally' parameter is required!"
```

src/interfaces/TallyInterface.C
```diff
@@ -0,0 +1,88 @@
...
+    _object->mooseError("No tallies are adding a " + score + " score!");
+
+  // When the problem has more then one tally accumulating the given score, the user needs to tell
+  // us which one to use.
+  std::string tally_name;
+  if (_openmc_problem->getNumScoringTallies(score) > 1)
+  {
+    checkRequiredParam(_object->parameters(),
+                       tally_param,
+                       "adding more then one tally with " + score + " in the [Tallies] block");
```
aprilnovak:
```suggestion
                       "adding more than one tally with " + score + " in the [Tallies] block");
```

aprilnovak:
If a user has the same score in > 1 tally, there will be a name conflict for that score (since most of the time, people rely on the default naming in Cardinal). The error they'll get is:

> Cardinal is trying to add an auxiliary variable named 'kappa_fission', but you already have a variable by this name. Please choose a different name for the auxiliary variable you are adding.

which might be a bit opaque, since they may not be aware of the name conflict. Perhaps we could add an input parameter to `CardinalProblem::checkDuplicateVariableName(name)` like `CardinalProblem::checkDuplicateVariableName(name, std::string system)` that will give the user a more helpful place to check their input file?  But open to other ideas, including just adding to the string to say "This name conflict could be arising due to default name choices in systems such as Tallies, FieldTransfers, etc."

aprilnovak:
Also, a stray thought (this does not have to be in this PR but am just posing it for consideration) - now that most of the tally functionality has been ported into the Tally system, do you think the relaxation also belongs there? I think the motivation would be purely from a code maintenance/accessibility standpoint (avoiding exposing tally stuff in the problem class), as I don't envision someone wanting to relax two different tallies with different strategies in practice. What do you think?

test/tests/neutronics/tally_system/tests
```diff
@@ -1,6 +1,6 @@
 [Tests]
```
aprilnovak, line 1: I tested out this branch locally, looks good! I created my own tests, I am wondering if we have coverage of these in our existing suite or if you anticipate needing a dedicated test for these (i.e. something uncovered by what you have here which existing tests don't catch):

- Two tallies which are otherwise identical mesh tallies, but one uses global normalization while the other uses local normalization
- Relaxation applied to > 1 tally

nuclearkevin:
> Also, a stray thought (this does not have to be in this PR but am just posing it for consideration) - now that most of the tally functionality has been ported into the Tally system, do you think the relaxation also belongs there? I think the motivation would be purely from a code maintenance/accessibility standpoint (avoiding exposing tally stuff in the problem class), as I don't envision someone wanting to relax two different tallies with different strategies in practice. What do you think?

I think this would make a great follow-on PR. I'll get an issue up.

test/tests/neutronics/tally_system/tests
nuclearkevin: We have tests for relaxation applied to multiple tallies already [here](https://github.com/neams-th-coe/cardinal/blob/devel/test/tests/neutronics/tally_system/multi_mesh_relax.i) and [here](https://github.com/neams-th-coe/cardinal/blob/devel/test/tests/neutronics/tally_system/multi_cell_relax.i) (from the tally system addition in the summer of 2024). I have a test [here](https://github.com/nuclearkevin/cardinal/blob/refactor_tally_system/test/tests/neutronics/tally_system/same_scores.i) which uses local normalization on a `MeshTally` and global normalization on a `CellTally`, however I don't have anything with two mesh tallies (with different normalization). 

I don't think we need to test two separate mesh tallies with the same score as (but different types of normalization) as the same code paths get hit for normalization regardless of the tally object (mesh or cell). If you think otherwise, I can add one in but I believe we're good on test coverage.

aprilnovak: Nope, that works for me!

PR #1256, opened 2025-12-05
Adding OpenMCCellTransform

### LINKED ISSUE #1255: Control OpenMC universe translations
## Reason
In some transient cases, translating an OpenMC universe can be necessary to avoid restarting the simulation and updating the CSG geometry 

## Design
Add a `OpenMCCellTranslator` user object which controls `openmc_cell_set_translation` OpenMC C interoperable function that can update translations between time steps

## Impact
Enhance user experience

meltawila:
closes #1255


test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,21 @@
...
+    requirement = 'The system shall correctly reinitialize OpenMC coupling when the geometry is translated.'
+                  'This is tested by translating the mesh while also translating OpenMC universe which contains the coupled cells.'
+    capabilities = 'openmc'
+  []
+  [rotate_cells]
+    type = Exodiff
+    input = rotate_cells.i
+    exodiff = rotate_cells_out.e
+    requirement = 'The system shall correctly reinitialize OpenMC coupling when the geometry is rotated.'
+                  'This is tested by rotating the mesh while also rotating OpenMC universe which contains the coupled cells.'
```
aprilnovak, line 26: Can you elaborate more on how you confirmed that the test gives the _result_ you expect? e..g not just "run without complaining," but are you comparing to a separate, standalone OpenMC case where you manually modify the geometry and then match k, tallies, etc.?

meltawila: I am here mainly testing that the mapping remains correct (fuel only maps to fuel) during mesh movement, where OpenMC geometry and Cardinal mesh are moving independently but doing the same geometry motion (as opposed to MoabSkinner where OpenMC geometry is generated from the [Mesh]). I can change the test to do the transforms at the 0th step and compare against results that start with the new state without any movement, if you prefer that

aprilnovak: I think you can leave your tests as-is, perhaps add `reset_seed = true` so that each run starts from the same point? But then yes, do an offline comparison of k

doc/content/source/userobjects/OpenMCCellTransform.md
```diff
@@ -0,0 +1,30 @@
+# OpenMCCellTransform
+
+!syntax description /UserObjects/OpenMCCellTransform
+
+## Description
+
+This user object can be used to either translate or rotate a universe inside an OpenMC cell.
+
+!alert! warning title=The specified cell must be filled with a universe which is what will be transformed
+This userobject doesn't translate or rotate the surfaces of the specifeid cell. Instead, if that cell is filled with a universe, the given transform will be used to shift/rotate that universe. If you want to use moving mesh capabilities, you may consider using the MOAB skinner with DAGMC geometries instead.
```
aprilnovak:
```suggestion
This userobject doesn't translate or rotate the surfaces of the specified cell. Instead, if that cell is filled with a universe, the given transform will be used to shift/rotate that universe. If you want to use moving mesh capabilities for general cell surfaces, you may consider using the [MoabSkinner](MoabSkinner.md) with DAGMC geometries instead.
```

doc/content/source/userobjects/OpenMCCellTransform.md
```diff
@@ -0,0 +1,30 @@
...
+!alert-end!
+
+## Example Input Syntax
+
+To translate a universe, the transform array takes ($\Delta x$, $\Delta y$, $\Delta z$) translations. Below is an example which will translate the universe inside a specified OpenMC cell:
+
+!listing test/tests/userobjects/openmc_cell_transform/translate_cells.i
+  block=UserObjects
+
+To rotate a universe, the transform array takes ($\phi$, $\theta$, $\psi$) which are the rotation angles in degrees about x, y, and z axes. Below is an example which will translate the universe inside a specified OpenMC cell:
```
aprilnovak:
```suggestion
To rotate a universe, the transform array takes ($\phi$, $\theta$, $\psi$) which are the rotation angles in degrees about x, y, and z axes. Below is an example which will rotate the universe inside a specified OpenMC cell:
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  for (const auto & cell_id : _cell_ids)
+  {
+    int32_t index = -1;
+
+    int err = openmc_get_cell_index(cell_id, &index);
+    if (err)
+      mooseError("In attempting to find OpenMC cell with ID " + std::to_string(cell_id) +
+                 ", OpenMC reported:\n\n" + std::string(openmc_err_msg));
+
+    if (_transform_type == "translation")
```
aprilnovak, line 100: Does OpenMC error when calling these two C-API functions if the cell is not filled by a universe? Let's double-check that, otherwise we should add an appropriate check here

include/userobjects/OpenMCCellTransform.h
```diff
@@ -0,0 +1,60 @@
...
+/********************************************************************/
+
+#pragma once
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "GeneralUserObject.h"
+#include "OpenMCBase.h"
+
+#include <vector>
```
aprilnovak:
```suggestion
```

include/userobjects/OpenMCCellTransform.h
```diff
@@ -0,0 +1,60 @@
...
+#include <cstdint>
```
aprilnovak:
```suggestion
```

include/userobjects/OpenMCCellTransform.h
```diff
@@ -0,0 +1,60 @@
...
+  virtual void finalize() override {}
+
+protected:
+  /// OpenMC cell IDs to which the translation will be applied
+  const std::vector<int32_t> _cell_ids;
+
+  /// Transform type: "translation" or "rotation"
+  const MooseEnum _transform_type;
+
+  /// Postprocessors providing the first entry of the transform array
```
aprilnovak:
```suggestion
  /// Postprocessor providing the first entry of the transform array
```

include/userobjects/OpenMCCellTransform.h
```diff
@@ -0,0 +1,60 @@
...
+  const PostprocessorValue * _t0_pp;
+
+  /// Postprocessors providing the second entry of the transform array
```
aprilnovak:
```suggestion
  /// Postprocessor providing the second entry of the transform array
```

include/userobjects/OpenMCCellTransform.h
```diff
@@ -0,0 +1,60 @@
...
+  const PostprocessorValue * _t1_pp;
+
+  /// Postprocessors providing the third entry of the transform array
```
aprilnovak:
```suggestion
  /// Postprocessor providing the third entry of the transform array
```

src/userobjects/OpenMCCellTransform.C
meltawila: Yes, it errors with "Cannot apply a translation to cell {} because it is not filled with another universe".

[here](https://github.com/openmc-dev/openmc/blob/a62e754bbb77646a81bcbe2dd6bcb58e9686286d/src/cell.cpp#L1747C30-L1748C78)

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+    paramError("transform_array",
+               "Provide exactly 3 postprocessors: 'dx dy dz' in mesh units for translation"
+               "transform or 'φ, θ, ψ' in degrees for rotation transform.");
+
+  _t0_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 0);
+  _t1_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 1);
+  _t2_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 2);
+
+  if (_openmc_problem)
+    _scaling = _openmc_problem->scaling();
```
aprilnovak: Do we need to create a member variable named `_scaling`? why not just use `_openmc_problem->scaling()`? It's only called in 3 lines, so IMO it's simpler to remove the extra member variable and just call `_openmc_problem->scaling()` directly

src/userobjects/OpenMCCellTransform.C
aprilnovak: nice - can you add a comment to our cardinal code here just noting that this check is handled by openmc, in case a future cardinal developer forgets?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "OpenMCCellTransform.h"
+#include "CardinalAppTypes.h"
+#include <string>
```
aprilnovak:
```suggestion
```

aprilnovak: General tip: don't add extra includes unless they're really necessary.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "OpenMCCellTransform.h"
+#include "CardinalAppTypes.h"
```
aprilnovak:
```suggestion
```

aprilnovak: if you open up this file, this contains some special execute flags internal to cardinal for controls, which are not necessary in this object. I guess you copied the control object as the starting point for this code which is how it snuck in

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  params += OpenMCBase::validParams();
+
+  params.addRequiredParam<std::vector<int32_t>>(
+      "cell_ids", "List of OpenMC cell IDs whose filled universes will be transformed.");
+
+  MooseEnum transform_type("translation rotation", "translation");
+  params.addParam<MooseEnum>("transform_type",
+                             transform_type,
+                             "Type of transform to apply: 'translation' (dx,dy,dz) or 'rotation'"
+                             "(φ, θ, ψ) in degrees.");
```
aprilnovak: do these symbols show up correctly in the documentation when compiled? also it may be clearer to say rotation angles about x, y, z because not everyone is consistent about these symbols

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+
+  params.addRequiredParam<std::vector<PostprocessorName>>(
+      "transform_array",
```
aprilnovak: I like `vector_value` slightly better to match the naming in `TransformGenerator`, but I don't have super strong feelings about it

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+      "An array of three postprocessors. For translation this array expects (dx, dy, dz) in mesh"
```
aprilnovak: Do you want to clarify that the values don't HAVE to be postprocessors? For constants, this could be a lot more convenient than having people create a bunch of `ConstantPostprocessors`

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  if (_cell_ids.empty())
+    paramError("cell_ids", "At least one OpenMC cell ID must be provided.");
+
+  const auto & t = getParam<std::vector<PostprocessorName>>("transform_array");
+  if (t.size() != 3)
+    paramError("transform_array",
+               "Provide exactly 3 postprocessors: 'dx dy dz' in mesh units for translation"
+               "transform or 'φ, θ, ψ' in degrees for rotation transform.");
+
+  _t0_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 0);
```
aprilnovak:
```suggestion
  _t0_pp = &getPostprocessorValue("transform_array", 0);
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  _t1_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 1);
```
aprilnovak:
```suggestion
  _t1_pp = &getPostprocessorValue("transform_array", 1);
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  _t2_pp = &GeneralUserObject::getPostprocessorValue("transform_array", 2);
```
aprilnovak:
```suggestion
  _t2_pp = &getPostprocessorValue("transform_array", 2);
```

aprilnovak: Your class already inherits from `GeneralUserObject`, so you have access to all the public and protected members of that class (no scoping needed)

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+}
+
+void
+OpenMCCellTransform::execute()
+{
+  const Real a0 = *_t0_pp;
+  const Real a1 = *_t1_pp;
+  const Real a2 = *_t2_pp;
+
+  double vec[3];
```
aprilnovak, line 81: why not set `vec` here to a0, a1, a2 and then only multiply by scaling inside the translation option?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+
+  if (_transform_type == "translation")
+  {
+    vec[0] = _scaling * a0;
+    vec[1] = _scaling * a1;
+    vec[2] = _scaling * a2;
+
+    _console << "Setting OpenMC cell translations for " << _cell_ids.size() << " cell(s) to ("
```
aprilnovak: how about adding the cell ID info here, in case someone had multiple of these UOs in the same simulation?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+             << vec[0] << ", " << vec[1] << ", " << vec[2] << ") cm." << std::endl;
+  }
+  else if (_transform_type == "rotation")
+  {
+    vec[0] = a0;
+    vec[1] = a1;
+    vec[2] = a2;
+
+    _console << "Setting OpenMC cell rotations for " << _cell_ids.size() << " cell(s) to ("
```
aprilnovak: same comment here for adding print about the cell IDs

aprilnovak: also, the name of the UO would suffice if that'd be more concise

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+  }
+  else
+    mooseError("Unhandled transform_type: " + _transform_type);
+
+  for (const auto & cell_id : _cell_ids)
+  {
+    int32_t index = -1;
+
+    int err = openmc_get_cell_index(cell_id, &index);
+    if (err)
```
aprilnovak: use `catchOpenMCError`, it's a bit more compact

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+      mooseError("In attempting to find OpenMC cell with ID " + std::to_string(cell_id) +
+                 ", OpenMC reported:\n\n" + std::string(openmc_err_msg));
+
+    if (_transform_type == "translation")
+      err = openmc_cell_set_translation(index, vec);
+    else
+      err = openmc_cell_set_rotation(index, vec, 3);
```
aprilnovak, line 125: need to technically add another `else` statement here, in case someone adds another option to the enum and forgets it here. it's tedious, but it's a not so uncommon occurence that we should protect against it everywhere.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,132 @@
...
+
+    if (err)
```
aprilnovak: use `catchOpenMCError`

test/tests/userobjects/openmc_cell_transform/plots.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
aprilnovak: do we need this file here? if not, let's just remove the plot export from the `model.py` and delete this file

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,145 @@
...
+    type = KEigenvalue
+  []
+  [power]
+    type = ElementIntegralVariablePostprocessor
+    variable = kappa_fission
+    block = '1'
+  []
+[]
+
+[UserObjects]
```
aprilnovak, line 116: Is all the disp_x, disp_y, disp_z stuff needed in this test case? (the functions and auxvariables/auxkernels) If not, let's remove to simplify

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,145 @@
...
+  [translate_cells]
+    type = OpenMCCellTransform
+    transform_type = 'rotation'
+    transform_array = 'phi theta psi'
+    cell_ids = '2011'
+    execute_on = 'timestep_begin'
```
aprilnovak, line 122: so, does this test apply a new rotation on each time step? if so, that's a better test than doing this just once, but wanted to confirm my understanding. let's therefore use `reset_seed=True` and confirm against a standalone OpenMC case that our modifications are identical to as if done entirely via OpenMC

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,21 @@
+[Tests]
+  design = 'OpenMCCellTransform.md'
+  issues = '#1255'
+
+  [translate_cells]
+    type = Exodiff
+    input = translate_cells.i
+    exodiff = translate_cells_out.e
+    requirement = 'The system shall correctly reinitialize OpenMC coupling when the geometry is translated.'
+                  'This is tested by translating the mesh while also translating OpenMC universe which contains the coupled cells.'
```
aprilnovak, line 10: Can we also add a "sister" test which does *not* translate the mesh? We should then be able to see the "shadow" of the cells as they move (I think as-is, this test would not, for instance, see if we failed to reset the "uncovered" elements to unmapped)

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
meltawila: yes, they are needed. The [Mesh] movement is independent from OpenMC geometry movement and is controlled by disp_x, disp_y, disp_z

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
meltawila: yes, it applies a new rotation on each step controlled by `psi_fn`

test/tests/userobjects/openmc_cell_transform/tests
meltawila: I'm not following here. if we don't translate the mesh the cell mapping will be wrong and the tallies won't be stored correctly

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
aprilnovak: They are needed for the way you are writing the test, but I'm proposing we add another test where the mesh doesn't change, just the OpenMC geometry. The reason I think this is worth doing is, we may want to automatically set `fixed_mesh = False` when one of these `OpenMCCellTransform` UOs is used. Theoretically the OpenMC geometry could be changing, but our [Mesh] is not. There may not be a user who wants to do this, but it's different enough that I think it will improve our testing to include something like this. 

for instance, if we have a cube [Mesh] and a cube OpenMC model, we should be able to see the "shadow" of a moving OpenMC model relative to a stationary [Mesh]. These are the same type as some of the skinner tests.

test/tests/userobjects/openmc_cell_transform/tests
aprilnovak: see comment above

doc/content/source/userobjects/OpenMCCellTransform.md
```diff
@@ -0,0 +1,30 @@
+# OpenMCCellTransform
+
+!syntax description /UserObjects/OpenMCCellTransform
+
+## Description
+
+This user object can be used to either translate or rotate a universe inside an OpenMC cell.
+
+!alert! warning title=The specified cell must be filled with a universe which is what will be transformed
+This userobject doesn't translate or rotate the surfaces of the specified cell. Instead, if that cell is filled with a universe, the given transform will be used to shift/rotate that universe. If you want to use moving mesh capabilities for general cell surfaces, you may consider using the [MoabSkinner](MoabSkinner.md) with DAGMC geometries instead.
```
lewisgross1296: FYI it would be possible to apply a transformation to the cell itself if this is something we'd like to support (not suggesting we add it to this PR). Here's an OpenMC discourse [page](https://openmc.discourse.group/t/hexagonal-prism-aligned-to-z-axis/4808/2) on this topic

lewisgross1296:
```suggestion
This user object can be used to either translate or rotate an `openmc.Universe` which is the fill of an `openmc.Cell`.

!alert! warning title=The specified transformation is only applied to the universe which fills the cell.
This userobject doesn't translate or rotate the surfaces of the specified cell. Instead, the given transform will be used to shift/rotate the universe fill. Note that it is possible to create undesired void regions with a poorly defined cell/universe/transformation combination, which may result in lost particles. If you want to use moving mesh capabilities for general cell surfaces, consider using the [MoabSkinner](MoabSkinner.md) with DAGMC geometries instead.
```

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I think this follows the notation listed in the `openmc.Cell` [documentation](https://docs.openmc.org/en/stable/pythonapi/generated/openmc.Cell.html). I agree that there's so many different versions of which angles are rotating about which axis over internet and textbooks, so it might be clearer to call them `rot_x` `rot_y` and `rot_z`? At the very least, we should reference the `openmc.Cell` documentation and make sure it's consistent with it

src/userobjects/OpenMCCellTransform.C
lewisgross1296: +1

src/userobjects/OpenMCCellTransform.C
lewisgross1296: +1. Just out of curiosity, what if someone applies a Translation UO and a Rotation UO to the same cell? Is (should) this allowed and if so how do we guarantee the order? I guess we might want to warn the user if they apply more than one UO to the same cell. Unless I missed it, I don't see a check that the cells for a UO are not already in another UO, though I'm not sure that's possible. Maybe it's just something the user should notice with these prints

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I think I like the cell IDs more than the name but I wouldn't be opposed to both being printed

test/tests/userobjects/openmc_cell_transform/model.py
```diff
@@ -0,0 +1,88 @@
...
+model.settings.run_mode = "eigenvalue"
+model.settings.batches = 50
+model.settings.inactive = 10
+model.settings.particles = 1000
+
+lower = (-2, -2, 0)
+upper = (16, 16, root_half_z*2)
+
+model.settings.source = openmc.IndependentSource(
+    space=openmc.stats.Box(lower_left=lower, upper_right=upper, only_fissionable=False)
```
lewisgross1296: Isn't `only_fissionable` deprecated? I think you can just omit it now

test/tests/userobjects/openmc_cell_transform/model.py
```diff
@@ -0,0 +1,88 @@
...
+plot = openmc.Plot()
+plot.filename = "lattice_xy"
+plot.origin = (0.0, 0.0, 15.0)
+plot.width = (6.0 * pitch, 6.0 * pitch)
+plot.pixels = (800, 800)
+plot.color_by = "material"
+plot.basis = "xy"
+
+model.plots = openmc.Plots([plot])
+model.plots.export_to_xml()
```
lewisgross1296:
```suggestion
```

test/tests/userobjects/openmc_cell_transform/model.py
```diff
@@ -0,0 +1,88 @@
...
+
+lower = (-2, -2, 0)
+upper = (16, 16, root_half_z*2)
+
+model.settings.source = openmc.IndependentSource(
+    space=openmc.stats.Box(lower_left=lower, upper_right=upper, only_fissionable=False)
+)
+
+
+model.export_to_xml()
```
lewisgross1296: might be worth considering `export_to_model_xml()` which is the future direction of OpenMC unless there's a Cardinal reason to keep them separated. would reduce the number of files in the repo too

test/tests/userobjects/openmc_cell_transform/plots.xml
lewisgross1296: +1

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,145 @@
...
+    expression = (cos(5*t*3.14159265359/180)-1)*(x-x0)-sin(5*t*3.14159265359/180)*(y-y0)
+    symbol_names = 'x0 y0'
+    symbol_values = '0 0'
+  []
+  [disp_y_fn]
+    type = ParsedFunction
+    expression = sin(5*t*3.14159265359/180)*(x-x0)+(cos(5*t*3.14159265359/180)-1)*(y-y0)
+    symbol_names = 'x0 y0'
+    symbol_values = '0 0'
+  []
```
lewisgross1296, line 97: Do you plan to change `x0` and `y0` ever? Maybe not really necessary to include them and just put the 0s in the `expression` line?

doc/content/source/userobjects/OpenMCCellTransform.md
meltawila: I believe that this is only possible from the Python side which we cannot really use in Cardinal during runtime between time steps, which is the motivation here, but lmk if I'm wrong!

lewisgross1296: Ah good point, I think I was mistaken. Though, if this was something people wanted, we could maybe create a C-API function similar to `extern "C" int openmc_cell_set_rotation`. Definitely outside the scope of this PR, feel free to resolve this comment.

meltawila: I think this would only happen with cell transformation (available only using the Python API) and should not be a concern here, right?

> Note that it is possible to create undesired void regions with a poorly defined cell/universe/transformation combination, which may result in lost particles.

src/userobjects/OpenMCCellTransform.C
meltawila: The symbols show up correctly in the documentation, and yes I wanted to be consistent with OpenMC documentation but lmk if you prefer otherwise. The "the rotation angles in degrees about x, y, and z axes." description is already in docs. under example input syntax.

https://mooseframework.inl.gov/cardinal/docs/PRs/1256/site/source/userobjects/OpenMCCellTransform.html

doc/content/source/userobjects/OpenMCCellTransform.md
aprilnovak: No, it is possible for "undefined" regions to still become exposed if translating/rotating universes. Take a look at the first example in this notebook: https://github.com/aprilnovak/openmc-education/blob/master/universes-lattices-02/solution-02.ipynb - if you set a transmission BC on the `rectangle` surface, particles will try to cross it within its contained cell and not find anything defined on the other side (seg fault).

src/userobjects/OpenMCCellTransform.C
meltawila: Translation and rotation of the same cell is not an issue. The execution order will follow the userobject execution order

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,126 @@
...
+                                          std::string(openmc_err_msg));
+
+    if (_transform_type == "translation")
+    {
+      // If a user tried to apply translation on a cell that doesn't contain a filled universe,
+      // OpenMC will return an error.
+      err = openmc_cell_set_translation(index, vec);
+      _console << "Setting OpenMC cell translations for cell with ID" + std::to_string(cell_id) +
+                      "to ("
+               << vec[0] << ", " << vec[1] << ", " << vec[2] << ") cm." << std::endl;
```
lewisgross1296, line 107: Is the user required to input translations in cm? OpenMCCellTransform.C mentions it's in [Mesh] units.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,126 @@
...
+  const auto & t = getParam<std::vector<PostprocessorName>>("vector_value");
+  if (t.size() != 3)
+    paramError("vector_value",
+               "Provide exactly 3 values/postprocessors: 'dx dy dz' in mesh units for translation"
+               "transform or 'φ, θ, ψ' in degrees for rotation transform.");
+
+  _t0_pp = &getPostprocessorValue("vector_value", 0);
+  _t1_pp = &getPostprocessorValue("vector_value", 1);
+  _t2_pp = &getPostprocessorValue("vector_value", 2);
+}
```
lewisgross1296, line 76: This might be more of a MOOSE question, but it seems here like you can specify both an actual Postprocessor or a value. How does `getPostProcessorValue` respond when you specify a number instead of a name (like I see in a few tests)? Does it just convert it for you?

Could we also add a check for the rotation case that the rotation is bounded [0,360). In the future, when we do criticality searches, there will be multiple solutions if your angular allowed range is anything, which we very much want to prevent.

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,130 @@
...
+  [disp_x_fn]
+    type = ParsedFunction
+    expression = (cos(5*t*3.14159265359/180)-1)*x-sin(5*t*3.14159265359/180)*y
+  []
+  [disp_y_fn]
+    type = ParsedFunction
+    expression = sin(5*t*3.14159265359/180)*x+(cos(5*t*3.14159265359/180)-1)*y
+  []
+[]
```
lewisgross1296, line 98: Do we need any of these `disp` related entries in the input file for a rotation only test? Might be best to only include the bare minimum to test the functionality

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,130 @@
...
+    pattern = '0 0;
+               0 0'
+    assign_type = 'cell'
+    id_name = 'pin_id'
+  []
+  [delete]
+    type = BlockDeletionGenerator
+    input = assembly
+    block = 2
+  []
```
lewisgross1296, line 25: It's not clear to me how you specify what is block 1 and what is block 2, but perhaps I'm just unfamiliar with this MeshGenerator / those are defaults. Looking at the rest of the input file, I can tell that block is 1 fuel, but maybe a comment would help make this more clear?

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,130 @@
...
+
+
+[GlobalParams]
+  displacements = 'disp_x disp_y disp_z'
+  use_displaced_mesh = true
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  cell_level = 2
```
lewisgross1296, line 49: I think it would be good to have a comment explaining why we're `cell_level = 2`, which originates from the requirement that OpenMC cells which get transformations/rotations must be filled with a universe containing a cell.

lewisgross1296 (commented):
Nothing major, just some more thoughts I had

src/userobjects/OpenMCCellTransform.C
aprilnovak: Throughout cardinal, we always use the convention that the dimensions in the `openmc.i` are in the same units as the [Mesh]. I think the code in this PR does the appropriate multiplication by `_scaling` before using it in OpenMC.

src/userobjects/OpenMCCellTransform.C
aprilnovak: Yes, this is a nice feature of MOOSE in that it will convert it to a postprocessor internally for you, if you just enter a number.

I think that's a good idea to limit the rotation angle to [0, 360) - @meltawila if we don't already have that, please add it as a `RangeCheckedParam`

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
meltawila: Yes, we do. This test checks OpenMC geometry mapping to the MOOSE mesh. Testing OpenMCCellTransform on a non-moving mesh was added in the cells_shadow test

meltawila: In fact for OpenMCCellAverageProblem to redo the cells-to-elements mapping at every step, we need to include displacement variables, so when testing on a non-moving mesh, I'd still add displacement variables and set them to 0. In practice this should make the workflow more convenient as it auto-detects mesh movement (previously we'd use a fixed_mesh=false parameter in OpenMCCellAverageProblem, but we got rid of that a while ago)

src/userobjects/OpenMCCellTransform.C
meltawila: I added the range check for [0, 360), but used if conditions instead of `RangeCheckedParam` since the same vector can be used for translations as well

src/userobjects/OpenMCCellTransform.C
aprilnovak: I suggest we make this entirely self-explanatory here as well in the docstring - not all users read all aspects of the documentation thoroughly :)

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+  params += OpenMCBase::validParams();
+
+  params.addRequiredParam<std::vector<int32_t>>(
+      "cell_ids", "List of OpenMC cell IDs whose filled universes will be transformed.");
+
+  MooseEnum transform_type("translation rotation", "translation");
+  params.addParam<MooseEnum>("transform_type",
+                             transform_type,
+                             "Type of transform to apply: 'translation' (dx,dy,dz) or 'rotation'"
+                             "(φ, θ, ψ) in degrees.");
```
aprilnovak:
```suggestion
                             "(φ, θ, ψ) in degrees, where the angles are the rotations about the x, y, and z axes, respectively.");
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+
+  params.addRequiredParam<std::vector<PostprocessorName>>(
+      "vector_value",
+      "An array of three values/postprocessors. For translation this array expects (dx, dy, dz) in "
+      "mesh"
+      " units. For rotation this array expects 'φ, θ, ψ' in degrees.");
+
+  params.addClassDescription(
+      "UserObject that applies either translation or rotation on one or more OpenMC cells. "
+      "The transform is driven by a transform array of three MOOSE postprocessors");
```
aprilnovak:
```suggestion
      "The transform is driven by a transform array of three MOOSE postprocessors/scalar values");
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+    vec[2] *= _openmc_problem->scaling();
+  }
+
+  for (const auto & cell_id : _cell_ids)
+  {
+    int32_t index = -1;
+
+    int err = openmc_get_cell_index(cell_id, &index);
+    catchOpenMCError(err,
+                     "In attempting to find OpenMC cell with ID " + std::to_string(cell_id) +
```
aprilnovak: look at `UserErrorChecking.C` - the `catchOpenMCError` will handle the "In attempting to..." and other parts of this string. You want to call this just slightly differently; I'll just write the comment one time, but I think you have a few `catchOpenMCError` calls which need to be modified

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+                         ", OpenMC reported:\n\n" + std::string(openmc_err_msg));
+
+    if (_transform_type == "translation")
+    {
+      // If a user tried to apply translation on a cell that doesn't contain a filled universe,
+      // OpenMC will return an error.
+      err = openmc_cell_set_translation(index, vec);
+      _console << "Setting OpenMC cell translations for cell with ID" + std::to_string(cell_id) +
```
aprilnovak:
```suggestion
      _console << "Setting OpenMC cell translations for cell with ID " + std::to_string(cell_id) +
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+                      "to ("
+               << vec[0] << ", " << vec[1] << ", " << vec[2] << ") cm." << std::endl;
+    }
+    else if (_transform_type == "rotation")
+    {
+      if (vec[0] < 0 || vec[0] >= 360 || vec[1] < 0 || vec[1] >= 360 || vec[2] < 0 || vec[2] >= 360)
+        mooseError("Rotation angles must be in the range [0, 360) degrees.");
```
aprilnovak: Let's make this a `paramError`, it gives slightly better debugging info by printing the line number in the input file (and this error is pretty clearly tied to something wrong with a given parameter). We may want to make it three different `paramError`'s then, to correctly point to the parameter name which is bad.

Since that'll be a bit verbose, I'm also ok if we alternatively keep this `mooseError` and just modify the text to name `'vector_value'` in there somewhere to help the user.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+      // If a user tried to apply rotation on a cell that doesn't contain a filled universe,
+      // OpenMC will return an error.
+      err = openmc_cell_set_rotation(index, vec, 3);
+      _console << "Setting OpenMC cell rotations for cell with ID" + std::to_string(cell_id) +
```
aprilnovak:
```suggestion
      _console << "Setting OpenMC cell rotations for cell with ID " + std::to_string(cell_id) +
```

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,30 @@
+[Tests]
```
aprilnovak, line 1: You'll need to have `RunException`-type tests to catch various errors - your source code is good, but we need to cement those catches in our test suite or we're only partway there. I have been surprised sometimes when I think my error catch is written correctly, but it does not actually trigger when I expect it to (so you can catch issues in how you detect the error, or sometimes even find other errors you forgot to protect against).

Here's what I think you need:

- user-provided cell ID needs to exist
- user-provided cell ID list cannot be empty
- Needs to be a universe fill (for both translation and rotation)
- Angles in [0, 360)

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,30 @@
...
+  design = 'OpenMCCellTransform.md'
+  issues = '#1255'
+
+  [translate_cells]
```
aprilnovak, line 5: It may also be worth adding a test which exercises the functionality where you actually have a list of cell IDs > length 1? Sometimes I've done silly things and not actually had a loop, for instance, while testing the code.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,128 @@
...
+registerMooseObject("CardinalApp", OpenMCCellTransform);
+
+InputParameters
+OpenMCCellTransform::validParams()
+{
+  InputParameters params = GeneralUserObject::validParams();
+  params += OpenMCBase::validParams();
+
+  params.addRequiredParam<std::vector<int32_t>>(
+      "cell_ids", "List of OpenMC cell IDs whose filled universes will be transformed.");
```
aprilnovak, line 33: What do you think should happen if the user does `cell_ids = '24 24'`? I think that should be a mistake (no duplicates allowed), in which case you may want to have the `_cell_ids` actually be a `std::set` (and then check that the length of `_cell_ids` is the same as `getParam<std::vector<int32_t>>("cell_ids")`

test/tests/userobjects/openmc_cell_transform/tests
aprilnovak: You can use `cli_args` for some of these (and could modify your test a little bit to introduce non-universe fills somewhere) to cut down on the need to add more directories/`.py` files

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,51 @@
+import openmc
```
aprilnovak, line 2: Are these files in the `cells_shadow` directory actually used anywhere? It's not clear to me, since there's not any `tests` or `.i` files in this folder. If they're not used, let's delete the directory.

meltawila: Yes, it’s in …/userobjects/openmc_cell_transform/tests

I wantex to have all the relevant tests in the same file but added the model in a seperate folder since the .xml is different. Should this has its own tests file?

test/tests/userobjects/openmc_cell_transform/cells_shadow.i
```diff
@@ -0,0 +1,86 @@
+[Mesh]
```
aprilnovak, line 1: When running these input files, I get a warning:

> *** Warning ***
/home/ajnovak2/tmep/cardinal/test/tests/userobjects/openmc_cell_transform/cells_shadow.i:32.1:
The following occurred in the Problem 'MOOSE Problem' of type OpenMCCellAverageProblem.

> Your problem has a moving mesh, but you have not provided a 'skinner'. The [Mesh] will move, but the underlying OpenMC geometry will remain unchanged. Unexpected behavior may occur.


I think we probably want to stop this warning from printing if one of the `OpenMCCellTransform` objects prints or else it could confused the user. I'd suggest we stop it from printing if `OpenMCCellAverageProblem` can find one of these user objects OR a skinner is used (you can look in `OpenMCProblemBase::getOpenMCUserObjects()` for an example of how the problem class can query which user objects exist). Depending on where the skinner warning is printed, it may need to move to `initialSetup` or something, since UOs may not be available yet in the constructor.

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
aprilnovak: Ah I see, I missed the `input` you have in the `.i` file pointing here. I'd suggest we put the `tests` part relevant to this input just into this directory (don't worry about the raw number of files added in this PR, IMO this is clearer organization)

test/tests/userobjects/openmc_cell_transform/tests
meltawila: The first 3 checks are handled by OpenMC -I added a comment that OpenMC checks for these errors- and OpenMC will show either:
```
    if (model::cells[index]->fill_ == C_NONE) {
      set_errmsg(fmt::format("Cannot apply a rotation to cell {}"
                             " because it is not filled with another universe",
        index));
```
or
 ```
   if (model::cells[index]->fill_ == C_NONE) {
      set_errmsg(fmt::format("Cannot apply a translation to cell {}"
                             " because it is not filled with another universe",
        index));
```
I added RunException tests for invalid rotation angles, will push these shortly

File: test/tests/userobjects/openmc_cell_transform/multiple_cells/model.py [hunk 50]
```diff
@@ -0,0 +1,92 @@
+import openmc
```
aprilnovak, line 1: the `translate_cells_out.e` file is huge (for version controlling) - can the test still accomplish it's goal but with a smaller mesh?

meltawila: yes, a coarser mesh should be fine

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1480,16 +1494,19 @@ OpenMCCellAverageProblem::getMaterialFills()
...
 
     _cell_to_material[cell_info] = material_index;
     vt.addRow(printCell(cell_info), materialID(material_index));
   }
 
   if (_verbose && _specified_density_feedback)
   {
-    _console << "\n ===================>       OPENMC MATERIAL MAPPING       <====================\n" << std::endl;
-    _console <<   "           Cell:  OpenMC cell receiving density feedback" << std::endl;
+    _console
```
aprilnovak, line 1507: FWIW, you might want to check your text editor to see if you have some auto formatting thing happening each time you save the file. It's leading to a lot of line diffs which aren't actually diffs. I don't care for cardinal, but it'd  be better etiquette on other projects to have only meaningful line diffs show up since other developers might not like it. Nothing to change here, just wanted to note.

meltawila: I ran clang-format on this file (all .C and .h in this PR) to pass MOOSE pre-check, which updated this formatting. when I don't do this I fail MOOSE pre-check every time.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -523,6 +522,15 @@ OpenMCCellAverageProblem::initialSetup()
 
   getOpenMCUserObjects();
 
+  if (_use_displaced && !_using_skinner && !hasCellTransform())
+    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner'. The "
```
```suggestion
    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner' or an OpenMCCellTransform user object (both of which move the OpenMC geometry). The "
```

aprilnovak: for the user, the correct fix may be to add one of these UOs, not the skinner, so I think this error message is slightly clearer

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,137 @@
...
+      err = openmc_cell_set_translation(index, vec);
+      _console << "Setting OpenMC cell translations for cell with ID " + std::to_string(cell_id) +
+                      " to ("
+               << vec[0] << ", " << vec[1] << ", " << vec[2] << ") cm." << std::endl;
+    }
+    else if (_transform_type == "rotation")
+    {
+      if (vec[0] < 0 || vec[0] >= 360)
+        paramError("vector_value",
+                   "Rotation angle φ must be in the range [0, 360) degrees. Got invalid φ = " +
```
aprilnovak:
```suggestion
                   "Rotation angle φ must be in the range (0, 360] degrees. Got invalid φ = " +
```

The open bracket ( means "not included" whereas the closed bracket [ means it is included. You have the notation swapped in these three error messages.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,137 @@
...
+                       std::to_string(vec[0]) + " degrees.");
+      if (vec[1] < 0 || vec[1] >= 360)
+        paramError("vector_value",
+                   "Rotation angle θ must be in the range [0, 360) degrees. Got invalid θ = " +
```
aprilnovak:
```suggestion
                   "Rotation angle θ must be in the range (0, 360] degrees. Got invalid θ = " +
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,137 @@
...
+                       std::to_string(vec[1]) + " degrees.");
+      if (vec[2] < 0 || vec[2] >= 360)
+        paramError("vector_value",
+                   "Rotation angle ψ must be in the range [0, 360) degrees. Got invalid ψ = " +
```
aprilnovak:
```suggestion
                   "Rotation angle ψ must be in the range (0, 360] degrees. Got invalid ψ = " +
```

test/tests/userobjects/openmc_cell_transform/gold/cells_shadow_out.e
aprilnovak, line 1: Can this be made smaller and still accomplish the purposes of the test?

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,51 @@
...
+    exodiff = cells_shadow_out.e
+    requirement = 'The system shall correctly reinitialize OpenMC coupling when the geometry is translated.'
+                  'This is tested by translating an OpenMC geometry on a stationary mesh and ensure elements are reset correctly.'
+    capabilities = 'openmc'
+  []
+  [invalid_phi]
+    type = RunException
+    input = rotate_cells.i
+    cli_args = 'UserObjects/rotate_cells/vector_value="360 0 0"'
+    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle φ must be in the range \[0, 360\) degrees. Got invalid φ = 360.000000 degrees.'
```
aprilnovak:
```suggestion
    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle φ must be in the range \(0, 360\] degrees. Got invalid φ = 360.000000 degrees.'
```

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,51 @@
...
+    capabilities = 'openmc'
+  []
+  [invalid_theta]
+    type = RunException
+    input = rotate_cells.i
+    cli_args = 'UserObjects/rotate_cells/vector_value="0 360 0"'
+    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle θ must be in the range \[0, 360\) degrees. Got invalid θ = 360.000000 degrees.'
```
aprilnovak:
```suggestion
    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle θ must be in the range \(0, 360\] degrees. Got invalid θ = 360.000000 degrees.'
```

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,51 @@
...
+    capabilities = 'openmc'
+  []
+  [invalid_psi]
+    type = RunException
+    input = rotate_cells.i
+    cli_args = 'UserObjects/rotate_cells/vector_value="0 0 360"'
+    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle ψ must be in the range \[0, 360\) degrees. Got invalid ψ = 360.000000 degrees.'
```
aprilnovak:
```suggestion
    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle ψ must be in the range \(0, 360\] degrees. Got invalid ψ = 360.000000 degrees.'
```

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,51 @@
...
+    capabilities = 'openmc'
+  []
+[]
```
aprilnovak, line 71: There are a few more error checks that aren't explicitly tested here. Let's make sure to add tests for:

- `cell_ids` being empty string
- `vector_value` not having length 3

I think it'd also be nice to have a case that actually uses `scaling` to confirm that this aspect works correctly (I have myself messed it up before, so even though the code looks foolproof it would be good to have). Can you make a test where the [Mesh] and displacements are all in units of meters and then set `scaling = 100` to show that the identical tallies are obtained to the reference case?

test/tests/userobjects/openmc_cell_transform/cells_shadow.i
```diff
@@ -0,0 +1,86 @@
...
+    xmin = -12
+    xmax =  12
+    ymin = -12
+    ymax =   4
+    zmin =   0
+    zmax =  30
+  []
+[]
+
+[GlobalParams]
```
aprilnovak: Can this be removed now that you changed how the `_reinit_coupling` is set? I think you should no longer need the displacement variables to exist

test/tests/userobjects/openmc_cell_transform/cells_shadow.i
```diff
@@ -0,0 +1,86 @@
...
+[Variables]
+  [disp_x]
+  []
+  [disp_y]
+  []
+  [disp_z]
+  []
+[]
+
+[Problem]
```
aprilnovak, line 17: The comment in `cells_shadow/model.py` says that this file should point to `cells_shadow/model.xml`, but I don't see the file path provided here?

meltawila: it's in the test cli_args

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+xmax = openmc.XPlane(surface_id=2, x0= 12.0, boundary_type="reflective")
+ymin = openmc.YPlane(surface_id=3, y0=-12.0, boundary_type="reflective")
+ymax = openmc.YPlane(surface_id=4, y0=  4.0, boundary_type="reflective")
+zmin = openmc.ZPlane(surface_id=5, z0=  0.0, boundary_type="reflective")
+zmax = openmc.ZPlane(surface_id=6, z0= 30.0, boundary_type="reflective")
+
+root_region = +xmin & -xmax & +ymin & -ymax & +zmin & -zmax
+
+fuel_radius = 1.0
+fuel_cyl = openmc.ZCylinder(surface_id=11, r=fuel_radius)
```
aprilnovak:
```suggestion
fuel_cyl = openmc.ZCylinder(r=fuel_radius)
```

aprilnovak: unless these IDs are strictly necessary (like it is for the moving cell), I suggest to remove. Someone trying to learn how to use this user object might look at these tests and get confused about what is required vs. what just happens to be in this test

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+
+moving_fuel_cell = openmc.Cell(cell_id=101, fill=fuel,  region=-fuel_cyl)
```
aprilnovak:
```suggestion
moving_fuel_cell = openmc.Cell( fill=fuel,  region=-fuel_cyl)
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+moving_water_cell = openmc.Cell(cell_id=102, fill=water, region=+fuel_cyl)
```
aprilnovak:
```suggestion
moving_water_cell = openmc.Cell(fill=water, region=+fuel_cyl)
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+
+moving_univ = openmc.Universe(universe_id=10, cells=[moving_fuel_cell, moving_water_cell])
```
aprilnovak:
```suggestion
moving_univ = openmc.Universe(cells=[moving_fuel_cell, moving_water_cell])
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+
+moving_root_cell = openmc.Cell(cell_id=2011, fill=moving_univ, region=root_region)
+
+root_universe = openmc.Universe(universe_id=0, cells=[moving_root_cell])
```
aprilnovak:
```suggestion
root_universe = openmc.Universe(cells=[moving_root_cell])
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+model.geometry = openmc.Geometry(root_universe)
+
+model.settings.run_mode = "eigenvalue"
```
aprilnovak:
```suggestion
```

aprilnovak: The default run mode is already eigenvalue, so this is not necessary

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
+# This input file is used by ../cells_shadow.i test where xml_directory parameter points here.
+import openmc
+
+model = openmc.Model()
+
+fuel = openmc.Material(material_id=1, name="fuel")
```
aprilnovak:
```suggestion
fuel = openmc.Material(name="fuel")
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+fuel.set_density("g/cm3", 10.0)
+fuel.add_nuclide("U235", 1.0)
+fuel.add_nuclide("O16", 2.0)
+
+water = openmc.Material(material_id=2, name="water")
```
aprilnovak:
```suggestion
water = openmc.Material(name="water")
```

test/tests/userobjects/openmc_cell_transform/cells_shadow/model.py
```diff
@@ -0,0 +1,52 @@
...
+water.set_density("g/cm3", 1.0)
+water.add_nuclide("H1", 2.0)
+water.add_nuclide("O16", 1.0)
+
+model.materials = openmc.Materials([fuel, water])
```
aprilnovak:
```suggestion
```

aprilnovak: When using the model object, you actually don't need to create an `openmc.Materials` collection like this - any materials used in the cells will automatically be written to the `model.xml` appropriately

test/tests/userobjects/openmc_cell_transform/cells_shadow.i
aprilnovak: Ah, I see it now - I'd suggest we move it back into the main input file. If it's easy to do, we should try and make the tests runnable without cli_args (and use cli_args to make one-line changes like inducing a failure). In this case, none of the other tests in the `tests` file run the `cells_shadow.i` test, so there's not a strong reason to have this particular line in `cli_args` (and it certainly confused me when trying to run the input!)

test/tests/userobjects/openmc_cell_transform/empty_cell_ids.i
```diff
@@ -0,0 +1,132 @@
...
+    block = '1'
+  []
+[]
+
+[UserObjects]
+  [translate_cells]
+    type = OpenMCCellTransform
+    transform_type = 'translation'
+    vector_value = 'shift_x shift_y 0'
+    cell_ids = ''
```
aprilnovak, line 120: For this test, I think this is a great use case for `cli_args` - we're only changing one line, and we're doing so in order to induce a failure. Why not do the following, so that we don't need to add this entire input?

```
cli_args = "UserObjects/translate_cells/cell_ids=''"
```

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -0,0 +1,64 @@
...
+    exodiff = cells_shadow_out.e
+    requirement = 'The system shall correctly reinitialize OpenMC coupling when the geometry is translated.'
+                  'This is tested by translating an OpenMC geometry on a stationary mesh and ensure elements are reset correctly.'
+    capabilities = 'openmc'
+  []
+  [invalid_phi]
+    type = RunException
+    input = rotate_cells.i
+    cli_args = 'UserObjects/rotate_cells/vector_value="361 0 0"'
+    expect_err = 'UserObjects/rotate_cells/vector_value: Rotation angle φ must be in the range \(0, 361\] degrees. Got invalid φ = 361.000000 degrees.'
```
aprilnovak: where did this 361 come from? shouldn't it be 360? (in the error message, it should show (0, 360] )

test/tests/userobjects/openmc_cell_transform/empty_cell_ids.i
meltawila: I tried it first. MOOSE ignored `cell_ids` parameter and continued the run when it's provided from cli_args (maybe because it's duplicate in the original file), but from this new file the expected error shows correctly

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,133 @@
...
+
+
+[GlobalParams]
+  displacements = 'disp_x disp_y disp_z'
+  use_displaced_mesh = true
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  # Updating cell_level to 2 to account for the nested
```
aprilnovak: What is this comment trying to convey exactly? The cells_shadow test uses a cell level of 1, but those tests are unrelated. This comment sounds like there is a comparison being made to some other scenario/example, but I'm not sure what.

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
```diff
@@ -0,0 +1,133 @@
...
+    id_name = 'pin_id'
+  []
+  [delete]
+    # Only keeping block 1 (fuel pins) for this test
+    type = BlockDeletionGenerator
+    input = assembly
+    block = 2
+  []
+  [extrude]
+    type = MeshExtruderGenerator
```
aprilnovak: I get a deprecation warning about this object - suggest to update to `AdvancedExtruderGenerator`

test/tests/userobjects/openmc_cell_transform/translate_cells.i
```diff
@@ -0,0 +1,132 @@
...
+    id_name = 'pin_id'
+  []
+  [delete]
+    # Only keeping block 1 (fuel pins) for this test
+    type = BlockDeletionGenerator
+    input = assembly
+    block = 2
+  []
+  [extrude]
+    type = MeshExtruderGenerator
```
aprilnovak: I get a deprecation warning about this object - suggest to update to `AdvancedExtruderGenerator`

test/tests/userobjects/openmc_cell_transform/translate_cells.i
```diff
@@ -0,0 +1,132 @@
...
+
+
+[GlobalParams]
+  displacements = 'disp_x disp_y disp_z'
+  use_displaced_mesh = true
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  # Updating cell_level to 2 to account for the nested
```
aprilnovak: Same comment here about this comment

test/tests/userobjects/openmc_cell_transform/empty_cell_ids.i
aprilnovak: huh, that's weird. Must be some default in MOOSE for order of precedence. Sounds good!

test/tests/userobjects/openmc_cell_transform/rotate_cells.i
meltawila: I thought that this model is not intuitive for someone not familiar with cell levels, not related to cells_shadow test at all, since we add the containing cell in OpenMC but only the filled universe couples to MOOSE mesh. I can remove it

aprilnovak: it's fine to keep, maybe I'd just change "Update" to "setting" and add a brief description of why level 2 is needed for this geometry (as opposed to just directing them to docs)

src/userobjects/OpenMCCellTransform.C
```diff
@@ -0,0 +1,137 @@
...
+InputParameters
+OpenMCCellTransform::validParams()
+{
+  InputParameters params = GeneralUserObject::validParams();
+  params += OpenMCBase::validParams();
+
+  params.addRequiredParam<std::vector<int32_t>>(
+      "cell_ids", "List of OpenMC cell IDs whose filled universes will be transformed.");
+
+  MooseEnum transform_type("translation rotation", "translation");
```
gonuke, line 35: Is this right? I am just guessing how these `MooseEnum` types work, but I'd expect this to be
```suggestion
  MooseEnum transform_type("rotation", "translation");
```
What am I missing @meltawila ?

meltawila: The first string is the list of enumeration values, and the second is the default value.

https://mooseframework.inl.gov/docs/doxygen/moose/classMooseEnum.html#ae0596da46f46c381310852331eeb925f

`MooseEnum::MooseEnum	(	std::string 	names, std::string 	default_name = "", bool 	allow_out_of_range = false )`

PR #1257, opened 2025-12-05
Updates for NekRS v25

nandu90:
Work in progress


src/base/NekInterface.C
```diff
@@ -293,19 +303,17 @@ initializeScratch(const unsigned int & n_slots)
...
 }
 
 void
 freeScratch()
 {
-  nrs_t * nrs = (nrs_t *)nrsPtr();
 
-  freePointer(nrs->usrwrk);
-  nrs->o_usrwrk.free();
+  freePointer(usrwrk);
```
aprilnovak, line 369: now that we're creating additional arrays to manage data NekRS used to, we'll also want to free them here/somewhere equivalent

src/base/NekInterface.C
```diff
@@ -513,33 +518,37 @@ limitTemperature(const double * min_T, const double * max_T)
   minimum = (minimum - scales.T_ref) / scales.dT_ref;
   maximum = (maximum - scales.T_ref) / scales.dT_ref;
 
-  nrs_t * nrs = (nrs_t *)nrsPtr();
   mesh_t * mesh = temperatureMesh();
 
+  const auto sid = nrs->scalar->nameToIndex.find("temperature")->second;
```
aprilnovak, line 577: out of curiosity, would the temperature scalar ID ever be different from 0, if temperature exists in the problem? If the scalars now have arbitrary ordering this will be important to know since it'll impact a lot of other places in the code where we'd need the same kind of index usage as here

src/base/NekRSProblem.C
```diff
@@ -939,11 +941,13 @@ NekRSProblem::writeVolumeDisplacement(const int elem_id,
                                       const std::vector<double> * add)
 {
   mesh_t * mesh = nekrs::entireMesh();
-  nrs_t * nrs = (nrs_t *)nekrs::nrsPtr();
+  auto nrs = nekrs::nrsPtr();
 
   auto vc = _nek_mesh->volumeCoupling();
   int id = vc.element[elem_id] * mesh->Np;
 
+  auto [x, y, z] = nekrs::host_xyz();
```
aprilnovak, line 956: do we need `host_xyz()`, or could we get away with just the `host_x`, `host_y`, and `host_z` functions?

src/postprocessors/NekPointValue.C
```diff
@@ -78,21 +78,21 @@ NekPointValue::execute()
     case field::velocity_x_squared:
     case field::velocity_y_squared:
     case field::velocity_z_squared:
-      n_values = n * nrs->NVfields;
+      n_values = n * 3; //nVFields
```
aprilnovak: did `NVfields` get deleted? not an issue, just asking for my understanding

aprilnovak:
Thanks @nandu90 , this is a great start. I think the overall design will be compatible with the rest of Cardinal - can you point me to where the host-side solution fields are being updated from the device-side solutions? I may have missed it or it hasn't been added yet to the draft PR (in which case we'll need to add it probably in `NekRSProblem::syncSolutions` so that any Cardinal objects which try to access U, P, S have up to date values.

src/base/NekRSProblem.C
```diff
@@ -493,10 +493,6 @@ NekRSProblem::externalSolve()
   // tell NekRS what the value of nrs->isOutputStep should be
   nekrs::checkpointStep(_is_output_step);
 
-  // NekRS prints out verbose info for the first 1000 time steps
```
aprilnovak, line 492: I feel like some of my commits perhaps got overwritten at some point with multiple people pushing/merging to Nadish's branch? I'm 99% sure I had pushed these same fixes at some point in the past. @nandu90 @kyleadm @maxidalinger @emerzari

aprilnovak:
closing, as this was resolved in #1340

PR #1260, opened 2025-12-12
Fix MOOSE language server crash

### LINKED ISSUE #1259: MOOSE language server crashes when an input deck contains an `OpenMCCellAverageProblem`
## Bug Description
The MOOSE language server crashes when parsing an input file containing an `OpenMCCellAverageProblem`. This is caused by the initialization of OpenMC on class construction, which prints the OpenMC header and corrupts the string output the language server relies on.

## Steps to Reproduce
Attempt to use the MOOSE input file syntax plugin (in any supported editor) to provide syntax highlighting in an input file containing an `OpenMCCellAverageProblem`. 

## Impact
Language server crashes, reducing productivity.

nuclearkevin:
This PR fixes the aforementioned crash by suppressing all OpenMC output when the CardinalApp is being executed by the MOOSE language server. I've tested this locally, and I get syntax highlighting/docstrings with the MOOSE syntax extension in VSCode when input files contain an `OpenMCCellAverageProblem`!

Closes #1259

### DIFF
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -124,7 +124,7 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
   // Suppress OpenMC output when the language server is active by
   // decreasing the verbosity to level 1 (the lowest).
   std::vector<std::string> argv_vec = {"openmc"};
-  if (_app.isParamValid("language_server") && _app.getParam<bool>("language_server")))
+  if (_app.isParamValid("language_server") && _app.getParam<bool>("language_server"))
   {
     argv_vec.push_back("-q");
     argv_vec.push_back("1");


src/base/OpenMCProblemBase.C
```diff
@@ -121,7 +121,16 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
     mooseError("The tally system used by OpenMCProblemBase derived classes has been deprecated. "
                "Please add tallies with the [Tallies] block instead.");
 
-  std::vector<std::string> argv_vec = {"openmc", _xml_directory};
+  // Suppress OpenMC output when the language server is active by
+  // decreasing the verbosity to level 1 (the lowest).
+  std::vector<std::string> argv_vec = {"openmc"};
+  if (_app.isParamValid("language_server"))
```
loganharbour:
```suggestion
  if (_app.isParamValid("language_server") && _app.getParam<bool>("language_server")))
```

I think that the way bool params work now is that they're always valid. So this is the safest check. I think that it might always be valid?

aprilnovak: out of curiosity, why would bool params always be valid? I though `isParamValid` only refers to a value having been set in an un-required parameter

nuclearkevin: Going through `MooseApp.C` it appears as if this is the case. I'll test this quickly, but I believe we need both checks. Might need to go through Cardinal's source code and update all of our `bool` parameters...

loganharbour: bool command line params are special. they can only ever have a default of false (because setting them sets them to true)

nuclearkevin: Yep, this is 100% required (all OpenMC output was getting suppressed regardless of the state of `language_server`). Thanks for catching this @loganharbour!

PR #1262, opened 2025-12-16
Re-order calls in setting OpenMC batches to prevent creating an empty set of  statepoint write batches

### LINKED ISSUE #1261: Cardinal loses statepoint files when Cardinal batch settings match OpenMC batch settings
## Bug Description
Cardinal will remove the desired statepoint batch from the set of statepoint batches when the XML settings exactly match the Cardinal OpenMC problem settings.

## Steps to Reproduce
Build an OpenMC XML model with a specific set of batch settings, i.e. `inactive_batches`, `batches`, and then specify identical settings in the OpenMC problem block of a Cardinal input file.

Reviewing the code:
* In Cardinal `OpenMCProblemBase::OpenMCProblemBase` at [line 230](https://github.com/neams-th-coe/cardinal/blob/73daf3541d7c10e77ce41c9320c59afcfd3c9f58/src/base/OpenMCProblemBase.C#L230), the value that was in the XML file is removed from the set of batches at which a statepoint will be produced.
* This comes immediately after the value from Cardinal is added to this set by a call to `openmc_set_n_batches` at [line 222](https://github.com/neams-th-coe/cardinal/blob/73daf3541d7c10e77ce41c9320c59afcfd3c9f58/src/base/OpenMCProblemBase.C#L222C15-L222C35)

If the value added at line 222 is the same as the value erased at line 230, then the policies of a `set` will result in that value being erased.

## Impact
Any workflows/problems that require access to the statepoint file are unable to be completed and that work cannot be accomplished.

lewisgross1296:
Closes #1261. Recently, my cardinal simulations were not writing statepoint files at all and we couldn't figure out why. It turns out that the order in which we erase/write OpenMC batches creates an empty set if your XML and `OpenMCCellAverageProblem` block have the same settings for `batches` since the `openmc::settings::statepoint_batch` object is a set. This should fix that issue and hopefully not break any other tests.

### DIFF
--- a/test/tests/neutronics/openmc_statepoint/tests
+++ b/test/tests/neutronics/openmc_statepoint/tests
@@ -1,6 +1,8 @@
 [Tests]
   # This test needs to be in its own directory to prevent a false positive if another test
   # writes a statepoint.50.h5 file
+  issues = '#1261'
+  design = 'OpenMCCelllAverageProblem.md'
   [check_statepoint_exists]
     type = CheckFiles
     input = openmc.i


test/tests/neutronics/openmc_xml/tests
```diff
@@ -41,6 +41,18 @@
                   "(with 60 batches) to ensure correctness."
     capabilities = 'openmc'
   []
+  [check_statepoint_exists]
+    type = CheckFiles
+    input = openmc.i
+    check_files = 'statepoint.50.h5'
+    cli_args = 'Problem/batches=50'
```
aprilnovak: Can we choose a different number? In this directory are some inputs which will already be creating a file named `statepoint.50.h5`, so based on the test execution order it's theoretically possible that one of the other tests makes a 50-batch statepoint but we fail to, although the test would still pass. The best fix would be to put your test in its own directory, but I think using a batch number unique to this input (within this directory) will also do the trick.

lewisgross1296: That's a good point. Is the execution order always first to last? I could do this test first. Otherwise, I need to update the XML to match whatever is in the `cli_args`. I think that is fine too as long as it isn't 60 for the `batches` test or 8 for the `skip_statepoint` test . I'm happy to put this in a diff directory too

aprilnovak: No, the tests run in parallel so there is no guaranteed execution order. True, I think the only way to guarantee this is to put it in its own directory.

gonuke (approved):
This looks good to me - thanks for the quick fix @lewisgross1296

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -0,0 +1,15 @@
+[Tests]
+  # This test needs to be in its own directory to prevent a false positive if another test
+  # writes a statepoint.50.h5 file
+  [check_statepoint_exists]
```
aprilnovak, line 6:
```suggestion
  issues = '#1261'
  design = 'OpenMCCelllAverageProblem.md'
  [check_statepoint_exists]
```

PR #1264, opened 2025-12-21
Move relaxation into `TallyBase` to improve maintainability.

### LINKED ISSUE #1263: Move relaxation to the tally system
## Reason
Tally normalization will be moved into `TallyBase` in #1106 to support multiple tallies which share the same score. With this change, we can also separate tally relaxation out and move it to `TallyBase`.

## Design
Move the contents of `OpenMCCellAverageProblem::relaxAndNormalizeTally` into `TallyBase::relaxAndNormalizeTally` and remove the function from `OpenMCCellAverageProblem`. 

## Impact
Essentially complete the separation of the tally system to improve maintainability of the tally system / `OpenMCCellAverageProblem`.

nuclearkevin:
Closes #1263

### DIFF
--- a/include/base/OpenMCCellAverageProblem.h
+++ b/include/base/OpenMCCellAverageProblem.h
@@ -663,28 +663,6 @@ protected:
   void storeElementPhase();
 
   /**
-   * Relax the tally and normalize it according to some "global" tally. If you set
-   * 'normalize_by_global_tally = true', you will be normalizing by a tally over the ENTIRE
-   * OpenMC geometry. Otherwise, you normalize only by the sum of the tally bins themselves.
-   *
-   * NOTE: This function relaxes the tally _distribution_, and not the actual magnitude of the sum.
-   * That is, we relax the shape distribution and then multiply it by the power
-   * (for k-eigenvalue) or source strength (for fixed source) of the current step before
-   * applying it to MOOSE. If the magnitude of the power or source strength is constant in time,
-   * there is zero error in this. But if the magnitude of the tally varies in time, we are basically
-   * relaxing the distribution of the tally, but then multiplying it by the _current_ mean tally
-   * magnitude.
-   *
-   * There will be very small errors in these approximations unless the power/source strength
-   * change dramatically with iteration. But because relaxation is itself a numerical approximation,
-   * this is still inconsequential at the end of the day as long as your problem has converged
-   * the relaxed tally to the raw (unrelaxed) tally.
-   * @param[in] local_score the local index of the tally score
-   * @param[in] local_tally the tally to relax and normalize
-   */
-  void relaxAndNormalizeTally(unsigned int local_score, std::shared_ptr<TallyBase> local_tally);
-
-  /**
    * Loop over all the OpenMC cells and count the number of MOOSE elements to which the cell
    * is mapped based on phase.
    */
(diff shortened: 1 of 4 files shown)


include/base/OpenMCCellAverageProblem.h
aprilnovak, line 792: should this also be removed?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2823,6 +2788,9 @@ OpenMCCellAverageProblem::addTally(const std::string & type,
   auto tally = addObject<TallyBase>(type, name, moose_object_pars, false)[0];
   _local_tallies.push_back(tally);
 
+  // Set the relaxation scheme.
+  tally->setRelaxation(_relaxation, getParam<Real>("relaxation_factor"));
```
aprilnovak, line 2792: I'm confused at to why this is here in the problem class - is it not possible for all the relaxation setting to be handled in the tally classes?

nuclearkevin: I might've been misinterpreted your [original thought on moving relaxation](https://github.com/neams-th-coe/cardinal/pull/1254#issuecomment-3662753177):
> Also, a stray thought (this does not have to be in this PR but am just posing it for consideration) - now that most of the tally functionality has been ported into the Tally system, do you think the relaxation also belongs there? I think the motivation would be purely from a code maintenance/accessibility standpoint (avoiding exposing tally stuff in the problem class), as I don't envision someone wanting to relax two different tallies with different strategies in practice. What do you think?

My interpretation was that we move the relaxation code while keeping the parameters themselves in `OpenMCCellAverageProblem` (since most users likely won't use different relaxation schemes per-tally). If that was erroneous, I'll move the input parameters to `TallyBase`.

nuclearkevin: Actually, now that I think about it Dufek-Gudowski relaxation makes it difficult to move the input parameters over to `OpenMCCellAverageProblem` as the number of particles per Picard iteration becomes variable.

include/base/OpenMCCellAverageProblem.h
nuclearkevin: `OpenMCCellAverageProblem` needs to know the relaxation type to change the number of particles run in a Picard iteration with Dufek-Gudowski, so I think this needs to stick around.

src/base/OpenMCCellAverageProblem.C
aprilnovak: my thinking was that all of the relaxation code could go into TallyBase, but you're right about the particle control aspect. Perhaps we could have a method like `int TallyBase::applyRelaxation()` which returns the number of particles to set back to `OpenMCCellAverageProblem` such that ALL the relaxation code could still be put in TallyBase?

nuclearkevin: That would definitely work, but my concern is how we handle the case where users mix Dufek-Gudowski with a different relaxation scheme? We could error if Dufek-Gudowski is selected in one tally and other tallies don't match, or we could override the other tallies and let the user know.

aprilnovak: true... maybe this idea isn't worth it and it should just stay in the problem class? I'm okay with whatever your opinion is on it (including your suggestion to check conflicting relaxation schemes)

nuclearkevin: After thinking about this some more I don't think moving the parameters over to the `TallyBase` is really justified when considering the extra user burden (needing to add relaxation to each tally), especially since there really aren't use cases for per-tally relaxation. I think the changes here strike a decent balance between maintainability (moving the parts of relaxation that directly impact `TallyBase` into that class), user convenience (only needing to specify the input parameters once), and avoiding extra code (since the problem needs to know about the relaxation type for Dufek-Gudowski).

PR #1276, opened 2026-01-17
Improve MeshTally block restriction

### LINKED ISSUE #1275: Avoid copying the mesh when block restriction is applied in a `MeshTally`
## Reason
When block restrictions are specified in a `MeshTally`, a copy of the mesh is created and passed to OpenMC for tallying. This results in substantial memory usage when the block restriction includes large subdomains of the mesh.

## Design
Block restriction must be specified in the point locators used by OpenMC for tallying, along with the equivalent bin->element and element-> bin mapping data structures. This was merged into OpenMC in https://github.com/openmc-dev/openmc/pull/3694. The Cardinal component of this implementation is quite simple - the `std::set` of blocks just needs to be passed to an `openmc::AdaptiveLibMesh`. 

## Impact
Large reduction in memory usage when providing block restrictions to a `MeshTally`. Not requiring a mesh copy also fixes a bug with on-the-fly element amalgamation that @magnoxemo found (extra element IDs were not being preserved during the copying process).

nuclearkevin:
FYI @magnoxemo 

Closes #1275

### DIFF
--- a/src/tallies/MeshTally.C
+++ b/src/tallies/MeshTally.C
@@ -143,7 +143,7 @@ MeshTally::spatialFilter()
     }
 
     openmc::model::meshes.emplace_back(std::make_unique<openmc::AdaptiveLibMesh>(
-      _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling(), _tally_blocks));
+        _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling(), _tally_blocks));
   }
   else
     openmc::model::meshes.emplace_back(


src/tallies/MeshTally.C
```diff
@@ -142,32 +142,8 @@ MeshTally::spatialFilter()
...
-    {
-      if (_is_adaptive)
-        openmc::model::meshes.emplace_back(std::make_unique<openmc::AdaptiveLibMesh>(
-            _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling()));
-      else
-        openmc::model::meshes.emplace_back(std::make_unique<openmc::LibMesh>(
-            _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling()));
-    }
+    openmc::model::meshes.emplace_back(std::make_unique<openmc::AdaptiveLibMesh>(
+        _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling(), _tally_blocks));
```
magnoxemo, line 146: Thanks Kevin. I just have a quick question. Are we gonna use `AdaptiveLibMesh` even if the problem isn't adaptive? 
```suggestion
    if (_is_adaptive)
       openmc::model::meshes.emplace_back(std::make_unique<openmc::AdaptiveLibMesh>( _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling(), _tally_blocks));
       
    else
       openmc::model::meshes.emplace_back(std::make_unique<openmc::LibMesh>(
        _openmc_problem.getMooseMesh().getMesh(), _openmc_problem.scaling(), _tally_blocks));
```

nuclearkevin: The infrastructure to perform block restriction is only available in `openmc::AdaptiveLibMesh` to insulated regular OpenMC users from any potential performance hits, so we're forced into it here (`openmc::LibMesh` doesn't have the constructor arguments to take `_tally_blocks`).

magnoxemo: Thanks. That answers my question.

src/tallies/MeshTally.C
```diff
@@ -142,32 +142,8 @@ MeshTally::spatialFilter()
...
-      _libmesh_mesh_copy->allow_find_neighbors(true);
-      _libmesh_mesh_copy->allow_renumbering(false);
-      _libmesh_mesh_copy->prepare_for_use();
-
-      openmc::model::meshes.emplace_back(
-          std::make_unique<openmc::LibMesh>(*_libmesh_mesh_copy.get(), _openmc_problem.scaling()));
-    }
-    else
-    {
-      if (_is_adaptive)
```
aprilnovak, line 164: is this change here independent of the PR? could we always have made only an `AdaptiveLibMesh`?

nuclearkevin: We could've always made an `AdaptiveLibMesh`, but there wasn't much of a point in doing so due to the slight performance penalty. Now that `AdaptiveLibMesh` supports block restriction I think it makes sense to use it over a standard `LibMesh`.

PR #1283, opened 2026-01-30
Allow assigning implicit complement material in the skinner

### LINKED ISSUE #1282: Allow assigning a material to the implicit complement with the MOAB Skinner
## Reason
Currently, a void material is always assigned to the implicit complement of the geometry when using the skinner. Users may want to assign different materials instead.

## Design
Add a new parameter in the MoabSkinner that takes an OpenMC material name and assigns it to the implicit complement of the geometry

## Impact
Enhance user experience

meltawila:
closes #1282

### DIFF
--- a/test/tests/neutronics/dagmc/implicit_comp_material/without_graveyard/tests
+++ b/test/tests/neutronics/dagmc/implicit_comp_material/without_graveyard/tests
@@ -6,7 +6,7 @@
     cli_args = 'Outputs/hide="cell_instance"'
     requirement = "The system shall assign a material to the implicit complement region when skinning "
                   " when there is no graveyard as compared to a CSG-equivalent version of the geometry"
-                  " which is in /csg_equivalent directory. The graveyard in this model exists as the "
+                  " which is in /csg_equivalent directory. The implicit complement region in this model exists as the "
                   "space between the mesh geometry and an enclosing CSG cell. Results agree within "
                   "statistics to the CSG reference."
     mesh_mode = 'replicated'


src/userobjects/MoabSkinner.C
```diff
@@ -80,9 +82,6 @@ MoabSkinner::validParams()
   params.addParam<bool>("use_displaced_mesh",
                         false,
                         "Whether the skinned mesh should be generated from a displaced mesh ");
-  params.addParam<bool>("use_displaced_mesh",
```
meltawila, line 83: Not related to this PR. I found that these lines are duplicate of the above

src/userobjects/MoabSkinner.C
```diff
@@ -111,6 +110,15 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
   _build_graveyard = getParam<bool>("build_graveyard");
   _use_displaced = getParam<bool>("use_displaced_mesh");
 
+  if (isParamValid("implicit_complement_material"))
+  {
+    _set_implicit_complement_material = true;
+    if (!_build_graveyard)
```
aprilnovak: is this strictly necessary? I think you can have a DAGMC geometry which, say, consists of a hollow sphere with vacuum condition on the sphere surface and with the implicit complement in the hollow region of the sphere.

meltawila: I think so, but I am not sure. DAGMC expects the material for the implicit complement to be specified as if it were being specified for the graveyard volume which is what I'm doing here (I need a graveyard volume to add to the comp_group)

https://svalinn.github.io/DAGMC/usersguide/codes/openmc.html#implicit-complement-materials

aprilnovak: true, but I think we just haven't had a use case yet where the skinner is being applied to a DAGMC model which had an implicit complement used elsewhere in addition to the graveyard. 

Can you walk me through what the test case is doing? Perhaps we can just modify that test so that it is, e.g., a hollow sphere and it has an implicit complement both for the graveyard and without (and then we can have two Cardinal tests, one with `build_graveyard = true` and the other not). This could get two birds with one stone and allow us to have an implicit complement separate from the graveyard volume.

meltawila: I removed this condition and added an arbitrary volume to the implicit complement material group. and added tests with both `build_graveyard = true` and  `build_graveyard = false`, but kept the slab geometry from 'test/tests/neutronics/dagmc' to reuse geometry files instead of the new hollow sphere

meltawila: The hollow sphere example you propose also requires that we support adding vacuum/reflective boundary conditions to the DAGMC geometry using the MoabSkinner. I think I can add this in a separate PR. For now, the test I added that uses `build_graveyard = false` adds the DAGMC geometry in a CSG cell which has the BCs

aprilnovak: yes, very true! I hadn't thought of that detail. Would you please open an issue to track the addition of vacuum/reflective boundaries for the skinner? Great idea to test by placing within a CSG cell!

src/userobjects/MoabSkinner.C
```diff
@@ -744,6 +750,31 @@ MoabSkinner::findSurfaces()
   if (_build_graveyard)
     buildGraveyard(vol_id, surf_id);
 
+  if (_set_implicit_complement_material)
+  {
+    moab::EntityHandle comp_group;
+    unsigned int comp_id = nBins() + 1;
```
```suggestion
    unsigned int comp_id = nBins() + 1 + _build_graveyard;
```

aprilnovak: to be fancy :)

src/userobjects/MoabSkinner.C
```diff
@@ -744,6 +750,31 @@ MoabSkinner::findSurfaces()
...
+    if (_build_graveyard)
```
```suggestion
```

src/userobjects/MoabSkinner.C
```diff
@@ -744,6 +750,31 @@ MoabSkinner::findSurfaces()
...
+      comp_id += 1;
```
```suggestion
```

test/tests/neutronics/dagmc/implicit_comp_material/tests
```diff
@@ -0,0 +1,13 @@
+[Tests]
+  [add_implicit_complement]
+    type = Exodiff
+    input = openmc.i
+    exodiff = openmc_out.e
+    cli_args = 'Outputs/hide="cell_instance"'
+    requirement = "The system shall assign a material to the implicit complement region when skinning "
```
aprilnovak, line 7: So just to check, are you doing a comparison of values of tallies between the skinned DAGMC model vs. an equivalent CSG model and thereby concluding this is implemented correctly?

src/userobjects/MoabSkinner.C
```diff
@@ -111,6 +110,13 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
   _build_graveyard = getParam<bool>("build_graveyard");
   _use_displaced = getParam<bool>("use_displaced_mesh");
 
+  if (isParamSetByUser("implicit_complement_material"))
+  {
+    _set_implicit_complement_material = true;
+    _implicit_complement_group_name =
```
aprilnovak, line 118: There should probably be an error check if the material specified by the user doesn't exist in the OpenMC model.

Also, out of curiosity what happens if the user provides`implicit_complement_material = None`, like to set a void? Does that work?

src/userobjects/MoabSkinner.C
```diff
@@ -744,6 +750,31 @@ MoabSkinner::findSurfaces()
...
+    {
+      const auto & vols = surf_pair.second;
+      if (!vols.empty())
+      {
+        arbitray_volume = vols.back().vol;
+        break;
+      }
+      else
+      {
+        mooseError("No volumes in the geometry found to assign to the implicit complement material "
```
aprilnovak: Should this be covered by a `RunException` test?

test/tests/neutronics/dagmc/implicit_comp_material/tests
meltawila: Yes, I followed the same logic in the rest of the tests inside test/tests/neutronics/dagmc - running locally with 2e6 particles, I get 1.33698 +/- 0.00008 for the skinner model and 1.33689 +/- 0.00009 for the CSG equivalent in the final time step. As for the `build_graveyard = false` in /without_graveyard, this is numerically identical compared to the base case with the graveyard with any number of particles, since they use the exact geometry

src/userobjects/MoabSkinner.C
meltawila: ~~I added an error check with an unknown material (will push shortly).~~ This error is being caught on the OpenMC side. `implicit_complement_material = Void` will work fine as expected, but `implicit_complement_material = None` will error similar to unrecognized material name

src/userobjects/MoabSkinner.C
meltawila: I realized now that I cannot get to this error. ~Cardinal~ OpenMC will fail first with `ERROR: No cells were found in the geometry.xml file` if there were no volumes in the geometry. I will remove this check

src/userobjects/MoabSkinner.C
```diff
@@ -62,7 +62,9 @@ MoabSkinner::validParams()
   params.addParam<Real>("graveyard_scale_outer",
                         1.10,
                         "Multiplier on mesh bounding box to form outer graveyard surface");
-
+  params.addParam<std::string>("implicit_complement_material",
```
aprilnovak, line 65: how does this actually work if the user provides either the name or ID? I tested it with your case, and it seems to work, but I'm not really sure why. What if my name for a material was a number?

test/tests/neutronics/dagmc/implicit_comp_material/openmc.i
```diff
@@ -0,0 +1,86 @@
+[Mesh]
+  type = FileMesh
+  file = ../mesh_tallies/slab.e
+  allow_renumbering = false
```
aprilnovak, line 4: In the initial DAGMC model, is the implicit complement set to void? So that the first time it runs via Cardinal, it is re-skinned so that on the first run the model has the water implicit complement? 

If so, why run 2 time steps in the test, why not just 1?

src/userobjects/MoabSkinner.C
meltawila: This is completely handled by DAGMC. Because both `group "mat:7" add vol 4` and `group "mat:fuel" add vol 4`  are allowed, it works here too. There's a note that material names must be unique in the materials.xml file for this to work properly
https://svalinn.github.io/DAGMC/usersguide/codes/openmc.html

test/tests/neutronics/dagmc/implicit_comp_material/openmc.i
meltawila: The initial DAGMC model is set to void, but it is skinned before the first OpenMC run so both time steps have water implicit complement. I took the test from `test/tests/neutronics/dagmc` which also had two time steps and I kept it as is. The only thing I updated in the test was assigning the implicit complement. One thing I noticed when comparing with /csg_equivalent was that the difference was a little over 1 sigma in the first time step but was within 1 sigma in the second time step for the number of particles I ran with

aprilnovak: let's see if we can drop down to just 1 time step to save on the test time.

For the implicit complement region, I'd suggest we make the implicit complement a bigger part of the domain - you can imagine that if the implicit complement is a tiny part of the domain, then it doesn't really matter if it's filled by water vs. vacuum vs. helium, because it'll have a small effect on the tallies (and could be within uncertainty). We want to be able to "see" easily if something breaks in your code and prove without doubt that it works. I tested this locally and made the graveyard bigger (there's graveyard_scale_inner/outer parameters in Cardinal that can let this be placed a greater distance away). This might help with the difference being so small it's on the order of sigma

aprilnovak: I modified the tests to CSV type and make the graveyard bigger to exaggerate the difference between void graveyward

test/tests/neutronics/dagmc/implicit_comp_material/without_graveyard/tests
```diff
@@ -0,0 +1,15 @@
+[Tests]
+  [no_graveyard_implicit_complement]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = openmc_out.csv
+    cli_args = 'Outputs/hide="cell_instance"'
+    requirement = "The system shall assign a material to the implicit complement region when skinning "
+                  " when there is no graveyard as compared to a CSG-equivalent version of the geometry"
+                  " which is in /csg_equivalent directory. The graveyard in this model exists as the "
```
meltawila:
```suggestion
                  " which is in /csg_equivalent directory. The implicit complement region in this model exists as the "
```

PR #1286, opened 2026-02-04
Add vacuum and reflective BCs using the MoabSkinner

### LINKED ISSUE #1285: Assign DAGMC boundary conditions using MoabSkinner
## Reason
Currently to apply reflective boundary conditions on a DAGMC geometry, we rely on filling the DAGMC universe inside a CSG cell and applying boundary conditions on the CSG surfaces. This can be simplified by applying the BCs on the DAGMC surfaces directly

## Design
Add `vacuum_bcs_surfaces` and `reflective_bcs_surfaces` parameters to `MoabSkinner`, which takes MOOSE sideset names/ids and creates a group that assigns corresponding BCs while generating the DAGMC geometry

## Impact
enhance user experience

meltawila:
closes #1285


src/userobjects/MoabSkinner.C
```diff
@@ -82,6 +82,10 @@ MoabSkinner::validParams()
   params.addParam<bool>("use_displaced_mesh",
                         false,
                         "Whether the skinned mesh should be generated from a displaced mesh ");
+  params.addParam<std::vector<BoundaryName>>("vacuum_bcs_surfaces",
+                                             "Sideset names/ids to assign vacuum BCs to");
+  params.addParam<std::vector<BoundaryName>>("reflective_bcs_surfaces",
+                                             "Sideset names/ids to assign reflective BCs to");
```
aprilnovak: Question: does the code work with either names or IDs?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
   // Write MOAB volume and/or skin meshes to file
   write();
 }
 
+void
+MoabSkinner::addBoundaryConditionGroups()
+{
+  auto namesToSidesetIDs =
+      [&](const std::vector<BoundaryName> & names, const std::string & param_name)
+  {
```
aprilnovak: this inline function seems more complicated than it needs to be? it does not use the `param_name` input. Since you just call this right below, I'd suggest to get rid of this inline thing to simplify the code.

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+    std::set<BoundaryID> sidesetIDs;
+    for (const auto & n : names)
+    {
+      const BoundaryID id = getMooseMesh().getBoundaryID(n);
+      if (!getMooseMesh().meshSidesetIds().count(id))
+        mooseError("Boundary '", n, "' is not a sideset.");
```
aprilnovak: Does this work properly if user provides IDs directly in the `.i` file?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+    return sidesetIDs;
+  };
+
+  const auto vac_ids = namesToSidesetIDs(_vacuum_bcs_surfaces, "vacuum_bcs_surfaces");
+  const auto ref_ids = namesToSidesetIDs(_reflective_bcs_surfaces, "reflective_bcs_surfaces");
+
+  for (auto id : vac_ids)
+    if (ref_ids.count(id))
+      mooseError(
+          "Same sideset ", id, " appears in both vacuum_bcs_surfaces and reflective_bcs_surfaces.");
```
aprilnovak: Is this covered by a test?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+  moab::Range vac_tris, ref_tris;
+
+  const auto & boundary_info = getMooseMesh().getMesh().get_boundary_info();
+
+  for (const auto * elem : getMooseMesh().getMesh().active_element_ptr_range())
+  {
+    auto elem_to_moab_tets_it = _id_to_elem_handles.find(elem->id());
+    if (elem_to_moab_tets_it == _id_to_elem_handles.end())
```
aprilnovak: does this ever get called or is this dead code?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+      std::unique_ptr<const Elem> side_elem = elem->build_side_ptr(s);
+      std::set<moab::EntityHandle> side_verts;
+      for (unsigned int i = 0; i < side_elem->n_nodes(); ++i)
+        side_verts.insert(_node_id_to_handle[side_elem->node_id(i)]);
+
+      for (const auto tet : elem_to_moab_tets_it->second)
+      {
+        moab::Range tri_faces;
+        check(_moab->get_adjacencies(&tet, 1, 2, true, tri_faces));
```
aprilnovak: what is this doing? can you add a comment here?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+        for (const auto tri : tri_faces)
+        {
+          const moab::EntityHandle * conn = nullptr;
+          int nconn = 0;
+          check(_moab->get_connectivity(tri, conn, nconn));
+          if (nconn != 3)
```
aprilnovak: can you explain this 3 here? please add a comment - I'm assuming this has to do with the elements being tets, but what does this connectivity represent?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+              vac_tris.insert(tri);
+            if (want_ref)
+              ref_tris.insert(tri);
+          }
+        }
+      }
+    }
+  }
+
+  if (vac_tris.empty() && ref_tris.empty())
```
aprilnovak: would it indicate some kind of user error if these were empty at this point?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+    return;
+
+  unsigned int next_surface_id = 1;
+  {
```
aprilnovak: why is this inside brace scope?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+        moab::intersect(triangles_in_surface_meshset, vac_tris);
+
+    const moab::Range reflective_triangles_on_surface =
+        moab::intersect(triangles_in_surface_meshset, ref_tris);
+
+    if (vacuum_triangles_on_surface.empty() && reflective_triangles_on_surface.empty())
+      continue;
+
+    // No BC specified (defaults to transmission)
+    moab::Range transmission_triangles_on_surface = triangles_in_surface_meshset;
```
aprilnovak: can you help me understand these lines? how would there be triangles on the surface indicated as vacuum but which are transmission? aren't we assigning the BCs ourselves, so where does the notion of transmission enter into the picture?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+    // No splitting needed unless:
+    //  - mixed vacuum+reflective, OR
+    //  - any transmission remainder exists (partial BC assignment)
+    if (!has_transmission && !(has_vacuum && has_reflective))
+      continue;
+
+    const std::vector<VolData> & surface_volume_data = surfsToVols[moab_surface_meshset];
+
+    enum Keep
```
aprilnovak: this seems random - why not just use booleans? In line 936, you're assinging a variable `keep` but it's a simple assignment which seems like it could just as easily be accomplished with the `has_transmission` boolean.

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+    if (has_reflective)
+    {
+      moab::EntityHandle new_surface = 0;
+      moab::Range surface_tris = reflective_triangles_on_surface;
+      createSurf(next_surface_id++, new_surface, surface_tris, surface_volume_data);
+    }
+  }
+
+  unsigned int gid = nBins() + 2;
```
aprilnovak: can you please explain the `2` here?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+  if (_build_graveyard)
+    gid += 1;
+  moab::EntityHandle vac_group = 0, ref_group = 0;
+  if (!vac_tris.empty())
+    createGroup(gid++, "boundary:vacuum", vac_group);
+  if (!ref_tris.empty())
+    createGroup(gid++, "boundary:Reflecting", ref_group);
```
aprilnovak: is it the convention that R is capitalized here but the v is not for vacuum?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+  for (const auto & surf_pair : surfsToVols)
+  {
+    const moab::EntityHandle surf_set = surf_pair.first;
+    moab::Range tris;
+    check(_moab->get_entities_by_handle(surf_set, tris));
```
aprilnovak: what is this `check` checking?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+    const bool has_vacuum_tris = (!vac_tris.empty() && !moab::intersect(tris, vac_tris).empty());
+    const bool has_reflective_tris =
+        (!ref_tris.empty() && !moab::intersect(tris, ref_tris).empty());
+
+    if (has_vacuum_tris)
+      check(_moab->add_entities(vac_group, &surf_set, 1));
```
aprilnovak: what is this `check` checking?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+    if (has_reflective_tris)
+      check(_moab->add_entities(ref_group, &surf_set, 1));
```
aprilnovak: what is this `check` checking?

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,61 @@
...
+                  "the DAGMC geometry in /equivalent_bcs."
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+  []
+  [all_vacuum]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = 'all_vacuum.csv'
+    cli_args = 'UserObjects/moab/vacuum_bcs_surfaces="top bottom front back left right" Outputs/file/file_base="all_vacuum"'
+    requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when adding "
```
aprilnovak: the `requirement` needs to be unique for each test. This is identical to the `all_reflective` test.

aprilnovak: and several other tests here, it looks like

test/tests/neutronics/dagmc/add_bcs/typo.i
```diff
@@ -0,0 +1,114 @@
...
+  #   old_boundary = '2 4'
+  #   new_boundary = 'reflective_boundary reflective_boundary'
+  # []
+  [hextotet]
+    type = ElementsToTetrahedronsConverter
+    input = name
+  []
+
+  allow_renumbering = false
+  parallel_type = replicated
```
aprilnovak: Since you already have the restriction to replicated mesh in the `tests` file, you don't need to include it here

aprilnovak (changes requested):

src/userobjects/MoabSkinner.C
gonuke: Don't you want to test this sooner, ie. in the constructor?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+      sidesetIDs.insert(id);
+    }
+    return sidesetIDs;
+  };
+
+  const auto vac_ids = namesToSidesetIDs(_vacuum_bcs_surfaces, "vacuum_bcs_surfaces");
+  const auto ref_ids = namesToSidesetIDs(_reflective_bcs_surfaces, "reflective_bcs_surfaces");
+
+  for (auto id : vac_ids)
+    if (ref_ids.count(id))
```
gonuke: Is there a simple way to use set logic to get all the overlapping ids at once and issue a single error with them all?

src/userobjects/MoabSkinner.C
gonuke: Could also replace those magic numbers (1,2) with a variable that clarifies their purpose.

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
   }
+  if (_set_bcs)
+    addBoundaryConditionGroups();
 
   // Write MOAB volume and/or skin meshes to file
   write();
 }
 
+void
+MoabSkinner::addBoundaryConditionGroups()
```
gonuke: I wonder if this would be simpler if the method only handled one of the boundary conditions at a time, and was called once for each boundary conditions (assuming that the user has requested it)

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+  for (auto id : vac_ids)
+    if (ref_ids.count(id))
+      mooseError(
+          "Same sideset ", id, " appears in both vacuum_bcs_surfaces and reflective_bcs_surfaces.");
+
+  moab::Range vac_tris, ref_tris;
+
+  const auto & boundary_info = getMooseMesh().getMesh().get_boundary_info();
+
+  for (const auto * elem : getMooseMesh().getMesh().active_element_ptr_range())
```
gonuke: (Caveat: I'm not deeply familiar with this particular implementation of skinning, but am very familiar with the concept.) This whole loop appears to be almost like it's performing a skinning operation again.  Could some of this information have been captured more easily during the process of skinning the mesh rather than going back after the fact to query the mesh again?

src/userobjects/MoabSkinner.C
```diff
@@ -764,11 +778,216 @@ MoabSkinner::findSurfaces()
...
+
+  unsigned int next_surface_id = 1;
+  {
+    unsigned int max_surface_id = 0;
+    for (const auto & surface_entry : surfsToVols)
+    {
+      const moab::EntityHandle moab_surface_meshset = surface_entry.first;
+      unsigned int surface_id = 0;
+      check(_moab->tag_get_data(id_tag, &moab_surface_meshset, 1, &surface_id));
+      max_surface_id = std::max(max_surface_id, surface_id);
```
gonuke: I wonder if it's more efficient to load all the surface meshset handles into a vector and get all the IDs in a single call to `tag_get_data`?

gonuke (changes requested):
A few questions/comments based on a first pass through this.  I'll try to take a deeper look soon.

Also, while I'm not fully versed in the style conventions of this project, it seems like the function to add the boundary conditions would be more readable if broken into a number of smaller functions each with a clear semantic purpose.

meltawila:
Sorry for being so late @aprilnovak @gonuke! I updated this PR and I think I addressed all your comments. The biggest change was mainly to address prof. Wilson's concern that the previous code was almost performing a skinning operation again to sort triangles into their corresponding sidesets. Now instead `_elem_handle_to_id `is populated during `createMOABElems()` which sorts skin triangles into vacuum/reflective/transmission sets directly inside `findSurface()`. Also `addBoundaryConditionGroups()` has also been broken into four smaller functions as suggested. I will respond to each review comment shortly

src/userobjects/MoabSkinner.C
meltawila: yes it does. updated parameter description to say this

src/userobjects/MoabSkinner.C
meltawila: Removed

src/userobjects/MoabSkinner.C
meltawila: Yes boundary names or IDs are accepted in the input file. I will add another test that does this

src/userobjects/MoabSkinner.C
meltawila: yes. test name is `sideset_added_twice`

src/userobjects/MoabSkinner.C
meltawila: This part is now removed, and the sorting happens inside `splitSkinByBoundaryCondition()`

src/userobjects/MoabSkinner.C
meltawila: This part is now moved into `splitSkinByBoundaryCondition()` and replaced (1,2) with `num_entities = 1` (one tet passed per call) and `surface_dimension = 2 ` (requesting 2D adjacencies) (https://www.mcs.anl.gov/~fathom/moab-docs/html/classmoab_1_1Interface.html#a2cb0794670be30e983b8ab2660e338fe). 

In the updated splitSkinByBoundaryCondition(), the logic is:
- add all skinned triangles into `transmission_tris`
- loop over each tet in the current region whose skin was produced by `find_skin()`
- for each tet look up the corresponding libmesh element and for each of its sides determine the BC type from the side's sideset membership in `_vacuum_bcs_surface_ids` or `_reflective_bcs_surface_ids`
- if a requested BC was found on a side in the previous step, get the tet's adjacent MOAB triangles and find the one lying on that side: a triangle matches if it's in the current skin and all 3 of its vertices are contained in the side's vertex handles - necessary because get_adjacencies returns triangles with no side association (e.g. an element at a corner touches multiple sidesets, and each triangle must be tied to its specific side)
- if the triangle matches move it from `transmission_tris` into `vacuum_tris` or `reflective_tris`

src/userobjects/MoabSkinner.C
meltawila: It is actually `vertices_per_triangle = 3`. this check would have been needed if I used a higher dimension in _moab->get_adjacencies() but it is actually unnecessary. will remove this now

src/userobjects/MoabSkinner.C
meltawila: yes, a mooseError now checks this in createBoundaryConditionGroups()

src/userobjects/MoabSkinner.C
meltawila: removed this entire next_surface_id scan

src/userobjects/MoabSkinner.C
meltawila: I added a comment explaining the logic above but mainly all the skinned triangles are labeled as transmission by default, then I loop over them and check if they correspond to a requested libMesh sideset that should be vacuum or reflective and change the label accordingly. The reason we can't just directly assign BC types without this checking loop is that there is no direct correspondence between MOAB skin triangles and libMesh sideset faces so we have to go through each tet, find its adjacent triangles and match them to libMesh sides to figure out which skin triangles belong to which sideset

src/userobjects/MoabSkinner.C
meltawila: This is now removed. splitSkinByBoundaryCondition() now sorts skins into transmission, vaccum and reflective upfront

src/userobjects/MoabSkinner.C
meltawila: replaced with `firstBoundaryConditionGroupID()` the original +2 was to make sure this group number is not used by material groups, graveyard group if it was enabled. implicit complement material group if it was requested

src/userobjects/MoabSkinner.C
meltawila: not sure about the convention - but both upper and lower case works fine as well as Reflective vs Reflecting?

src/userobjects/MoabSkinner.C
meltawila: the check() wrapping was already being used on every _moab-> call before this PR and I kept it consistent

meltawila: The `_moab->get_entities_by_handle(surf_set, tris)` part is populating `tris` with the triangles that belong to the surface meshset

src/userobjects/MoabSkinner.C
meltawila: the check() wrapping was already being used on every _moab-> call before this PR and I kept it consistent. this specific part is now removed

src/userobjects/MoabSkinner.C
meltawila: same as above

test/tests/neutronics/dagmc/add_bcs/tests
meltawila: updated these

test/tests/neutronics/dagmc/add_bcs/typo.i
meltawila: updated

src/userobjects/MoabSkinner.C
meltawila: I added a check for this in the constructor

src/userobjects/MoabSkinner.C
meltawila: I added `checkBoundaryConditionOverlap()` and used `std::set_intersection()` to issue a single error with them all

src/userobjects/MoabSkinner.C
meltawila: In the updated code after breaking this function into smaller ones, `createSurfacesFromSkin()` calls `createSurfaces()` 3 times (for transmission, vacuum, and reflective groups) to address this

src/userobjects/MoabSkinner.C
meltawila: I removed this and replaced it with `_elem_handle_to_id` (a reverse map from MOAB tet handle to libMesh element ID) in `createMOABElems()`. This allows `splitSkinByBoundaryCondition()` to iterate only over the tets in the current region then look up their libMesh elements and classify the resulting triangles to their corresponding requested BCs. I believe this now avoids the "preforming another skinning operation" concern since it runs while the skin result is in hand and only iterates over the tets in the current region not the full mesh

src/userobjects/MoabSkinner.C
meltawila: This part was removed as these are now recorded directly during the skinning

include/userobjects/MoabSkinner.h
```diff
@@ -38,7 +38,9 @@ class MoabSkinner : public GeneralUserObject
    */
   virtual moab::ErrorCode check(const moab::ErrorCode input) const;
 
-  std::string materialName(const unsigned int & block, const unsigned int & density, const unsigned int & temp) const;
+  std::string materialName(const unsigned int & block,
```
aprilnovak, line 41: note to future selves: I think we don't need to create new materials anymore to apply density feedback, ever since @nuclearkevin added density multipliers in OpenMC. We could update the skinner to use this in the future (would save memory and improve code simplicity in Cardinal) - not needed here in this PR to keep things focused.

include/userobjects/MoabSkinner.h
```diff
@@ -353,6 +370,54 @@ class MoabSkinner : public GeneralUserObject
...
+  void checkBoundaryConditionOverlap() const;
+
+  /// Determine which BC type, if any, applies to a libMesh side based on its boundary IDs.
+  BoundaryConditionType
+  boundaryConditionType(const std::vector<boundary_id_type> & side_boundary_ids) const;
+
+  /**
+   * Classify the skinned triangles for one region into transmission, vacuum, and
+   * reflective sets by checking each triangle against the BC sidesets.
+   */
```
aprilnovak, line 425: for all of these new function definitions, please add the full doxygen documentation which has the `param[in]`, `param[out]`, and `return` flags. You can check other include files for examples

src/userobjects/MoabSkinner.C
```diff
@@ -82,6 +82,16 @@ MoabSkinner::validParams()
   params.addParam<bool>("use_displaced_mesh",
                         false,
                         "Whether the skinned mesh should be generated from a displaced mesh ");
+  params.addParam<std::vector<BoundaryName>>(
+      "vacuum_bcs_surfaces",
+      "Mesh sideset names or numeric sideset IDs to assign DAGMC vacuum boundary conditions to. "
+      "Both string names and integer IDs (as strings) are accepted. "
```
aprilnovak: what does the "(as strings)" specifically mean? Can't the user just do `vacuum_bcs_surfaces = "this that 4"` in the input file? If that's the intended syntax, I'd just remove the "(as strings)" thing because IMO it's kind of confusing

src/userobjects/MoabSkinner.C
```diff
@@ -82,6 +82,16 @@ MoabSkinner::validParams()
...
+      "Sides not listed here default to transmission.");
```
```suggestion
      "If not specified, the surface defaults to transmission.");
```

src/userobjects/MoabSkinner.C
```diff
@@ -82,6 +82,16 @@ MoabSkinner::validParams()
...
+  params.addParam<std::vector<BoundaryName>>(
+      "reflective_bcs_surfaces",
+      "Mesh sideset names or numeric sideset IDs to assign DAGMC reflective boundary conditions "
+      "to. Both string names and integer IDs (as strings) are accepted. "
+      "Sides not listed here default to transmission.");
```
```suggestion
      "If not specified, the surface defaults to transmission.");
```

src/userobjects/MoabSkinner.C
```diff
@@ -82,6 +82,16 @@ MoabSkinner::validParams()
...
                         "Whether the skinned mesh should be generated from a displaced mesh ");
+  params.addParam<std::vector<BoundaryName>>(
+      "vacuum_bcs_surfaces",
+      "Mesh sideset names or numeric sideset IDs to assign DAGMC vacuum boundary conditions to. "
+      "Both string names and integer IDs (as strings) are accepted. "
+      "Sides not listed here default to transmission.");
+  params.addParam<std::vector<BoundaryName>>(
+      "reflective_bcs_surfaces",
+      "Mesh sideset names or numeric sideset IDs to assign DAGMC reflective boundary conditions "
+      "to. Both string names and integer IDs (as strings) are accepted. "
```
aprilnovak: same comment here about "(as strings)"

src/userobjects/MoabSkinner.C
```diff
@@ -118,6 +128,28 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
     _implicit_complement_group_name =
         "mat:" + getParam<std::string>("implicit_complement_material") + "_comp";
   }
+  if (isParamSetByUser("vacuum_bcs_surfaces"))
+    _set_bcs = true;
+
+  if (isParamSetByUser("reflective_bcs_surfaces"))
+    _set_bcs = true;
```
aprilnovak: this is a good one to move to the member initialization list.

```
_set_bcs(isParamSetByUser("vacuum_bcs_surfaces") || isParamSetByUser("reflective_bcs_surfaces"))
```

src/userobjects/MoabSkinner.C
```diff
@@ -118,6 +128,28 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
...
+
+  if (_set_bcs)
+  {
+    const auto vac_names = isParamSetByUser("vacuum_bcs_surfaces")
```
aprilnovak: This code block seems like it could be combined with the other `if (_set_bcs)` piece at line 313.

src/userobjects/MoabSkinner.C
```diff
@@ -118,6 +128,28 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
...
+                               ? getParam<std::vector<BoundaryName>>("reflective_bcs_surfaces")
+                               : std::vector<BoundaryName>{};
+
+    std::set<BoundaryName> vac_name_set(vac_names.begin(), vac_names.end());
+    for (const auto & name : ref_names)
+      if (vac_name_set.count(name))
+        mooseError("Boundary '",
+                   name,
+                   "' appears in both 'vacuum_bcs_surfaces' and 'reflective_bcs_surfaces'.");
+  }
```
aprilnovak: do we need to check for duplicates in the `vacuum_bcs_surfaces` and `reflective_bcs_surfaces` themselves? What happens if user does `vacuum_bcs_surfaces = '1 2 3 2'`?

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+std::set<BoundaryID>
+MoabSkinner::boundaryNamesToIDs(const std::vector<BoundaryName> & names,
+                                const std::string & param_name)
+{
+  std::set<BoundaryID> ids;
+  for (const auto & name : names)
+  {
+    const BoundaryID id = getMooseMesh().getBoundaryID(name);
+    if (!getMooseMesh().meshSidesetIds().count(id))
+      paramError(param_name, "Boundary '", name, "' is not a sideset.");
```
aprilnovak: Is this the most accurate error message? If we get to this part of the code, I would have thought the error message would be "Boundary 4 does not exist", not that it's not a SIDESET?

aprilnovak: someone could get confused because there are concepts called nodesets which do refer to boundary data

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+{
+  bool want_vacuum = false;
+  bool want_reflective = false;
+  for (const auto bid : side_boundary_ids)
+  {
+    want_vacuum = want_vacuum || _vacuum_bcs_surface_ids.count(bid);
+    want_reflective = want_reflective || _reflective_bcs_surface_ids.count(bid);
+  }
+  if (want_vacuum && want_reflective)
+    mooseError("A mesh side is assigned both vacuum and reflective boundary conditions. "
```
aprilnovak: also confused on how we get to here? wouldn't this also already be caught in the `checkBoundaryConditionsOverlap` function?

aprilnovak: In general, all mooseErrors should be tested - if we can't get to an error message, then it's just dead code

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+               "Check 'vacuum_bcs_surfaces' and 'reflective_bcs_surfaces'.");
+  if (want_vacuum)
+    return BoundaryConditionType::Vacuum;
+  if (want_reflective)
+    return BoundaryConditionType::Reflective;
+  return BoundaryConditionType::None;
```
aprilnovak: for this enumeration, treat it with a switch-case statement so that there is a default. This helps protect against future scenarios where we add a third boundary condition (e.g., say, an albedo BC) but forget to update this code here.

aprilnovak: you can grep for `default:` in Cardinal and find examples

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+  createSurfaces(transmission_tris, voldata, surf_id, BoundaryConditionType::None);
+  createSurfaces(vacuum_tris, voldata, surf_id, BoundaryConditionType::Vacuum);
+  createSurfaces(reflective_tris, voldata, surf_id, BoundaryConditionType::Reflective);
+}
+
+void
+MoabSkinner::recordBoundaryConditionSurface(moab::EntityHandle surface_set,
+                                            BoundaryConditionType bc_type)
+{
+  if (bc_type == BoundaryConditionType::None)
```
aprilnovak: use switch-case in this function and anywhere else in the code that you're using these enumeration values

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+    return;
+
+  if (bc_type == BoundaryConditionType::Vacuum)
+  {
+    if (_reflective_bc_surface_sets.count(surface_set))
+      mooseError("A DAGMC surface was assigned both vacuum and reflective boundary conditions.");
```
aprilnovak: do we ever get to this error? this is fourth time I'm seeing a similar error check, I think it's dead code?

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+  }
+}
+
+unsigned int
+MoabSkinner::firstBoundaryConditionGroupID() const
+{
+  // IDs 1..nBins() are used by material groups.
+  // buildGraveyard() and the implicit complement each consume one additional ID when enabled.
+  unsigned int gid = nBins() + 1;
+  if (_build_graveyard)
```
aprilnovak: `gid += _build_graveyard + _set_implicit_complement_material`

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+}
+
+void
+MoabSkinner::createBoundaryConditionGroups()
+{
+  if (!_set_bcs)
+    return;
+
+  if (!_vacuum_bcs_surface_ids.empty() && _vacuum_bc_surface_sets.empty())
+    mooseError("'vacuum_bcs_surfaces' was specified but no skinned DAGMC surfaces were assigned "
```
aprilnovak: do we ever get here? add test for this

include/userobjects/MoabSkinner.h
```diff
@@ -379,12 +444,33 @@ class MoabSkinner : public GeneralUserObject
   /// Map from libmesh id to MOAB element entity handles
   std::map<dof_id_type, std::vector<moab::EntityHandle>> _id_to_elem_handles;
 
+  /// Map from libMesh id to MOAB vertex handles
+  std::map<dof_id_type, moab::EntityHandle> _node_id_to_handle;
```
GiudGiud:
```suggestion
  std::unordered_map<dof_id_type, moab::EntityHandle> _node_id_to_handle;
```
scales better

GiudGiud: unless you need to use this for an ordered range?

src/userobjects/MoabSkinner.C
GiudGiud: right when you are about to error, look at all nodesets to see if it exists as a nodeset and give a more useful message then

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+                        _reflective_bcs_surface_ids.end(),
+                        std::inserter(overlap, overlap.begin()));
+  if (overlap.empty())
+    return;
+
+  std::string ids_str;
+  for (const auto id : overlap)
+    ids_str += (ids_str.empty() ? "" : ", ") + std::to_string(id);
+
+  mooseError("The following sideset ID(s) appear in both 'vacuum_bcs_surfaces' and "
```
GiudGiud: could be a param error

src/userobjects/MoabSkinner.C
GiudGiud: you could get here if an element side is part of two sidesets, one having a reflective BC and one having a vacuum bc

GiudGiud: I'd recommend using a mooseAssert if you can't get to the error message

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,73 @@
...
+                  "boundary conditions through the skinner when all external surfaces are assigned vacuum boundary conditions "
+                  "via 'vacuum_bcs_surfaces'.as compared to applying BCs on a CSG cell that contains"
```
GiudGiud:
```suggestion
                  "via 'vacuum_bcs_surfaces', as compared to applying BCs on a CSG cell that contains"
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,73 @@
...
+                  "the DAGMC geometry in /equivalent_bcs."
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+  []
+  [mix_bcs]
+      type = CSVDiff
+      input = openmc.i
+      csvdiff = 'mix_bcs.csv'
+      cli_args = 'UserObjects/moab/vacuum_bcs_surfaces="left right" UserObjects/moab/reflective_bcs_surfaces="front back top bottom" Outputs/file/file_base="mix_bcs"'
```
GiudGiud:
```suggestion
      cli_args = "UserObjects/moab/vacuum_bcs_surfaces='left right'
                         UserObjects/moab/reflective_bcs_surfaces='front back top bottom'
                         Outputs/file/file_base='mix_bcs'"
```
formatting like this is easier to read

test/tests/neutronics/dagmc/add_bcs/openmc.i
```diff
@@ -0,0 +1,103 @@
+[Mesh]
+[f]
```
GiudGiud:
```suggestion
  [f]
```

test/tests/neutronics/dagmc/add_bcs/conflicting_bcs.i
```diff
@@ -0,0 +1,114 @@
+[Mesh]
+[f]
```
GiudGiud:
```suggestion
  [f]
```

test/tests/neutronics/dagmc/add_bcs/gold/transmission_surface.csv
```diff
@@ -0,0 +1,3 @@
+time,k
+0,0
+1,0.39439900002641
```
GiudGiud: why does this match the mix_bcs case that has a vacuum BC on the right instead of a transmission BC?

are you really testing transmission if the case is effectively vacuum

GiudGiud: I think you need to test a case with an internal surface

src/userobjects/MoabSkinner.C
```diff
@@ -911,17 +1192,20 @@ MoabSkinner::findSurface(const moab::Range & region,
...
   skinner->find_skin(0, region, false, tris, &rtris);
 
-  // Create surface sets for the forwards tris
+  // Classify and create surfaces. BC sorting happens here, while the current
+  // region and its skin result are in hand, rather than in a separate post-processing pass.
   VolData vdata = {volume_set, Sense::FORWARDS};
-  createSurfaces(tris, vdata, surf_id);
+  createSurfacesFromSkin(region, tris, vdata, surf_id);
 
-  // Create surface sets for the reversed tris
```
GiudGiud, line 1094: adapt the comments instead of deleting them

src/userobjects/MoabSkinner.C
```diff
@@ -944,22 +1228,21 @@ MoabSkinner::createSurfaces(moab::Range & faces, VolData & voldata, unsigned int
...
       }
       else
       {
-        // If overlap is subset, subtract shared tris from this surface and create a new shared
-        // surface
+        // Overlap is a subset: remove shared tris from this surface and create a new
+        // shared surface carrying both volume relationships and the BC type
         check(_moab->remove_entities(surf, overlap));
 
-        // Append our new volume to the list that share this surf
```
GiudGiud, line 1132: same dont delete previous comments

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+  const auto & boundary_info = getMooseMesh().getMesh().get_boundary_info();
+
+  // Number of entities passed to get_adjacencies per call
+  constexpr int num_entities = 1;
+  // Topological dimension of the faces we want (2 = triangles for a 3D tet mesh)
+  constexpr int surface_dimension = 2;
+
+  for (const auto tet : region)
+  {
+    const auto elem_id_it = _elem_handle_to_id.find(tet);
```
GiudGiud, line 1025: these repeated finds make me think you should be using a unordered_map

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+      mooseError("Could not map a MOAB tet back to a libMesh element while assigning "
+                 "DAGMC boundary conditions.");
+
+    const Elem * const elem = getMooseMesh().queryElemPtr(elem_id_it->second);
+    if (!elem)
+      mooseError("Could not find libMesh element ",
+                 elem_id_it->second,
+                 " while assigning DAGMC boundary conditions.");
+
+    for (unsigned int side = 0; side < elem->n_sides(); ++side)
```
GiudGiud:
```suggestion
    for (const auto side : make_range(elem->n_sides()))
```
otherwise you could be calling elem->n_sides() every iteration

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+      boundary_info.boundary_ids(elem, side, side_bids);
+
+      const auto bc_type = boundaryConditionType(side_bids);
+      if (bc_type == BoundaryConditionType::None)
+        continue;
+
+      // Build the set of MOAB vertex handles covering this libMesh side
+      std::unique_ptr<const Elem> side_elem = elem->build_side_ptr(side);
+      std::set<moab::EntityHandle> side_verts;
+      for (unsigned int i = 0; i < side_elem->n_nodes(); ++i)
```
GiudGiud: samne

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+    return;
+
+  const auto & boundary_info = getMooseMesh().getMesh().get_boundary_info();
+
+  // Number of entities passed to get_adjacencies per call
+  constexpr int num_entities = 1;
+  // Topological dimension of the faces we want (2 = triangles for a 3D tet mesh)
+  constexpr int surface_dimension = 2;
+
+  for (const auto tet : region)
```
GiudGiud, line 1023: what is a tet? how big of a tuple is it?
if bigger than trivial, then use const auto & tet : region

src/userobjects/MoabSkinner.C
```diff
@@ -700,9 +750,238 @@ MoabSkinner::materialName(const unsigned int & block,
...
+                     side_elem->node_id(i),
+                     " to a MOAB vertex while assigning DAGMC boundary conditions.");
+        side_verts.insert(node_it->second);
+      }
+
+      // Retrieve the MOAB triangle faces adjacent to this tet
+      moab::Range tri_faces;
+      check(_moab->get_adjacencies(&tet, num_entities, surface_dimension, true, tri_faces));
+
+      for (const auto tri : tri_faces)
```
GiudGiud, line 1057: same question, what is packed into tri_faces, is it trivially copyable or not

GiudGiud (changes requested):

src/userobjects/MoabSkinner.C
meltawila: it's a `moab::EntityHandle` a typedef for an unsigned integer type. trivially copyable and pointer sized

src/userobjects/MoabSkinner.C
meltawila: also a `moab::EntityHandle` a typedef for an unsigned integer type. trivially copyable and pointer sized

src/userobjects/MoabSkinner.C
meltawila: `boundaryNamesToIDs` returns a `std:set` so duplicates within the same line ('1 2 3 2') will not cause an error and will do the right thing

test/tests/neutronics/dagmc/add_bcs/gold/transmission_surface.csv
meltawila: yes you're right. I removed this and added internal surface reflective and transmission tests. the new tests also caught a bug in my code: when a second region's skinning split an existing surface that already had a BC recorded, the skinned triangles took the new pass's BC type and lost the original BC (e.g. the reflective BC assigned to the interface sideset via block 1 was being overwritten as transmission when block 2 was skinned). Fixed by merging the previously recorded BC with the current classification when the shared surface is created

src/userobjects/MoabSkinner.C
meltawila: the new tests: sideset_added_twice, side_in_two_sidesets, conflicting_interface_bcs show these cases

include/userobjects/MoabSkinner.h
```diff
@@ -161,6 +161,14 @@ class MoabSkinner : public GeneralUserObject
   const std::shared_ptr<moab::Interface> & moabPtr() const { return _moab; }
 
 protected:
+  /// Boundary condition types that can be assigned to DAGMC surfaces
+  enum class BoundaryConditionType
+  {
+    None,
```
aprilnovak: Is `None` here the best description? In OpenMC, this would be referred to as `Transmission`.

src/userobjects/MoabSkinner.C
```diff
@@ -345,6 +356,20 @@ MoabSkinner::initialize()
   createTags();
 
   createMOABElems();
+
+  // Resolve sideset names/IDs and check for overlaps (The check in the constructor use the string
+  // which might not be correct).
```
aprilnovak: Can you clarify what this "might not be correct" part is referring to? Is this an old comment or something that signifies an outstanding TODO?

src/userobjects/MoabSkinner.C
aprilnovak: Thanks, that makes sense - in this case, can we just slightly modify the phrasing to something like:

"Element 4, side 1 is assigned both vacuum and reflective boundary conditions because this side is part of sidesets 'back' and 'left'. An element side cannot be set to competing types of surface boundary conditions."

The "mesh side" is confusing to me because IMO it conveys something about sidesets, not that this error is more specific to a side of a given element. 

What do you think?

src/userobjects/MoabSkinner.C
aprilnovak: I'm not sure why this was marked resolved, the switch-case was not implemented here? At least somewhere in the code base we want something that is going to insulate us in the future, if a fourth BC type is added (e.g. white boundary is a possibility). We want somewhere the code to error and remind us to update here.

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+        transmission_tris.erase(tri);
+        switch (bc_type)
+        {
+          case BoundaryConditionType::Vacuum:
+            vacuum_tris.insert(tri);
+            break;
+          case BoundaryConditionType::Reflective:
+            reflective_tris.insert(tri);
+            break;
+          default:
```
aprilnovak, line 1084: do we need the `None` case here too? Even if we don't call this function/arrive here for the `None` type, it's more custom to write it explicitly to be clear that there are 3 values of the enumeration. e.g. add:

```
case BoundaryConditionType::None:
  break;
default:
```

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+void
+MoabSkinner::recordBoundaryConditionSurface(moab::EntityHandle surface_set,
+                                            BoundaryConditionType bc_type)
+{
+  switch (bc_type)
+  {
+    case BoundaryConditionType::None:
+      return;
+    case BoundaryConditionType::Vacuum:
+      if (_reflective_bc_surface_sets.count(surface_set))
```
aprilnovak: is it possible to execute this error, or do we already protect against it from the other errors present?

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+        mooseError("A DAGMC surface was assigned both vacuum and reflective boundary conditions.");
+      _vacuum_bc_surface_sets.insert(surface_set);
+      return;
+    case BoundaryConditionType::Reflective:
+      if (_vacuum_bc_surface_sets.count(surface_set))
```
aprilnovak: is it possible to execute this error, or do we already protect against it from the other errors present?

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+      return;
+    default:
+      mooseError("Unhandled Boundary condition type!");
+  }
+}
+
+MoabSkinner::BoundaryConditionType
+MoabSkinner::recordedBoundaryCondition(moab::EntityHandle surface_set) const
+{
+  if (_vacuum_bc_surface_sets.count(surface_set))
```
aprilnovak: same comment here about switch-case for enumerations. It's kind of tedious/verbose but it is a really strong protection future-proofing this code.

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+}
+
+void
+MoabSkinner::createBoundaryConditionGroups()
+{
+  if (!_set_bcs)
+    return;
+
+  if (!_vacuum_bcs_surface_ids.empty() && _vacuum_bc_surface_sets.empty())
+    mooseError("'vacuum_bcs_surfaces' was specified but no skinned DAGMC surfaces were assigned "
```
aprilnovak: I think a `paramError` would be slightly better

src/userobjects/MoabSkinner.C
```diff
@@ -809,9 +838,263 @@ MoabSkinner::materialName(const unsigned int & block,
...
+               "vacuum boundary conditions. Verify the sideset names or IDs correspond to "
+               "boundary faces of the mesh.");
+
+  if (!_reflective_bcs_surface_ids.empty() && _reflective_bc_surface_sets.empty())
+    mooseError("'reflective_bcs_surfaces' was specified but no skinned DAGMC surfaces were "
```
aprilnovak: I think a `paramError` would be slightly better

src/userobjects/MoabSkinner.C
aprilnovak: What is the scenario where we arrive at this (or the next one) error message? To me it seems like it's only possible if a sideset exists but somehow has zero faces on it? (I'm not even sure that is possible).

Disclaimer: I'm reading from top to bottom so I have not seen below yet if you indeed have a test that triggers this :)

test/tests/neutronics/dagmc/add_bcs/equivalent_bcs/openmc.i
```diff
@@ -0,0 +1,103 @@
...
+  [hextotet]
+    type = ElementsToTetrahedronsConverter
+    input = name
+  []
+
+  allow_renumbering = false
+  parallel_type = replicated
+[]
+
+[AuxVariables]
```
aprilnovak: Does this test need `cell_temperature`? If not, delete it to keep tests slimmer

test/tests/neutronics/dagmc/add_bcs/conflicting_bcs.i
```diff
@@ -0,0 +1,114 @@
...
+    input = f
+    expression = 'if (y > 12.5, 1, 2)'
+  []
+  [name]
+    type = RenameBlockGenerator
+    input = subdomains
+    old_block = '1 2'
+    new_block = 'block_1 block_2'
+  []
+  # [vacuum_sideset]
```
aprilnovak: remove these if commented anyways

test/tests/neutronics/dagmc/add_bcs/conflicting_bcs.i
```diff
@@ -0,0 +1,114 @@
...
+  [hextotet]
+    type = ElementsToTetrahedronsConverter
+    input = name
+  []
+
+  allow_renumbering = false
+  parallel_type = replicated
+[]
+
+[AuxVariables]
```
aprilnovak: same comment about cell_temperature

test/tests/neutronics/dagmc/add_bcs/conflicting_bcs.i
```diff
@@ -0,0 +1,114 @@
...
+[AuxKernels]
+  [temp]
+    type = ConstantAux
+    variable = temp
+    value = 500.0
+    execute_on = timestep_begin
+  []
+[]
+
+[Problem]
```
aprilnovak: this block can be simplified by removing tallies - I think they are not needed for the error that is being triggered

test/tests/neutronics/dagmc/add_bcs/conflicting_bcs.i
```diff
@@ -0,0 +1,114 @@
...
+    temperature_max = 1000.0
+    n_temperature_bins = 1
+    temperature = temp
+    build_graveyard = false
+    vacuum_bcs_surfaces = 'left top bottom front back'
+    reflective_bcs_surfaces = 'left right'
+  []
+[]
+
+[Postprocessors]
```
aprilnovak: if this is an error test, you can slim down a lot of things. Remove this PP, can also remove the Outputs block

test/tests/neutronics/dagmc/add_bcs/openmc.i
```diff
@@ -0,0 +1,138 @@
...
+  [hextotet]
+    type = ElementsToTetrahedronsConverter
+    input = corner_nodes
+  []
+
+  allow_renumbering = false
+  parallel_type = replicated
+[]
+
+[AuxVariables]
```
aprilnovak: remove?

test/tests/neutronics/dagmc/add_bcs/openmc_hex.i
```diff
@@ -0,0 +1,134 @@
...
+    input = midplane
+    new_boundary = 'corner_nodes'
+    coord = '-12.5 -12.5 -12.5'
+  []
+
+  allow_renumbering = false
+  parallel_type = replicated
+[]
+
+[AuxVariables]
```
aprilnovak: remove?

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
+[Tests]
+  [sideset_added_twice]
+    type = RunException
+    input = conflicting_bcs.i
+    expect_err = "The following sideset ID\(s\) appear in both 'vacuum_bcs_surfaces' and 'reflective_bcs_surfaces': 4"
```
aprilnovak, line 8: could you delete `conflicting_bcs.i` altogether, and accomplish this error with `cli_args` instead?

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    requirement = "The system shall error if an internal DAGMC surface shared between two skinned "
+                  "regions is assigned vacuum boundary conditions from one side and reflective "
+                  "boundary conditions from the other."
+    capabilities = 'dagmc'
+  []
+  [unused_bcs_sideset]
+    type = RunException
+    input = openmc.i
+    cli_args = 'UserObjects/moab/vacuum_bcs_surfaces=midplane'
+    expect_err = "'vacuum_bcs_surfaces' was specified but no skinned DAGMC surfaces were assigned"
```
aprilnovak, line 36: Do we need to add a second test just like this one, but for reflective BCs?

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    expect_err = "is a nodeset; boundary conditions can only be assigned to sidesets"
+    requirement = "The system shall error if a BC is assigned to a boundary that exists only as a "
+                  "nodeset."
+    capabilities = 'dagmc'
+  []
+  [typo]
+    type = RunException
+    input = typo.i
+    expect_err = "Boundary 'topp' does not exist in the mesh."
+    requirement = "The system shall error if an unknown sideset is assigned a BC."
```
aprilnovak: The `requirement` for every test must be unique, but this one is identical to the `nonexistent_boundary_id` test. Also suggest to make the name `nonexistent_boundary_name` instead of `typo` for a bit of extra clarity

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    capabilities = 'dagmc'
+  []
+  [all_reflective]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = 'all_reflective.csv'
+    cli_args = 'UserObjects/moab/reflective_bcs_surfaces="top bottom front back left right" Outputs/file/file_base="all_reflective"'
+    requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when adding "
```
aprilnovak:
```suggestion
    requirement = "The system shall give identical Monte Carlo solution (k) when adding "
```

aprilnovak: I don't think the test is actually covering the tallies themselves, so this proposed change to the description is more accurate. If you feel like checking the tally values is necessary to ensure correctness, you could add it with a few `PointValue` pp's

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+                  "via 'reflective_bcs_surfaces' as compared to applying BCs on a CSG cell that contains the DAGMC geometry in /equivalent_bcs."
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+  []
+  [all_vacuum]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = 'all_vacuum.csv'
+    cli_args = 'UserObjects/moab/vacuum_bcs_surfaces="top bottom front back left right" Outputs/file/file_base="all_vacuum"'
+    requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when adding "
```
aprilnovak:
```suggestion
    requirement = "The system shall give identical Monte Carlo solution (k) when adding "
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+                  "boundary conditions through the skinner when all external surfaces are assigned vacuum boundary conditions "
+                  "via 'vacuum_bcs_surfaces', as compared to applying BCs on a CSG cell that contains"
+                  "the DAGMC geometry in /equivalent_bcs."
```
aprilnovak:
```suggestion
                  "the DAGMC geometry in ./equivalent_bcs."
```

aprilnovak:
```suggestion
                  "the DAGMC geometry in ./equivalent_bcs."
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    capabilities = 'dagmc'
+  []
+  [mix_bcs]
+      type = CSVDiff
+      input = openmc.i
+      csvdiff = 'mix_bcs.csv'
+      cli_args = "UserObjects/moab/vacuum_bcs_surfaces='left right'
+                         UserObjects/moab/reflective_bcs_surfaces='front back top bottom'
+                         Outputs/file/file_base='mix_bcs'"
+      requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when adding "
```
aprilnovak:
```suggestion
      requirement = "The system shall give identical Monte Carlo solution (k) when adding "
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+                    "boundary conditions through the skinner via 'reflective_bcs_surfaces' and 'vacuum_bcs_surfaces' as compared to applying BCs on a CSG cell that contains "
+                    "the DAGMC geometry in /equivalent_bcs."
```
aprilnovak:
```suggestion
                    "the DAGMC geometry in ./equivalent_bcs."
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+      mesh_mode = 'replicated'
+      capabilities = 'dagmc'
+  []
+  [bcs_via_boundary_ids]
+      type = CSVDiff
+      input = openmc.i
+      csvdiff = 'mix_bcs.csv'
+      cli_args = 'UserObjects/moab/vacuum_bcs_surfaces="4 2" UserObjects/moab/reflective_bcs_surfaces="5 back 3 1" Outputs/file/file_base="mix_bcs"'
+      requirement = "The system shall give identical Monte Carlo solution (file mesh tallies and k) when adding boundary conditions through the "
```
aprilnovak:
```suggestion
      requirement = "The system shall give identical Monte Carlo solution (k) when adding boundary conditions through the "
```

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+                    " skinner via 'reflective_bcs_surfaces' and 'vacuum_bcs_surfaces' using sideset id or name as compared to applying BCs on a CSG cell that contains "
+                    "the DAGMC geometry in /equivalent_bcs."
```
aprilnovak:
```suggestion
                    "the DAGMC geometry in ./equivalent_bcs."
```

src/userobjects/MoabSkinner.C
meltawila: No this is referring to a previous review comment about overlap check which suggested checking for reflecting/vacuum surfaces in the constructor instead. I'm saying here that the check in the constructor might not be enough because it only checks for overlap between `vacuum_bcs_surfaces` and `reflective_bcs_surfaces` entries however here these are resolved to boundary IDs. example: sideset ID 4 named "left" and 4 is entered as vacuum and "left" entered as reflective would be catched here

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+      capabilities = 'dagmc'
+  []
+  [internal_transmission]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = 'internal_transmission.csv'
+    cli_args = "UserObjects/moab/vacuum_bcs_surfaces='top'
+                UserObjects/moab/reflective_bcs_surfaces='bottom front back left right'
+                Outputs/file/file_base='internal_transmission'"
+    requirement = "The system shall treat an internal skinned surface (the interface between the "
```
aprilnovak, line 132: Is this test specific to the fact that it is an _internal_ skinned surface? Does this behavior also occur if a side is not specified in either of the `vacuum_bcs_surfaces` or `reflective_bcs_surfaces`? (I think so? because that's what the skinner was doing formerly by handling all particle BCs in the graveyard?)

src/userobjects/MoabSkinner.C
meltawila: not an outstanding TODO

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    capabilities = 'dagmc'
+  []
+  [hexmesh_internal_transmission]
+    type = CSVDiff
+    input = openmc_hex.i
+    csvdiff = 'internal_transmission.csv'
+    cli_args = "UserObjects/moab/vacuum_bcs_surfaces='top'
+                UserObjects/moab/reflective_bcs_surfaces='bottom front back left right'
+                Outputs/file/file_base='internal_transmission'"
+    requirement = "The system shall use a hexmesh and treat an internal skinned surface (the interface between the "
```
aprilnovak, line 161: Can you clarify what this test is covering specifically? Is it doing an internal conversion from quads to tris on the surfaces?

test/tests/neutronics/dagmc/add_bcs/tests
```diff
@@ -0,0 +1,163 @@
...
+    input = openmc.i
+    cli_args = 'UserObjects/moab/vacuum_bcs_surfaces=corner_nodes'
+    expect_err = "is a nodeset; boundary conditions can only be assigned to sidesets"
+    requirement = "The system shall error if a BC is assigned to a boundary that exists only as a "
+                  "nodeset."
+    capabilities = 'dagmc'
+  []
+  [typo]
+    type = RunException
+    input = typo.i
```
aprilnovak: can probably accomplish this test with `openmc.i` combined with `cli_args` to allow deletion of `typo.i`?

src/userobjects/MoabSkinner.C
aprilnovak: does this mean that we have duplicative checks happening? If we can, we should just have one location in the code which is checking the full range of possibilities (a boundary, either as named or ID, is duplicated in `vacuum_bcs_surfaces` and `reflective_bcs_surfaces`). This would make it easier to read/understand, from my perspective

meltawila: there's no duplication, I removed the constructor check in a previous commit. I'm updating this comment to clear this up

src/userobjects/MoabSkinner.C
meltawila: This part with the latest commit looks like the below - the changes are mainly to take elem, side, boundary info to show them in error messages, but the logic you're commenting on for consutrcting the enums is unchanged
```
MoabSkinner::BoundaryConditionType
MoabSkinner::boundaryConditionType(const Elem * const elem,
                                   const unsigned int side,
                                   const libMesh::BoundaryInfo & boundary_info) const
{
  std::vector<boundary_id_type> side_bids;
  boundary_info.boundary_ids(elem, side, side_bids);

  std::vector<boundary_id_type> vacuum_bids, reflective_bids;
  for (const auto bid : side_bids)
  {
    if (_vacuum_bcs_surface_ids.count(bid))
      vacuum_bids.push_back(bid);
    if (_reflective_bcs_surface_ids.count(bid))
      reflective_bids.push_back(bid);
  }

...
...

  if (!vacuum_bids.empty())
    return BoundaryConditionType::Vacuum;
  if (!reflective_bids.empty())
    return BoundaryConditionType::Reflective;
  return BoundaryConditionType::Transmission;
}
```
The switch case enums you asked me for are now in `boundaryConditionGroupName()` where it's used to `createBoundaryConditionGroups()`, and also `splitSkinByBoundaryCondition()` uses the switch case to list all BC options with a `default:mooseError` for unhandled future new types. However there's no switching to do here really since this function constructs the enum value from sideset membership not consuming them. Adding a future BC type will still need a membership check in this function

src/userobjects/MoabSkinner.C
meltawila: Added your suggestion. `test:neutronics/dagmc/add_bcs.side_in_two_sidesets` shows this error

aprilnovak:
@gonuke did you have any more comments/review you wanted to add here? I'm not that familiar with Moab so I was mostly reviewing with design-by-test mentality.

src/userobjects/MoabSkinner.C
meltawila: The corresponding test is `add_bcs.unused_bcs_sideset`. the scenario is not an empty sideset but an interior sideset to a skinned region when going from the TET4 to TRI3 mesh so no corresponding DAGMC surface to assign the BC to

include/userobjects/MoabSkinner.h
meltawila: changed `None` to `Transmission` everywhere. functionally we're still not assigning anything to these as OpenMC defaults to transmission

aprilnovak: right, I just think the `Transmission` name is more aligned with what OpenMC calls it. Thanks!

src/userobjects/MoabSkinner.C
meltawila: added explicitly. the corresponding change to this comment is ~ line 1073:
```
          case BoundaryConditionType::Transmission:
            // Nothing to do - tris stay in transmission_tris
            break;
```

src/userobjects/MoabSkinner.C
meltawila: The test to this error is `add_bcs.conflicting_interface_bcs`. the overlap check doesn't catch it because these are seperate sidesets with different names and IDs created from two different blocks but in reality point to the same face.
for this comment and the following two comments, now instead of using `_reflective_bc_surface_sets ` and `_vacuum_bc_surface_sets ` a `std::map<surface, BCType> _surface_bc_types` is used to generalize recording and looking up BC types and no branch per type is used. new enums for future BC supported types would go in `boundaryConditionGroupName()`

src/userobjects/MoabSkinner.C
meltawila: same as the comment above

src/userobjects/MoabSkinner.C
meltawila: same as above. The switch case enumerations are now in `boundaryConditionGroupName()`and `_surface_bc_types ` map now replaces `_vacuum_bc_surface_sets ` here

test/tests/neutronics/dagmc/add_bcs/tests
meltawila: it's similar to add_bcs.internal_transmission test except that it's starting with a hex mesh yes and doing the conversion to TET4 internally then compare against the same csvdiff from the first test. added it to check that nothing is conflicting with the recently added internal conversion PR, I can remove it if you think it's extra

test/tests/neutronics/dagmc/add_bcs/tests
meltawila: yes, not being listed in either parameter is the default for every skinned surface (external or internal) and defaults to transmission in OpenMC and the graveyard handles particle termination similar to before adding this PR. I updated this test to have this as an internal surface because previously when I was testing a transmission external surface Guillaume pointed out that the k_eff result in the csv was identical to that same surface being a vacuum from another test (a particle crossing the transmission surface enters the implicit complement and dies when hitting the graveyard) and it wasn't testing something new between the two cases

test/tests/neutronics/dagmc/add_bcs/tests
aprilnovak: I think it's worthwhile to keep - can you explain the additional context here in the `requirement` field? (or add it as a comment in `tests`, just so we have that extra info documented)

src/userobjects/MoabSkinner.C
```diff
@@ -873,9 +902,296 @@ MoabSkinner::materialName(const unsigned int & block,
...
+std::string
+MoabSkinner::boundaryConditionGroupName(BoundaryConditionType bc_type) const
+{
+  switch (bc_type)
+  {
+    case BoundaryConditionType::Vacuum:
+      return "boundary:Vacuum";
+    case BoundaryConditionType::Reflective:
+      return "boundary:Reflecting";
+    case BoundaryConditionType::Transmission:
```
aprilnovak, line 1141: I think what this will do is that IF you enter this function for `Transmission`, because you don't have a return statement here, it'll flow through to `default`: [google](https://www.google.com/search?q=c%2B%2B+switch-case+fall+through&sca_esv=deedb4ef25b2019f&sxsrf=APpeQnvvI_Q6e3P1HHhCnIfOHY4E1OtkCQ%3A1785179351408&ei=16xnavzHGPu5wN4P16rg-Ao&biw=1512&bih=823&ved=0ahUKEwi8j-eox_OVAxX7HNAFHVcVGK8Q4dUDCBA&uact=5&oq=c%2B%2B+switch-case+fall+through&gs_lp=Egxnd3Mtd2l6LXNlcnAiHGMrKyBzd2l0Y2gtY2FzZSBmYWxsIHRocm91Z2gyBhAAGBYYHjIGEAAYFhgeMgsQABiABBiKBRiGAzILEAAYgAQYigUYhgMyBRAAGO8FMgUQABjvBUiPIlDbBFi8IXAIeAGQAQCYAYEBoAHUFqoBBDI3LjW4AQPIAQD4AQGYAiigAvQXwgIKEAAYRxjWBBiwA8ICCxAAGIAEGIoFGJECwgIKEAAYgAQYigUYQ8ICDRAAGIAEGIoFGEMYsQPCAhEQABiABBiKBRiRAhixAxiDAcICBRAAGIAEmAMAiAYBkAYIkgcEMzIuOKAHxKIBsgcEMjQuOLgH0xfCBwcwLjIxLjE5yAdygAgB&sclient=gws-wiz-serp)

my guess is that this function just never gets entered for `Transmission? (otherwise you would fall-through to the default `mooseError`). My suggestion is to put a `mooseAssert` under `Transmission` case here which catches this

src/userobjects/MoabSkinner.C
```diff
@@ -1120,22 +1443,34 @@ MoabSkinner::createSurfaces(moab::Range & faces, VolData & voldata, unsigned int
...
+        // The shared tris may have been assigned a BC when 'surf' was created (e.g. by
+        // the region on the other side of an internal surface); merge that record with
+        // the current classification so the BC is not lost when the tris move to the
+        // new shared surface
+        auto merged_bc = bc_type;
+        const auto existing_bc = recordedBoundaryCondition(surf);
+        if (merged_bc == BoundaryConditionType::Transmission)
+          merged_bc = existing_bc;
+        else if (existing_bc != BoundaryConditionType::Transmission && existing_bc != merged_bc)
+          mooseError("A DAGMC surface was assigned both vacuum and reflective boundary "
```
aprilnovak, line 1472: can we add anything to this error message to help the user debug?

src/userobjects/MoabSkinner.C
meltawila: Added

src/userobjects/MoabSkinner.C
meltawila: yes, added

meltawila:
`test:userobjects/subchannel_layered.ordering_error` failure shows with the new moose submodule update and is not related to this PR

aprilnovak:
ready for the rebase now

aprilnovak:
LGTM!

PR #1270, opened 2026-01-09
Fix functor system with ExternalProblem-derived classes

nuclearkevin:
This PR also modifies a test to ensure we hit this in the future.

Closes #1269

PR #1288, opened 2026-02-11
update moose

aprilnovak:
This PR brings over the changes from #1270 and adds some additional test updates for `--heavy` tests. @nuclearkevin I would appreciate your review of the two heavy tests which I had to update (diffs on order of 1e-3 for point fields and 1e-5 for postprocessors).


test/tests/conduction/nonidentical_volume/cylinder/nek_master.i
```diff
@@ -102,16 +102,22 @@ o=0.15
...
 [Transfers]
   [temperature]
-    type = MultiAppNearestNodeTransfer
+    type = MultiAppGeneralFieldNearestLocationTransfer
     source_variable = temp
     from_multi_app = nek
     variable = nek_temp
+    search_value_conflicts = true
+    greedy_search = true
+    value_conflicts_output = '2000'
```
nuclearkevin, line 111:
```suggestion
```
I had this line here to debug the transfer non-determinism. It'll need to be removed to avoid polluting test output.

test/tests/conduction/nonidentical_volume/cylinder/nek_master.i
```diff
@@ -102,16 +102,22 @@ o=0.15
...
   []
   [source]
-    type = MultiAppNearestNodeTransfer
+    type = MultiAppGeneralFieldNearestLocationTransfer
     source_variable = source
     to_multi_app = nek
     variable = heat_source
+    search_value_conflicts = true
+    greedy_search = true
+    value_conflicts_output = '2000'
```
nuclearkevin, line 120:
```suggestion
```

test/tests/conduction/nonidentical_volume/cylinder/tests
```diff
@@ -16,6 +16,7 @@
                   "heat source 'computed' by some other app. The solution matches the standalone MOOSE "
                   "case very well - postprocessors for the volume-averaged and maximum temperature match the "
                   "MOOSE standalone case to within 0.1%."
+    exodiff_opts = '-show_all_diffs'
```
nuclearkevin, line 19:
```suggestion
```

nuclearkevin (changes requested):
It looks like one of the tests is still non-deterministic. The solution Guillaume proposed is to project the nodal variables to elemental variables, then transfer those. It should fix the non-determinism at the cost of requiring another regold.

Alternatively, we could flag the test so it only runs with a replicated mesh.

test/tests/conduction/nonidentical_volume/cylinder/tests
aprilnovak: github was glitching and I couldn't add this to a batch, so I added manually

lewisgross1296:
How is this going? Was hoping to use an updated Cardinal on INL HPC which I think needs this MOOSE update

aprilnovak:
@nuclearkevin @MohamedElkamash @meltawila I'd like to propose we skip the one problematic test because IMO it is more mission-critical to get this MOOSE update in and the distributed mesh failure for the single test is likely a peculiarity of the test and not any actual failure in Cardinal/MOOSE (since all other tests pass and we have many tests covering CHT). But I understand this is not usually acceptable. What are your thoughts?

nuclearkevin:
I'm not a massive fan of needing to add a `skip` to this test, but it appears to be necessary. I think it's acceptable if we open an issue (bug report in this case) to track the non-determinism and aim to address it at some point.

aprilnovak:
I agree, but I've had around 7-8 different users of Cardinal reporting problems associated with our very old MOOSE version, and I want to get a fix for them ASAP. I will open an issue and promise to resolve it after the semester ends!

nuclearkevin:
Works for me!

nuclearkevin (approved):

PR #1297, opened 2026-02-25
RotationSearch: allow CriticalitySearch to use modify OpenMCCellTransform and update OpenMC model

### LINKED ISSUE #1294: Enable the CriticalitySearch system to modify OpenMCCellTransforms for multiphysics search for keff
## Reason

Currently, users can run a `CriticalitySearch` on each iteration with the current step multiphysics solutions. This replaces each step's normal OpenMC solve with a search for keff. However, there is no way to save the critical guess and update the OpenMC model/MOOSE mesh for the next iteration. 

As of #1256, users can specify an `OpenMCCellTransform` UserObject that applies either a rotation and/or translation to some cell(s). To do a multiphysics search for keff, we'd have Cardinal update the OpenMC model and  the MOOSE mesh so that the next iteration solves using the last steps critical guess. This will allow the feedback effects of changing control to be included in the multiphysics model.

## Design

The most straightforward way would be to design a new class that inherits `CriticalitySearchBase` that is able to create / modify an `OpenMCCellTransform` UserObject for the next timestep based on the current iteration critical guess. We likely want to have a separate class for each kind of search, e.g. `DrumAngleSearch` or `RodInsertionSearch`. Probably best to separate the modifications to the OpenMC model and modifications to the MOOSE mesh in two separate PRs.

## Impact

This shouldn't need to modify the API for the existing features mentioned, but it will add a new `CriticalitySearch` capability.

**Additional context**
Tagging @gonuke @meltawila

lewisgross1296:
Added a new `DrumAngleSearch` that inherits from `CriticalitySearchBase` and modifies the current time step `OpenMCCellTransform` to the critical guess from the last guess. Should make rotational searches more efficient. Closes #1294.


lewisgross1296:
This is compiling locally and on CIVET, so I think the code part of this task is ready for review. I'll get to working on docs and tests tomorrow

aprilnovak:
For your test cases, I recommend that you follow a similar approach as for the other criticality tests - make a standalone OpenMC model that runs its own criticality search, to find the converged configuration. Then, repeat the search but driven by Cardinal, to see if you can converge to the same value. For this to be "fair", since the search methods are different, you should run with a huge number of particles to ensure the 2 simulations match within statistics.

Then, for the actual test, just bump down the number of particles so that the test runs quickly -- and we'd then check against that "frozen" criticality search. Should the code change for any reason/break in the future, we'd still catch it.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,65 @@
...
+}
+
+void
+DrumAngleSearch::updateOpenMCModel(const Real & angle)
+{
+  _console << "Searching for drum angle = " << angle << units() << std::endl;
+
+  switch (_rotation_axis_char)
+  {
+    case 'x':
```
aprilnovak: If you're doing this type of check here, there's no point in defining the enumeration in the `CardinalEnums.C` file - you can delete the enumeration there and just have it in `DrumAngleSearch::validParams()` like

```
MooseEnum xyz("x y z");
```

We put enumerations in that header file if they're used across multiple classes, but this one is simple.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -55,6 +55,9 @@ OpenMCCellTransform::validParams()
 OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
   : GeneralUserObject(parameters),
     OpenMCBase(this, parameters),
+    _t0_pp_name(getPostprocessorName("vector_value", 0)),
```
aprilnovak: Why can't get `getPostprocessorName` be called from `DrumAngleSearch`? `OpenMCCellTransform` itself does not need to know the names, so for OOP it is preferred to have member variables be as local as possible.

src/criticality/DrumAngleSearch.C
```diff
@@ -0,0 +1,65 @@
...
+
+InputParameters
+DrumAngleSearch::validParams()
+{
+  auto params = CriticalitySearchBase::validParams();
+  params.addRequiredParam<UserObjectName>(
+      "transform_name",
+      "The OpenMCCellTransform UserObject that will control the criticality search");
+  params.addRequiredParam<MooseEnum>(
+      "rotation_axis", getRotationAxisEnum(), "Axis about which to rotate the cell. [x, y, z]");
```
aprilnovak:
```suggestion
      "rotation_axis", getRotationAxisEnum(), "Axis about which to rotate the cell's fill.");
```

aprilnovak: you don't need to manually write the possible enumeration values, MooseDocs will fill it for you

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,65 @@
...
+InputParameters
+DrumAngleSearch::validParams()
+{
+  auto params = CriticalitySearchBase::validParams();
+  params.addRequiredParam<UserObjectName>(
+      "transform_name",
+      "The OpenMCCellTransform UserObject that will control the criticality search");
+  params.addRequiredParam<MooseEnum>(
+      "rotation_axis", getRotationAxisEnum(), "Axis about which to rotate the cell. [x, y, z]");
+  params.addClassDescription("Searches for criticality rotating a drum angle in degrees");
```
aprilnovak: Do we want to make this class have a more generic name? Could you envision someone using it for rotating something which might not be a control drum?

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,65 @@
...
+
+DrumAngleSearch::DrumAngleSearch(const InputParameters & parameters)
+  : CriticalitySearchBase(parameters),
+    UserObjectInterface(this),
+    _transform_name(getParam<UserObjectName>("transform_name")),
+    _rotation_axis_char(getParam<MooseEnum>("rotation_axis"))
+{
+  // update the OpenMCCellTransform UserObject with the last iteration critical guess
+  _t =
+      const_cast<OpenMCCellTransform *>(&getUserObjectByName<OpenMCCellTransform>(_transform_name));
```
aprilnovak: you need to check that this type cast didn't return a nullpointer - what if you passed a userobejct which wasn't of type `OpenMCCellTransform`?

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,65 @@
...
+
+void
+DrumAngleSearch::updateOpenMCModel(const Real & angle)
+{
+  _console << "Searching for drum angle = " << angle << units() << std::endl;
+
+  switch (_rotation_axis_char)
+  {
+    case 'x':
+      _t->setTransformPPValues(std::make_tuple(_t->_t0_pp_name, angle),
```
aprilnovak: this looks weird with the tuples - why not have 3 members functions, each which only changes x, y, or z?

src/criticality/RotationSearch.C
lewisgross1296: I think that's fair. Perhaps just `RotationSearch`?

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I was running into protected variable issues when I tried to use `_t` to get the `PostprocessorName` in the `DrumAngleSearch` class, so I made the names a public member of the transform. Probably I was just doing it the wrong way by trying to use `_t` / creating more public members. 

Maybe I can probably just call `getPostprocessorName("vector_value", N)` for `N=0,1,2` from `DrumAngleSearch` and it will work the same. This should work as long as the parameter "vector_value" exists, right?

src/criticality/RotationSearch.C
lewisgross1296: My C++ rustiness is showing... thanks for the catch! I'll add it in the next commit

src/criticality/RotationSearch.C
lewisgross1296: I can add some functions to accomplish the same goal in the next commit

src/criticality/DrumAngleSearch.C
```diff
@@ -0,0 +1,65 @@
...
+    case 'y':
+      _t->setTransformPPValues(std::make_tuple(_t->_t0_pp_name, Real(0.0)),
+                               std::make_tuple(_t->_t1_pp_name, angle),
+                               std::make_tuple(_t->_t2_pp_name, Real(0.0)));
+    case 'z':
+      _t->setTransformPPValues(std::make_tuple(_t->_t0_pp_name, Real(0.0)),
+                               std::make_tuple(_t->_t1_pp_name, Real(0.0)),
+                               std::make_tuple(_t->_t2_pp_name, angle));
+  }
+  // _t->execute(); // update transform with new angle guess
```
gonuke:
```suggestion
  std::vector<Real> angles = {Real(0.0), Real(0.0), Real(0.0)};
  angles[int(_rotation_axis_char)] = angle;
  _t->setTransformPPValues(std::make_tuple(_t->_t0_pp_name, angles[0]),
                           std::make_tuple(_t->_t1_pp_name, angles[1]),
                           std::make_tuple(_t->_t2_pp_name, angles[2]);
```

src/userobjects/OpenMCCellTransform.C
gonuke: A high level design question: Should `DrumAngleSearch` be updating the `PostProcessor` directly and then letting `OpenMCCellTransform` only read from the `PostProcessor` **OR** Should `DrumAngleSearch` update `OpenMCCellTransform` which is responsible for updating the `PostProcessor`.

Putting this another way: Should `OpenMCCellTransform` have a read-only relationship with the `PostProcessor`, and `DrumAngleSearch` have a write-only relationship with the `PostProcessor`?  **OR** should `DrumAngleSearch` only have a relationship with `OpenMCCellTransform` and rely on it to have a read/write relationship with the `PostProcessor`?

gonuke: If we go with the latter, then the new function to set the parameters for `OpenMCCellTransform` can be very simple and just have a `std::vector<Real>` as its argument with assumed ordering, just as the initializer for `OpenMCCellTransform` has an assumed ordering.

gonuke: If we go with the former, then `DrumAngleSearch` will need to look up the `PostProcessor` by name, and therefore rely on assumed knowledge of the name of the parameter that `OpenMCCellTransform` uses to specify the `PostProcessor`. I'd argue that this is a bad idea.

include/userobjects/OpenMCCellTransform.h
```diff
@@ -30,6 +31,11 @@ class OpenMCCellTransform : public GeneralUserObject, public OpenMCBase
 
   OpenMCCellTransform(const InputParameters & parameters);
 
+  /// Transform type: "translation" or "rotation"
+  const MooseEnum _transform_type;
```
gonuke: We probably want to implement a getter for this rather than exposing it on the public interface

include/userobjects/OpenMCCellTransform.h
```diff
@@ -30,6 +31,11 @@ class OpenMCCellTransform : public GeneralUserObject, public OpenMCBase
...
+
+  void setTransformPPValues(const Real t0_pp_value, const Real t1_pp_value, const Real t2_pp_value);
```
gonuke: Why can't we just pass the vector?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,20 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
+                                          const Real t1_pp_value,
+                                          const Real t2_pp_value)
+{
+  const auto & pp_name_vector = getParam<std::vector<PostprocessorName>>("vector_value");
+  if (!MooseUtils::isFloat(pp_name_vector[0]))
+    _openmc_problem->setPostprocessorValueByName(pp_name_vector[0], t0_pp_value, 0);
+  if (!MooseUtils::isFloat(pp_name_vector[1]))
+    _openmc_problem->setPostprocessorValueByName(pp_name_vector[1], t1_pp_value, 0);
+  if (!MooseUtils::isFloat(pp_name_vector[2]))
+    _openmc_problem->setPostprocessorValueByName(pp_name_vector[2], t2_pp_value, 0);
```
gonuke: If we passed in a vector, these could be in a loop

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,20 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
   _t2_pp = &getPostprocessorValue("vector_value", 2);
 }
 
+void
+OpenMCCellTransform::setTransformPPValues(const Real t0_pp_value,
+                                          const Real t1_pp_value,
+                                          const Real t2_pp_value)
+{
+  const auto & pp_name_vector = getParam<std::vector<PostprocessorName>>("vector_value");
+  if (!MooseUtils::isFloat(pp_name_vector[0]))
```
gonuke: I'm not 100% sure what this is testing for, but shouldn't/couldn't we do it at initialization? Isn't it an error condition if we can't set this?

gonuke (changes requested):
Here are a few more C++ design questions.

I don't want to hold this up, but I don't actually see any reason for this to be limited to a rotation search.  If one implemented a translation search (that was confined to the primary axes of the problem) I think the code would be literally identical (other than the test to ensure that the transform is a rotation).

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,68 @@
...
+               "You have attempted search for critical drum angle on the OpenMCCellTransform " +
+                   _transform_name +
+                   ", which does not modify a cell rotation."
+                   "Please select a transform that rotates a control drum cell.");
+}
+
+void
+RotationSearch::updateOpenMCModel(const Real & angle)
+{
+  _console << "Searching for drum angle = " << angle << units() << std::endl;
```
gonuke: Probably should remove the reference to `drum angle` since the search is more generically about rotation

gonuke (changes requested):
One more thing...

include/userobjects/OpenMCCellTransform.h
lewisgross1296: Sure, I can do that. Will address it in the next commit

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I was thinking about looping this yesterday yesterday. I agree it would be better.

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I'm not actually conviced the `!MooseUtils::isfloat()` is the right way to go because we should set all three values (number or postprocessor), since this setter is more general, and we happen to use it in `RotationSearch.C`.

The reason this is currently here is due to an issue I had yesterday. For example
```
vector_value = '0.0 0.0 drum_angle'
```
was trying to find a `Postprocessor` named `0.0` and set its `value`, but one does not exist.  It was suggested to type check via `isFloat` in the MOOSE slack to prevent this. 

For the rotation search case, we want exactly one `Postprocessor` to exist and we want the other two components to be set to exactly zero.  I'm currently worried about the case where someone uses this method in a more general sense, sets three numbers in `vector_value` and then nothing gets set. 

My current issue here is that `OpenMCCellTransform` has three member variables for the three `PostprocesorValue`s in the header

```C++
  /// Postprocessor providing the first entry of the transform array
  const PostprocessorValue * _t0_pp;

  /// Postprocessor providing the second entry of the transform array
  const PostprocessorValue * _t1_pp;

  /// Postprocessor providing the third entry of the transform array
  const PostprocessorValue * _t2_pp;
```
but if these don't actually correspond to an existing `Postprocessor` with a valid name, as far as I can tell, you can't set their value via `setPostprocessorValueByName` since
```C++
const auto & pp_name_vector = getParam<std::vector<PostprocessorName>>("vector_value");
```
will populate a vector entry with just the number if a numeric value is given to any `vector_value` entry and not a name of an existing Postprocessor. Looking for advice from @meltawila @aprilnovak

src/criticality/RotationSearch.C
lewisgross1296: I noticed this too and have some changes to fix it locally, will push later

meltawila:
> Since the reference to the PP stored in OpenMCCellTransform may be either a numerical constant or a named PP, it's not entirely clear (to us) how to set it correctly

PostprocessorInterface treats user provided constants as default postprocessor values. "If the Postprocessor is a default value (either set by addParam or set to a constant value by the user in input/an action), we covert the name to a value, store said value locally, and return it so that it fits in with the rest of the interface (needing a reference to a value)". But to use setPostprocessorValueByName() it needs to have an actual declared Postprocessor

This can be checked with isDefaultPostprocessorValue()

so we can use:
`if (!isDefaultPostprocessorValue(pp_name_vector[pp_idx]))
       _openmc_problem->setPostprocessorValueByName(pp_name_vector[pp_idx], pp_values[pp_idx], 0);`

but this is probably functionally the same as if (!MooseUtils::isFloat(pp_name_vector[pp_idx])) currently used

gonuke:
Just by its name, I think `isDefaultPostprocessorValue` may be a choice that both conveys a better intent and may match its design intent

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,53 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
 }
 
+MooseEnum
+OpenMCCellTransform::getTransformType() const
+{
+  return _transform_type;
+}
+
+void
+OpenMCCellTransform::checkTransformIsValidRotationForCriticalitySearch() const
```
gonuke: Separation of concerns: This function is too specific to rotation to be here.

gonuke: Don't you also have to check that the PP represents the correct variable?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,53 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
+
+  int zero_count = 0;
+  int one_PP_name = 0;
+  for (PostprocessorName name : pp_name_vector)
+    if (MooseUtils::isFloat(name) && std::abs(std::stof(name)) < 1e-6)
+      zero_count++;
+    else if (!MooseUtils::isFloat(name))
+      one_PP_name++;
+    else
+      paramError("vector_value",
```
gonuke: Is this an error if it is the angle that you are changing?

gonuke (changes requested):
I think the new test may be too intrusive in `OpenMCCellTransform`.

It seems like you need:
- One function to return the transform type or to returned whether it is a particular type
- One function that confirms that the appropriate PP is valid, for some definition of valid
- One function to set the particular PP

src/userobjects/OpenMCCellTransform.C
lewisgross1296: Do you think it should be in `RotationSearch`? I guess I imagined there could be future functions similar for a translational search, but maybe it fits more in the search class. I wanted to call it on the transform object in `RotationSearch`, which is why I put it here. And WRT the name, I think it's fine if it's specific to a rotation search because it's called `checkTransformIsValidRotationForCriticalitySearch`. I basically wanted the ability to confirm that a `vector_value` is properly defined with one Postprocessor and two zeros, which is why it's so specific.

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I don't think so? I'm able to run `CriticalitySearch` in my docker container with this line. This else here means that a non-Postprocessor that is not zero is specified, e.g.
```
vector_value = 1.0 0.0 drum_angle
```
would trigger this else becasue of the `1.0`

lewisgross1296: Or do you mean if someone just specifies a number in `vector_value` for their rotation axis. I guess it would fail then. I could add an `else if` to see what component it is and allow it to work if they specify a number for the correct component / add a test for this case

src/userobjects/OpenMCCellTransform.C
gonuke: (Capturing in-person discussion) `RotationSearch` should be responsible for determining validity for use in a rotation search and not `OpenMCCellTransform`. Some form of getter method can be implemented in `OpenMCCellTransform` to provide access to sufficient information - perhaps just what is returned from `getParam<std::vector<PostprocessorName>>("vector_value")` - to allow `RotationSearch` to implement the logic for this.

Why is this the right design choice? As a form of separation of concerns/object oriented design, `OpenMCCellTransform` should not need to have any special knowledge of what constitutes "validity" for the `RotationSearch` class.

src/userobjects/OpenMCCellTransform.C
gonuke: (Capturing in-person discussion) It is reasonable at this point to constrain the `RotationSearch` to operate on an `OpenMCCellTransform` object that has a `vector_value` with two entries of 0 and one entry that is a named `PostProcessor`. Furthermore, the entry that is the named `PostProcessor` must align with the rotation axis identified in the `RotationSearch` entry.

This means that testing at initialization can include:
* test that the named UserObject exists and is an `OpenMCCellTransform`
* test that the `OpenMCCellTransform` is a rotation
* test that the results of `getParam<std::vector<PostprocessorName>>("vector_value")` is a set of `PostProcessors` that:
   * has only one non-zero entry
   * that non-zero entry is a named `PostProcessor`
   * that non-zero entry represents the correct dimension specified in the `RotationSearch`

It may be possible/appropriate to store the name of that `PostProcessor` to facilitate changing it in later steps.

src/userobjects/OpenMCCellTransform.C
lewisgross1296: Happy to implement a getter that returns a `const` reference to the output of `getParam<std::vector<PostprocessorName>>("vector_value")` so that `RotationSearch` can be aware of the entries in `vector_value` without being able to modify them. We can then port the error checking over to the search class to keep the parts specific to what kind of `vector_value` input is well-posed for a `RotationSearch` within the class that handles it.

src/userobjects/OpenMCCellTransform.C
lewisgross1296: Agree with all of this outside the last part:
> It may be possible/appropriate to store the name of that PostProcessor to facilitate changing it in later steps.

l think since the user will define the `Postprocessor` from their input file, we probably shouldn't do anything to change the name of that `Postprocessor`, since it will lead to conflicts with the input file in the output files created; i.e. it's easily confusing.

gonuke: I meant changing the value not changing the name

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,20 @@
+# OpenMCMaterialDensity
+
+## Description
+
+`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) that searches for critical configuration by updating the value of one or more `openmc.Cell.rotation`(s). This is done by relying on an existing [OpenMCCellTransform](OpenMCCellTransform.md) UserObject that can update the OpenMC model with some added restrictions. The update will be applied to all cells specified by the user, so the same angle of rotation will be applied to all cells specified.
```
nuclearkevin:
```suggestion
`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) that searches for a critical configuration by updating the value of one or more `openmc.Cell.rotation`(s). This is done by relying on an existing [OpenMCCellTransform](OpenMCCellTransform.md) UserObject that can update the OpenMC model with some added restrictions. If the [OpenMCCellTransform](OpenMCCellTransform.md) modifies multiple cells, the same rotation angle will be applied to each cell during the search.
```

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,20 @@
...
+
+Only axis aligned searches are allowed, i.e. you can only rotate cells around the `x`, `y`, or `z` axis. The main reason is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. If you desire to apply more than one rotation to a cell, it's reccomended to instead build your model in such a way the the rotation that changes reactivity in the search is only around one principal axis.
```
nuclearkevin:
```suggestion
Only axis aligned searches are allowed, i.e. you can only rotate cells around the `x`, `y`, or `z` axis. The main reason for this restriction is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. If you desire to apply more than one rotation to a cell, it's reccomended to instead build your model in such a way the the rotation that changes reactivity in the search is only around one principal axis.
```

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,20 @@
...
+
+While a generic `OpenMCCellTransform` allows modifications to all components of an `openmc.Cell.rotation` tuple via the `vector_value` parameter, for performing a `RotationSearch`, the `vector_value` must contain exactly two `0.0` values and one name of a `Postprocessor`, typically a `Receiver` `Postrprocesor`.
+
+## Example Input File Syntax
+Here is a valid `RotationSearch` which shows the corresponding `OpenMCCellTransform` that will be used in the search for critical.
```
nuclearkevin:
```suggestion
Here is a valid `RotationSearch` which shows the corresponding `OpenMCCellTransform` that will be used in the search for a critical state.
```

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "CriticalitySearchBase.h"
+#include "OpenMCCellTransform.h"
+#include "CardinalEnums.h"
+#include "UserObjectInterface.h"
+
+/**
+ * Perform a criticality search based  Pon a material
```
nuclearkevin:
```suggestion
 * Perform a criticality search based on a rotation angle
```

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,36 @@
...
+class RotationSearch : public CriticalitySearchBase,
+                       public UserObjectInterface,
+                       public PostprocessorInterface
+{
+public:
+  static InputParameters validParams();
+
+  RotationSearch(const InputParameters & parameters);
+
+  void checkValidVectorValueForRotationSearch();
```
nuclearkevin: This function needs a description (doxygen comments).

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,36 @@
...
+
+  virtual void updateOpenMCModel(const Real & input) override;
+
+protected:
+  virtual std::string units() const override { return "[degrees]"; }
+  virtual std::string quantity() const override { return "UO: [" + _transform_name + "]"; }
+
+  // name of the OpenMCCellTransform UserObject used to control the criticality search
```
nuclearkevin:
```suggestion
  /// name of the OpenMCCellTransform UserObject used to control the criticality search
```

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,36 @@
...
+  const UserObjectName _transform_name;
+
+  // the rotational axis used to search for criticality
```
nuclearkevin:
```suggestion
  /// the rotational axis used to search for criticality
```

include/userobjects/OpenMCCellTransform.h
```diff
@@ -30,6 +31,12 @@ class OpenMCCellTransform : public GeneralUserObject, public OpenMCBase
 
   OpenMCCellTransform(const InputParameters & parameters);
 
+  MooseEnum getTransformType() const;
```
nuclearkevin: This function needs doxygen comments (description and return).

include/userobjects/OpenMCCellTransform.h
```diff
@@ -30,6 +31,12 @@ class OpenMCCellTransform : public GeneralUserObject, public OpenMCBase
...
+
+  std::vector<PostprocessorName> getVectorValue() const;
```
nuclearkevin: This function needs doxygen comments (description and return).

include/userobjects/OpenMCCellTransform.h
```diff
@@ -30,6 +31,12 @@ class OpenMCCellTransform : public GeneralUserObject, public OpenMCBase
...
+
+  void setTransformPPValues(const std::vector<Real> pp_values);
```
nuclearkevin: This function needs doxygen comments (description and parameters).

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+}
+
+RotationSearch::RotationSearch(const InputParameters & parameters)
+  : CriticalitySearchBase(parameters),
+    UserObjectInterface(this),
+    PostprocessorInterface(this),
+    _transform_name(getParam<UserObjectName>("transform_name")),
+    _rotation_axis_char(getParam<MooseEnum>("rotation_axis"))
+{
+  // update the OpenMCCellTransform UserObject with the last iteration critical guess
```
nuclearkevin: This appears to be an old comment that snuck through. 
```suggestion
```

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+  try
```
nuclearkevin: This entire try-catch block can be made redundant through the use of `getUserObject<OpenMCCellTransform>(...)` on line 45 (which will throw a paramError for you) instead of `getUserObjectByName<OpenMCCellTransform>(...)`. 

You'll often find that if you dumpster dive enough in the framework, you can find a random function that does exactly what you need. The wonders of over-abstraction...

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+    std::string s = e.what();
+    paramError(
+        "transform_name",
+        "In attempting to get the OpenMCCellTransform UserObject, the following error occurred: " +
+            s +
+            "\nThis likely means that the name provided either is incorrect or is not an "
+            "OpenMCCellTransform UserObject.");
+  }
+  _t =
+      const_cast<OpenMCCellTransform *>(&getUserObjectByName<OpenMCCellTransform>(_transform_name));
```
nuclearkevin:
```suggestion
      const_cast<OpenMCCellTransform *>(&getUserObject<OpenMCCellTransform>("transform_name"));
```
^ This is the aforementioned change you can make. The `UserObjectInterface` will fetch the `UserObject` name from this class's input parameters, check to see if the `UserObject` exists and is the right type. If yes, you'll get the `UserObject` cast to the right type. If no, it'll throw a `paramError` for you.

Really not a fan of casting away `const`, but I can't find a better way to do what you need to do here...

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+  // check that the specified transform is a rotational transform
+  if (_t->getTransformType() != "rotation")
+    paramError("transform_name",
+               "You have attempted search for critical drum angle on the OpenMCCellTransform " +
```
nuclearkevin:
```suggestion
               "You have attempted to search for critical rotation angle on the OpenMCCellTransform " +
```

Rotation search, not drum search.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+                   _transform_name +
+                   ", which does not modify a cell rotation."
+                   "Please select a transform that rotates a control drum cell.");
```
nuclearkevin:
```suggestion
                   "Please select a transform that rotates a cell.");
```

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void
+RotationSearch::checkValidVectorValueForRotationSearch()
+{
+  std::vector<PostprocessorName> vv = _t->getVectorValue();
+  int non_zero_idx = int(_rotation_axis_char); // this position in vector_value should contain the
+                                               // only non-default Postprocessor
+
+  int num_postprocessors = 0;
```
nuclearkevin:
```suggestion
```
This is unused.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+        "transform_name",
+        "In attempting to get the OpenMCCellTransform UserObject, the following error occurred: " +
+            s +
+            "\nThis likely means that the name provided either is incorrect or is not an "
+            "OpenMCCellTransform UserObject.");
+  }
+  _t =
+      const_cast<OpenMCCellTransform *>(&getUserObjectByName<OpenMCCellTransform>(_transform_name));
+  // check that the specified transform is a rotational transform
+  if (_t->getTransformType() != "rotation")
```
nuclearkevin: This error is not tested.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+  // confirm that vector_value member of the specified OpenMCCellTransform is valid for a
+  // RotationSearch
+  this->checkValidVectorValueForRotationSearch();
+}
+
+void
+RotationSearch::checkValidVectorValueForRotationSearch()
+{
+  std::vector<PostprocessorName> vv = _t->getVectorValue();
+  int non_zero_idx = int(_rotation_axis_char); // this position in vector_value should contain the
```
nuclearkevin: Is there a reason why you're explicitly casting to an `int` here? The MOOSE way would be to do the following:

```suggestion
 const int non_zero_idx = _rotation_axis_char; // this position in vector_value should contain the
```
(also, mark as const when a value is not changing)

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+
+  // confirm that vector_value member of the specified OpenMCCellTransform is valid for a
+  // RotationSearch
+  this->checkValidVectorValueForRotationSearch();
+}
+
+void
+RotationSearch::checkValidVectorValueForRotationSearch()
+{
+  std::vector<PostprocessorName> vv = _t->getVectorValue();
```
nuclearkevin:
```suggestion
  const auto vv = _t->getVectorValue();
```
The MOOSE style guide requires the use of auto when possible, so long as it doesn't interfere with readability.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,36 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
+std::vector<PostprocessorName>
+OpenMCCellTransform::getVectorValue() const
+{
+  return getParam<std::vector<PostprocessorName>>("vector_value");
+}
+
+void
+OpenMCCellTransform::setTransformPPValues(const std::vector<Real> pp_values)
+{
+  size_t num_pp_values = pp_values.size();
```
nuclearkevin:
```suggestion
  const auto num_pp_values = pp_values.size();
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,36 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
+  {
+    mooseError(
+        "Attempting to use OpenMCCellTransform::setTransformPPValues with a vector that has size " +
+        std::to_string(num_pp_values) + " but it must be exactly of size 3.");
+  }
+  const auto & pp_name_vector = getParam<std::vector<PostprocessorName>>("vector_value");
+  for (size_t pp_idx = 0; pp_idx < num_pp_values; pp_idx++)
+  {
+    if (!MooseUtils::isFloat(pp_name_vector[pp_idx]))
+      _openmc_problem->setPostprocessorValueByName(pp_name_vector[pp_idx], pp_values[pp_idx], 0);
```
nuclearkevin: We should be checking to make sure the `Postprocessor` that is getting set here is a `Receiver` (for all of the reasons Logan mentioned).

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,20 @@
...
+
+Only axis aligned searches are allowed, i.e. you can only rotate cells around the `x`, `y`, or `z` axis. The main reason is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. If you desire to apply more than one rotation to a cell, it's reccomended to instead build your model in such a way the the rotation that changes reactivity in the search is only around one principal axis.
+
+While a generic `OpenMCCellTransform` allows modifications to all components of an `openmc.Cell.rotation` tuple via the `vector_value` parameter, for performing a `RotationSearch`, the `vector_value` must contain exactly two `0.0` values and one name of a `Postprocessor`, typically a `Receiver` `Postrprocesor`.
```
nuclearkevin: IMO, this class should only be used with `Receiver` post-processors due to the issues Logan mentioned.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -102,8 +133,8 @@ OpenMCCellTransform::execute()
       // If a user tried to apply translation on a cell that doesn't contain a filled universe,
       // OpenMC will return an error.
       err = openmc_cell_set_translation(index, vec);
-      _console << "Setting OpenMC cell translations for cell with ID " + std::to_string(cell_id) +
-                      " to ("
+      _console << "Setting OpenMC cell(s) translation for cell(s) with ID " +
```
nuclearkevin:
```suggestion
      _console << "Setting OpenMC cell translation for cell with ID " +
```
This is in a loop, so the plural is unnecessary as multiple outputs occur.

src/userobjects/OpenMCCellTransform.C
```diff
@@ -123,7 +154,7 @@ OpenMCCellTransform::execute()
       // If a user tried to apply rotation on a cell that doesn't contain a filled universe,
       // OpenMC will return an error.
       err = openmc_cell_set_rotation(index, vec, 3);
-      _console << "Setting OpenMC cell rotations for cell with ID " + std::to_string(cell_id) +
+      _console << "Setting OpenMC cell rotation(s) for cell(s) with ID " + std::to_string(cell_id) +
```
nuclearkevin:
```suggestion
      _console << "Setting OpenMC cell rotation for cell with ID " + std::to_string(cell_id) +
```

nuclearkevin (changes requested):
Most of my comments are missing doxygen comments / formatting / missed tests. 

The only major concern I have is not enforcing the use of a `Receiver` post-processor for the criticality search (you're potentially allowing the user to perturb global state, which is prone to hard-to-diagnose errors).

src/criticality/RotationSearch.C
gonuke: It probably better object oriented design to implement `bool OpenMCCellTransform.isRotation()` since then you can let `OpenMCCellTransform` have full control over whether that is true or not.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+  if (_t->getTransformType() != "rotation")
+    paramError("transform_name",
+               "You have attempted search for critical drum angle on the OpenMCCellTransform " +
+                   _transform_name +
+                   ", which does not modify a cell rotation."
+                   "Please select a transform that rotates a control drum cell.");
+
+  // confirm that vector_value member of the specified OpenMCCellTransform is valid for a
+  // RotationSearch
+  this->checkValidVectorValueForRotationSearch();
```
gonuke: You shouldn't need to specify `this` (unless style guides suggest otherwise?):

```suggestion
  checkValidVectorValueForRotationSearch();
```

src/criticality/RotationSearch.C
gonuke: Doesn't it still need to use `_transform_name` rather than `"transform_name"` since the latter is a param?  Or at the very least, we'd need to access the params inline here?

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+  int non_zero_idx = int(_rotation_axis_char); // this position in vector_value should contain the
+                                               // only non-default Postprocessor
+
+  int num_postprocessors = 0;
+  // check vector_value to make sure that the rotation axis position is the only non-zero entry
+  for (int idx = 0; idx < vv.size(); idx++)
+  {
+    // if a numeric value was specified as an entry in vector_value it will be recognizable as a
+    // float to MooseUtils
+    if (MooseUtils::isFloat(vv[idx]))
```
gonuke: It makes more sense to me to invert the logic here and combine some of it:
```
if (idx == non_zero_idx && MooseUtils::isFloat(vv[idx]))
{
   // error
} 
else if (idx != non_zero_idx)
{
   if (!MooseUtils::isFloat(vv[idx]))
   {
      // another error
   } 
   
   if (std::abs(std::stof(vv[idx])) > 1e-6)
   {
      // yet another error
   }
   
}
```

If you crafted the error message properly, the last two could be combined, and possibly combined with the one they are nested within.

src/criticality/RotationSearch.C
gonuke: You could also make this a member variable and set it in the initializer since you also use it in `updateOpenMCModel()`.

You may not even need to store the `char` version and only need to store the integer version since you only ever use the integer version.

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,116 @@
...
+                   ", which does not modify a cell rotation."
+                   "Please select a transform that rotates a control drum cell.");
+
+  // confirm that vector_value member of the specified OpenMCCellTransform is valid for a
+  // RotationSearch
+  this->checkValidVectorValueForRotationSearch();
+}
+
+void
+RotationSearch::checkValidVectorValueForRotationSearch()
```
gonuke: I'm wondering about naming this more plainly:
```
RotationSearch::checkSingleAxisPostProcessorRotation()
```

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,36 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
...
+}
+
+std::vector<PostprocessorName>
+OpenMCCellTransform::getVectorValue() const
+{
+  return getParam<std::vector<PostprocessorName>>("vector_value");
+}
+
+void
+OpenMCCellTransform::setTransformPPValues(const std::vector<Real> pp_values)
```
gonuke: I think it's still an open question about whether the values of the PP get set by a method in `OpenMCCellTransform` or a method in `RotationSearch`.  

Notwithstanding all the warnings about making sure it's safe to change these values at all...  Doing it here fundamentally changes the relationship between `OpenMCCellTransform` and the PPs that is is using.  Until now it has a read-only relationship with those PPs. It may be best to allow it to retain that read-only relationship and require `RotationSearch` to have a `write-only` relationship with the PPs.

I'd be interested others' thoughts on this design question: @nuclearkevin @meltawila @aprilnovak ?

src/userobjects/OpenMCCellTransform.C
```diff
@@ -75,6 +76,36 @@ OpenMCCellTransform::OpenMCCellTransform(const InputParameters & parameters)
   _t2_pp = &getPostprocessorValue("vector_value", 2);
 }
 
+MooseEnum
+OpenMCCellTransform::getTransformType() const
+{
+  return _transform_type;
+}
```
gonuke: Elsewhere, I suggested replacing this with:

```suggestion
bool
OpenMCCellTransform::isRotation() const
{
  return (_transform_type == "rotation");
}
```

It would then also make sense to have
```
bool
OpenMCCellTransform::isTranslation() const
{
  return (_transform_type == "translation");
}
```

src/criticality/RotationSearch.C
gonuke: Might want to also incorporate this into the same method that is checking for validity of the rotation

src/criticality/RotationSearch.C
gonuke: Another alternative, but perhaps too clever (??) is to not put all of these in a loop:
* check the non_zero_idx case explicitly before the loop
* loop from tmp_idx = non_zero_idx+1 to non_zero_idx+2
   * check the case where index is (tmp_idx % 2)

gonuke (changes requested):
A few recommendations for simplifying things further.

src/criticality/RotationSearch.C
loganharbour: - `getXXX()`: Takes the name of a parameter, gets the value from parameter
- `getXXXByName()`: Takes the raw name of the object (should be used sparingly, the former is preferred)

loganharbour: Separate comment here just to make it explicit - the const cast here is quite dirty. User objects are meant to be executed using a dependency graph, by moose itself. Hence why we only give you const references to a UO when you try to couple to it.

Going against this paradigm causes applications to become brittle when optimizations and API changes are made in moose

loganharbour (changes requested):

src/userobjects/OpenMCCellTransform.C
loganharbour: I didn't look through the other comments much, but I think the concern here is aligned with https://github.com/neams-th-coe/cardinal/pull/1297#discussion_r2957692849

src/criticality/RotationSearch.C
lewisgross1296: Gotcha, seems I found the wrong MOOSE command. I will update this / remove the try {} catch {} in the next commit

src/userobjects/OpenMCCellTransform.C
lewisgross1296: I'm happy to include that restriction and I think it makes sense for the Postprocessors that we would use with the OpenMCCellTransform. I'm not finding a simple way to check the type, do you happen to know which function call that is?

src/criticality/RotationSearch.C
lewisgross1296: It was Paul's idea to cast from the actual axis in the enum to an int because "x", "y", and "z" have default enum indices that perfectly correspond to their vector indices of 0,1,2. 

My  C++ isn't refined enough to know if `const int non_zero_idx = _rotation_axis_char;` would compile, but my hunch is that it wouldn't because the types don't match on both sides.

@gonuke are you suggesting something like having
```c++
int _rotation_axis_idx;
```
in the header and then list initializing it as
```c++
    _rotation_axis_idx(int(getParam<MooseEnum>("rotation_axis")))
```
instead of the current list intializer?

src/criticality/RotationSearch.C
lewisgross1296: I guess I'm not sure the alternative. I don't want to write code that could break with future changes, but I don't know of a way to make this `_t` member const. We need to be able to update the `OpenMCCellTransform`'s Postprocessor that corresponds to the drum angle rotation each time we compute a new critical guess so that the OpenMC model actually runs with the new guess. If you have an idea of how to get rid of the const cast, I'm happy to use it

src/criticality/RotationSearch.C
lewisgross1296: We are inherently checking the entries in `"vector_value"` to make sure that there's exactly one Postrprocessor and two zeros, so I think the name `checkValidVectorValue...` makes more sense. E.g. the first thing the method does is 
```c++
const auto vv = _t->getVectorValue();
```

src/userobjects/OpenMCCellTransform.C
gonuke: Note that `OpenMCCellTransform` doesn't care if it's a `Receiver` since it has a read-only relationship.

This is part of the reason that I think `RotationSearch` should directly change the PP rather than change it through `OpenMCCellTransform`

src/criticality/RotationSearch.C
gonuke: > My  C++ isn't refined enough to know if `const int non_zero_idx = _rotation_axis_char;` would compile, but my hunch is that it wouldn't because the types don't match on both sides.

I think it will work, because C++ knows how to do type conversions like this when they are unambiguous.

> @gonuke are you suggesting something like having
> 
> ```c++
> 
> int _rotation_axis_idx;
> 
> ```
> 
> in the header and then list initializing it as
> 
> ```c++
> 
>     _rotation_axis_idx(int(getParam<MooseEnum>("rotation_axis")))
> 
> ```
> 
> instead of the current list intializer?

Yes - since you never actually ever use the Enum except to convert it to an integer anyway.

src/criticality/RotationSearch.C
lewisgross1296: I added a test for it now! `wrong_transform_type.i`

lewisgross1296: There will be a `TranslationSearch` one day and perhaps even more types related to `OpenMCCellTransform`. In that case, then there would need to be `isTranslation` and `is????`, so I think it might just be best to have the `getTransformType()` to allow for more types in the future

lewisgross1296: I'm happy to move this to the validation function and make the constructor a little cleaner

src/userobjects/OpenMCCellTransform.C
lewisgross1296: So far, all the criticality searches have an `updateOpenMC(const Real & var)` method which overrides the virtual version in `CriticalitySearchBase`. In this criticality search, we originally planned to use the transform to update the OpenMC model to facilitate changing reactivity via rotation as we root-find `k=1`. Since the `PostprocessorValue` data members related to the entries of `vector_value` of the `OpenMCCellTransform` class are protected, I made the setter in that class. 

IIRC I tried to have `RotationSearch` inherit from `OpenMCCellTransform`, but I think it caused an inheritance error (said something like `... ambiguous ...`). I'm not sure how to have an `updateOpenMC` method that changes these values without calling an `OpenMCCellTransform` class method without unprotecting them.

src/criticality/RotationSearch.C
gonuke: It will be good to get other perspectives.  The `MooseEnum` for this particular variable is only defined locally in `OpenMCCellTransform` so I think it's fragile & bad OO design to get back this response. 

We will certainly need `isTranslation()` but it's unlikely we'll need more than that ever.  If the types of transform ever change, it will be up to those folks to add additional methods

src/userobjects/OpenMCCellTransform.C
gonuke: You don't need to change any member variable in the `OpenMCCellTransform` (OCT) object. You need to update the PP that is referenced by the OCT.

I don't know for sure, but you should be able to get the name of that PP and interact with that PP without accessing the OCT at all.

src/criticality/RotationSearch.C
gonuke: (I won't die on this hill, but...) I think it is useful for the semantics of what constitutes validity to be communicated in the name of the method. Therefore, when reading the function call, it is immediately clear what it being checked, not just some abstract notion of validity.

I also think the check for transform type can be moved into this method so that all the validity checking is happening in one place

src/userobjects/OpenMCCellTransform.C
lewisgross1296: So I tried something simple to see what would happen. I placed these lines (similar to how we use them in `OpenMCCellTransform`) both in the constructor and the `updateOpenMC` method
```c++
  const auto & pp_name_vector = getParam<std::vector<PostprocessorName>>("vector_value");

  for(auto pn : pp_name_vector)
    std::cout << pn << std::endl;
```
and it compiles, but when I run a test case, I get errors
```
*** ERROR ***
/cardinal/test/tests/criticality/rotation/openmc.i:25.3:
The following occurred in the CriticalitySearch of type RotationSearch.

The parameter "vector_value" is being retrieved before being set.

```

gonuke: But this is different from how OCT accesses it at initialization.  The OCT captures a read-only reference/pointer to the PP  value and doesn't try to use the value until the transformation actually happens

lewisgross1296: Though, if you want to `setPostprocessorValueByName` from `RotationSearch` you need the `PostprocessorName`

include/transforms/OpenMCCellTransformBase.h
```diff
@@ -0,0 +1,64 @@
...
+{
+public:
+  static InputParameters validParams();
+
+  /// Enumeration for the transform type
+  static const MooseEnum transform_type;
+  /// The symbols for each dimension in the transform vector: ["φ", "θ", "ψ"]
+  static const std::array<std::string, 3> transform_vector_symbols;
+  /// The symbols for the transform vector in a list form: "φ, θ, ψ"
+  static const std::string transform_vector_symbols_list;
```
gonuke: These variables are only used for rotation so I recommend changing "transform" to "rotation" in the variable names & comments

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,63 @@
...
+
+void
+RotationSearch::updateOpenMCModel(const Real & angle)
+{
+  _console << "OpenMC will run with next guess for rotation = " << angle << units() << std::endl;
+
+  // make a vectorized version of the rotation angle with 0 for the non-rotating axes
+  Point angles;
+  // the enum default indices correspond to which vector component is non zero
+  angles(_rotation_axis_idx) = angle;
```
gonuke, line 45: I'm not sure about the syntax of a `Point`, but shouldn't this be
```suggestion
  angles[_rotation_axis_idx] = angle;
```

test/tests/criticality/rotation/tests
```diff
@@ -0,0 +1,47 @@
+[Tests]
```
gonuke, line 1: You can probably get rid of many of these tests with the simpler design

gonuke (changes requested):
This is certainly a lot simpler!!

I copied over one comment from the PR and noted a couple of other things.  You'll also need to update the README

src/criticality/RotationSearch.C
lewisgross1296: It compiled and ran locally, so I'm going to trust Logan

loganharbour: Yeah, a libmesh Point is weird: https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1Point.html; operator () instead of operator []

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,48 @@
...
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
```
gonuke: This was not written by ANL - this header is not required

src/criticality/RotationSearch.C
```diff
@@ -0,0 +1,72 @@
...
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
```
gonuke: Header not required

test/tests/criticality/rotation/bad_high_bound.i
```diff
@@ -0,0 +1,60 @@
+!include common_input.i
+
+[Mesh]
```
gonuke: I think we can do this with command line changes to the info in a file rather than separate files that are all similar.

See `test/tests/criticality/material_density/tests`

gonuke (changes requested):
Some high level changes

include/criticality/RotationSearch.h
lewisgross1296: Logan added this from his PR on code I did not write, so I left it as is

test/tests/criticality/rotation/bad_high_bound.i
lewisgross1296: That is a good idea. I can just overwrite the min or max from `openmc.i`, will push in next commit

test/tests/criticality/rotation/tests
```diff
@@ -0,0 +1,28 @@
...
+    expect_err = "The 'minimum' specified rotation"
+    requirement = 'The system shall error if the minmum search value passed to a RotationSearch is negative.'
+    capabilities = 'openmc'
+  []
+  [bad_upper_bound]
+    type = RunException
+    input = openmc.i
+    cli_args = 'Problem/CriticalitySearch/maximum=370.0'
+    expect_err = "The 'maximum' specified rotation"
+    requirement = 'The system shall error if the maximum search value passed to a RotationSearch is negative.'
```
gonuke:
```suggestion
    requirement = 'The system shall error if the maximum search value passed to a RotationSearch is greater than 360 [degrees].'
```

gonuke (changes requested):
One tweak to an error message in testing

test/tests/criticality/rotation/tests
lewisgross1296: Nice catch, ty

lewisgross1296:
A bit confused with the `a9d78ef` passing all tests except many OpenMC timeouts and `44a40fa` passing OpenMC but failing Nek and DAGMC (no changes to either of those in this PR AFAIK) with the [comparison](https://github.com/neams-th-coe/cardinal/compare/44a40fa9dfc61cb56ecb157cd3786a5cfd8bb29f..a9d78ef6adc5d087f086d2100a9cfd84dd298595), i.e my latest force push only being a docs formatting update

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,27 @@
+# RotationSearch
+
+## Description
+
+`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) tagerts a specified $k$-eigenvalue by updating an `openmc.Cell.rotation`, which can be applied to one or more cells in the OpenMC model.
```
gonuke:
```suggestion
`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) and targets a specified $k$-eigenvalue by updating an `openmc.Cell.rotation`, which can be applied to one or more cells in the OpenMC model. 
```

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,27 @@
+# RotationSearch
+
+## Description
+
+`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) tagerts a specified $k$-eigenvalue by updating an `openmc.Cell.rotation`, which can be applied to one or more cells in the OpenMC model. 
+
+!alert note
+The cells specified in the search must be filled with a universe. All cells included in the search will have the same rotation applied.
+
+The actual rotation is handled internally using an [OpenMCCellTransform](OpenMCCellTransform.md) UserObject, so the user does not need to (and should not) define one related to the search.
```
gonuke: I don't think this sentence is needed.  It only makes sense in the context of how you thought it was going to be done before, and everyone else it may just be confusing.  It's also not quite correct since it's not an `OpenMCCellTransform` user object that does anything.

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,27 @@
...
+
+Only axis aligned searches are allowed, so the user needs to specify which rotation axis ( `x`, `y`, or `z`) will be modified when searching for a critical configuration. The main reason for this restriction is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. 
+
+!alert tip
+If you desire to specify rotation about more than one axis to the cell(s) involved with the search, it's reccomended to instead build your model in such a way the the rotation that changes reactivity in the search is only around one principal axis.
```
gonuke: This tip may be a distraction, and my be too challenging in some cases?  I wonder if it would be better to say something like:
"For some instances of multi-axis rotation, it may be possible to build the geometry with a different orientation that only requires a single axis rotation."

doc/content/source/criticalitysearch/RotationSearch.md
gonuke: Should we mention that this is also an `OpenMCCellTransformBase` ?

gonuke (changes requested):
Some final documentation improvements.

doc/content/source/criticalitysearch/RotationSearch.md
lewisgross1296: Yeah we may as well, it's derived from both

lewisgross1296: And we probably need to add documentation for that class as well

doc/content/source/criticalitysearch/RotationSearch.md
lewisgross1296: I can demote it from tip to just general text in the docs. I added this after our discussion about "what if someone wants to specify more than one value of rotation in vector value?" but we don't even have them specify their own vector value atp

doc/content/source/criticalitysearch/RotationSearch.md
lewisgross1296: Actually it seems like the docs doesn't contain any info via markdown file for Base classes, so maybe just mentioning it without the link (to a separate docs page detailing the OpenMCCellTransformBase) is fine

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,21 @@
+# RotationSearch
+
+## Description
+
+`RotationSearch` is a [CriticalitySearch](AddCriticalitySearchAction.md) that targets a specified $k$-eigenvalue by updating an `openmc.Cell.rotation`, which can be applied to one or more cells in the OpenMC model. The `RotationSearch` class inherits from `OpenMCCellTransformBase`, which handles all modifications to the cells involved in a search.
+
+!alert note
+The cells specified in the search must be filled with a universe. All cells included in the search will have the same rotation applied.
+
+Only axis aligned searches are allowed, so the user needs to specify which rotation axis ( `x`, `y`, or `z`) will be modified when searching for a critical configuration. The main reason for this restriction is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. If your model desires to rotate around a non-principal axis or rotate around multiple axes, it is advised to instead build a version in a different reference frame that searches by rotating along one of the three available axes.
```
loganharbour: For the first sentence, it would be good to list the parameter that controls the rotation axis. Like:

```
which rotation axis ( `x`, `y`, or `z`) via the [!param](/Problem/CriticalitySearch/RotationSearch/rotation_axis) parameter ...
```

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "CriticalitySearchBase.h"
+#include "OpenMCCellTransformBase.h"
+#include "CardinalEnums.h"
+#include "UserObjectInterface.h"
+
+/**
+ * Perform a criticality search based on a rotation angle
+ */
+class RotationSearch : public CriticalitySearchBase, public OpenMCCellTransformBase
```
loganharbour, line 9: Does this need to be ifdef-ed for OpenMC?

gonuke (approved):
LGTM - thanks for all your work here @lewisgross1296

src/transforms/OpenMCCellTransformBase.C
```diff
@@ -0,0 +1,112 @@
...
+                                   " must be in the range [0, 360) degrees. Got invalid ",
+                                   rotation_vector_symbols[i],
+                                   " = ",
+                                   vec[i],
+                                   " degrees.");
+      // If a user tried to apply rotation on a cell that doesn't contain a filled universe,
+      // OpenMC will return an error.
+      err = openmc_cell_set_rotation(index, vec.data(), 3);
+      _moose_object._console << "Setting OpenMC cell rotation for cell with ID " +
+                                    std::to_string(cell_id) + " to ("
```
loganharbour:
```suggestion
      _moose_object._console << "Setting OpenMC cell rotation for cell with ID " << cell_id << " to ("
```

loganharbour (commented):
It's beautiful

include/criticality/RotationSearch.h
nuclearkevin: By Cardinal's present convention, only source files (`.C`) need the `#ifdef`.

include/transforms/OpenMCCellTransformBase.h
```diff
@@ -0,0 +1,67 @@
...
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#pragma once
+
+#ifdef ENABLE_OPENMC_COUPLING
```
nuclearkevin:
```suggestion
```
This `#ifdef` shouldn't be necessary (it's in a header file).

include/transforms/OpenMCCellTransformBase.h
```diff
@@ -0,0 +1,67 @@
...
+
+private:
+  /// The MooseObject owning this transform
+  const MooseObject & _moose_object;
+
+  /// The OpenMC problem; for calling the transforms
+  OpenMCCellAverageProblem & _openmc_problem;
+};
+
+#endif
```
nuclearkevin:
```suggestion
```
See comment above.

test/tests/criticality/rotation/tests
```diff
@@ -0,0 +1,28 @@
+[Tests]
```
nuclearkevin, line 1:
```suggestion
[Tests]
  design = 'RotationSearch.md'
```
You're missing a design document here. This is necessary for SQA purposes.

nuclearkevin (changes requested):
Just a few minor things from me at this point. Thanks for sticking through with this @lewisgross1296!

include/criticality/RotationSearch.h
loganharbour: Not sure if I agree with that. Could end up with missing symbol errors instead of compile errors. Oh well

nuclearkevin (approved):

doc/content/source/criticalitysearch/RotationSearch.md
```diff
@@ -0,0 +1,21 @@
...
+
+!alert note
+The cells specified in the search must be filled with a universe. All cells included in the search will have the same rotation applied.
+
+Only axis aligned searches are allowed, so the user needs to specify which rotation axis ( `x`, `y`, or `z`) will be modified via the [!param](/Problem/CriticalitySearch/RotationSearch/rotation_axis) parameter when searching for a critical configuration. The main reason for this restriction is that it is very easy to create an OpenMC model with voids prone to lost particles by modifying multiple rotation angles at once. If your model desires to rotate around a non-principal axis or rotate around multiple axes, it is advised to instead build a version in a different reference frame that searches by rotating along one of the three available axes.
+
+## Example Input File Syntax
+
+Here is a valid `RotationSearch` which shows the corresponding `CriticalitySearch` block used to defines the search which will happen on each iteration.
```
aprilnovak:
```suggestion
Here is a valid `RotationSearch` which shows the corresponding `CriticalitySearch` block used to define the search which will happen on each iteration.
```

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "CriticalitySearchBase.h"
+#include "OpenMCCellTransformBase.h"
+#include "CardinalEnums.h"
```
aprilnovak:
```suggestion
```

aprilnovak: I think you can get rid of this? I don't see where it is used, please correct me if I'm mistaken

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,30 @@
...
+#include "UserObjectInterface.h"
```
aprilnovak: is this also unecessary?

include/criticality/RotationSearch.h
```diff
@@ -0,0 +1,30 @@
...
+class RotationSearch : public CriticalitySearchBase, public OpenMCCellTransformBase
+{
+public:
+  static InputParameters validParams();
+
+  RotationSearch(const InputParameters & parameters);
+
+  /** Verifies that the definition of the OpenMCCellTransform
+   * UserObject is valid for a rotational criticality search
+   */
```
aprilnovak, line 18: For Doxygen comments, please add `@param[in] angle the rotation angle` for a description of the input parameter

aprilnovak (changes requested):
Thanks for the contribution @lewisgross1296 ! Looks really good - just a few minor questions/requests

lewisgross1296:
Also, realizing I forgot to report here the results of my standalone search, so I'll share for posterity. OpenMC standalone got `rotation = 82.4272 -> 1.00022+/-0.00007`
```
lgross@ulam:~/test_crit_search_for_PR  $ python grsecant_search.py -s
/home/lgross/openmc/openmc/material.py:1052: UserWarning: A uranium enrichment of 19.95 was given for Material ID="1". OpenMC assumes the U234/U235 mass ratio is constant at 0.008, which is only valid at low enrichments. Consider setting the isotopic composition manually for enrichments over 5%.
  warnings.warn(msg)
Iteration 1: batches=300, x=0, keff=0.94744+/-0.00008
Iteration 2: batches=300, x=90, keff=1.00475+/-0.00008
Iteration 3: batches=151, x=82.5475, keff=1.00051+/-0.00019
Iteration 4: batches=413, x=82.4272, keff=1.00022+/-0.00007
lgross@ulam:~/test_crit_search_for_PR  $ echo $OPENMC_CROSS_SECTIONS
/home/lgross/cross_sections/nndc_hdf5/cross_sections.xml
lgross@ulam:~/test_crit_search_for_PR  $ echo $OMP_NUM_THREADS
24
```
With the same cross-sections, particles per batch, batches, and same number of OpenMP threads, Cardinal got `rotation = 82.20667 -> 1.000072e+00 +/-  0.00008818693`
```
Running criticality search in OpenMC for Rotation in range 0.000000 - 90.000000 [degrees] 
-----------------------------------------------------------------
| Iteration | Rotation [degrees] |   k (mean)   |  k (std dev)  |
-----------------------------------------------------------------
|         0 |       0.000000e+00 | 9.474415e-01 |  8.156514e-05 |
|         1 |       9.000000e+01 | 1.004867e+00 |  9.397798e-05 |
|         2 |       8.237252e+01 | 1.000168e+00 |  9.731419e-05 |
|         3 |       8.210041e+01 | 9.999023e-01 |  8.578146e-05 |
|         4 |       8.220042e+01 | 9.999002e-01 |  8.498350e-05 |
|         5 |       8.228647e+01 | 1.000391e+00 |  8.386472e-05 |
|         6 |       8.221792e+01 | 1.000037e+00 |  9.800045e-05 |
|         7 |       8.221292e+01 | 1.000059e+00 |  8.653317e-05 |
|         8 |       8.220667e+01 | 1.000072e+00 |  8.818693e-05 |
-----------------------------------------------------------------
```
0.26% relative error seemed very good to me

I would have used `brentq` with OpenMC's `openmc.search_for_keff`, but for some reason, it was having problems with the version of OpenMC that Cardinal's contrib points to for this PR, so I used the GRSecant way that we merged in this PR (and matches the other search examples in the repo)

PR #1302, opened 2026-03-10
OpenMC initialization action

JoffreyDorville:
This PR adds a new action to initialize OpenMC.

### Context

I am using Cardinal to transfer temperatures to OpenMC not to OpenMC cells, but to a custom temperature field using a regular mesh on the OpenMC side. The multiphysics coupling seems to work well with my local prototype so I am working on cleaning the different features I have to submit them as PRs here.

### Description

I moved the initialization of OpenMC from the constructor of OpenMCProblemBase to its own action. The objective is to be able to initialize OpenMC once and potentially before the [Mesh] keyword is interpreted as I have a custom mesh generator (coming in a next PR) that requires having access to information from the XML file of OpenMC.

### Open question

I mainly copied the NekInitAction mechanism. However, I am not sure if we need to protect this action against multiple calls as I do not know if Cardinal is able to call this action more than once. Please let me know if you think that I should add a protection against that.

### DIFF
--- a/src/actions/OpenMCInitAction.C
+++ b/src/actions/OpenMCInitAction.C
@@ -45,7 +45,7 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters)
 void
 OpenMCInitAction::act()
 {
-  if (_type != "OpenMCCellAverageProblem") // TODO: add more types?
+  if (_type != "OpenMCCellAverageProblem")
     return;
 
   const auto timeStart = std::chrono::high_resolution_clock::now();
@@ -73,7 +73,7 @@ OpenMCInitAction::act()
   // Initialize OpenMC
   openmc_init(argv.size() - 1, argv.data(), &_communicator.get());
 
-  // Timer - TODO - maybe not needed
+  // Timer
   double elapsedTime = 0;
   const auto timeStop = std::chrono::high_resolution_clock::now();
   elapsedTime += std::chrono::duration<double, std::milli>(timeStop - timeStart).count() / 1e3;
--- a/src/actions/OpenMCInitAction.C
+++ b/src/actions/OpenMCInitAction.C
@@ -1,0 +1,85 @@
+/********************************************************************/
+/*                  SOFTWARE COPYRIGHT NOTIFICATION                 */
+/*                             Cardinal                             */
+/*                                                                  */
+/*                  (c) 2021 UChicago Argonne, LLC                  */
+/*                        ALL RIGHTS RESERVED                       */
+/*                                                                  */
+/*                 Prepared by UChicago Argonne, LLC                */
+/*               Under Contract No. DE-AC02-06CH11357               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*             Prepared by Battelle Energy Alliance, LLC            */
+/*               Under Contract No. DE-AC07-05ID14517               */
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "OpenMCInitAction.h"
+
+#include "MooseApp.h"
+#include <chrono>
+
+registerMooseAction("CardinalApp", OpenMCInitAction, "openmc_init");
+
+InputParameters
+OpenMCInitAction::validParams()
+{
+  InputParameters params = MooseObjectAction::validParams();
+  params.addParam<std::string>("type", "Problem type");
+  params.addParam<FileName>(
+      "xml_directory", "./", "The directory in which to look for OpenMC XML files.");
+
+  return params;
+}
+
+OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters)
+  : MooseObjectAction(parameters),
+    _xml_directory(getParam<FileName>("xml_directory"))
+{
+}
+
+void
+OpenMCInitAction::act()
+{
+  if (_type != "OpenMCCellAverageProblem") // TODO: add more types?
+    return;
+
+  const auto timeStart = std::chrono::high_resolution_clock::now();
+
+  // Suppress OpenMC output when the language server is active by
+  // decreasing the verbosity to level 1 (the lowest).
+  std::vector<std::string> argv_vec = {"openmc"};
+  if (_app.isParamValid("language_server") && _app.getParam<bool>("language_server"))
+  {
+    argv_vec.push_back("-q");
+    argv_vec.push_back("1");
+  }
+  // Add the parameter for the XML directory at the end.
+  argv_vec.push_back(_xml_directory);
+
+  std::vector<char *> argv;
+
+  for (const auto & arg : argv_vec)
+  {
+    argv.push_back(const_cast<char *>(arg.data()));
+  }
+  // Add terminating nullptr
+  argv.push_back(nullptr);
+
+  // Initialize OpenMC
+  openmc_init(argv.size() - 1, argv.data(), &_communicator.get());
+
+  // Timer - TODO - maybe not needed
+  double elapsedTime = 0;
+  const auto timeStop = std::chrono::high_resolution_clock::now();
+  elapsedTime += std::chrono::duration<double, std::milli>(timeStop - timeStart).count() / 1e3;
+
+  _console << "OpenMC initialization took " << elapsedTime << " s" << std::endl;
+
+}
+
+#endif
--- a/src/actions/OpenMCInitAction.C
+++ b/src/actions/OpenMCInitAction.C
@@ -37,8 +37,7 @@ OpenMCInitAction::validParams()
 }
 
 OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters)
-  : MooseObjectAction(parameters),
-    _xml_directory(getParam<FileName>("xml_directory"))
+  : MooseObjectAction(parameters), _xml_directory(getParam<FileName>("xml_directory"))
 {
 }
 
@@ -79,7 +78,6 @@ OpenMCInitAction::act()
   elapsedTime += std::chrono::duration<double, std::milli>(timeStop - timeStart).count() / 1e3;
 
   _console << "OpenMC initialization took " << elapsedTime << " s" << std::endl;
-
 }
 
 #endif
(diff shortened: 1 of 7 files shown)


src/base/OpenMCProblemBase.C
meltawila:
```suggestion
```

meltawila:
```suggestion
```

meltawila: Should this be removed as well?

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,83 @@
...
+#include "MooseApp.h"
+#include <chrono>
+
+registerMooseAction("CardinalApp", OpenMCInitAction, "openmc_init");
+
+InputParameters
+OpenMCInitAction::validParams()
+{
+  InputParameters params = MooseObjectAction::validParams();
+  params.addParam<std::string>("type", "Problem type");
```
meltawila: Isn't this redundant since it's already in MooseObjectAction?

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,83 @@
...
+
+  // Initialize OpenMC
+  openmc_init(argv.size() - 1, argv.data(), &_communicator.get());
+
+  // Timer
+  double elapsedTime = 0;
+  const auto timeStop = std::chrono::high_resolution_clock::now();
+  elapsedTime += std::chrono::duration<double, std::milli>(timeStop - timeStart).count() / 1e3;
+
+  _console << "OpenMC initialization took " << elapsedTime << " s" << std::endl;
```
meltawila: Shouldn't we include `OpenMCProblemBase::initialSetup()`/`OpenMCCellAverageProblem::initialSetup()` in the timing? if this is irrelevant for this PR then maybe I'd suggest making this message clearer on what is being counted here as it might get confusing, since a significant amount of extra time can be spent during these other initialization steps

meltawila (changes requested):
I only have minor comments below, and I agree we should protect against multiple calls of this action. Cardinal shouldn't initialize OpenMC more than once

src/actions/OpenMCInitAction.C
meltawila: @JoffreyDorville  Another PR (https://github.com/neams-th-coe/cardinal/pull/1339) related to this got merged recently FYI

aprilnovak: I think we can remove this console print here - the reason it's here is that we built the NekRS initailization action to try and mimic as much as possible what NekRS prints to the screen when it spins up (which includes some timings). Because OpenMC doesn't print such timings when it spins up, I think we can remove them here for simplicity.

src/actions/OpenMCInitAction.C
JoffreyDorville: Yes, it was redundant and has been removed.

src/actions/OpenMCInitAction.C
JoffreyDorville: I removed the timer in the function.

JoffreyDorville:
Thanks for the review and comments @meltawila and @aprilnovak!

I modified my original approach so that the OpenMC init action does not only depends on the Problem block. This will be helpful to extend the use of the OpenMC init action to other blocks and particularly to the Mesh block with the upcoming OpenMCMeshGenerator I am working on.

Also, the action is no longer a MooseObjectAction as it was not really necessary as pointed by @meltawila.

JoffreyDorville:
Regarding the protection against multiple calls of the action, I only rely on the way I registered the task and the action in the code which should prevent Cardinal from instantiating the OpenMC init action more than once. We could also add a flag to strictly prevent multiple calls but I am not sure if this is necessary. Please let me know what you think about it!

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,105 @@
...
+
+      if (obj_params.have_parameter<FileName>("xml_directory"))
+      {
+        xml_directory = obj_params.get<FileName>("xml_directory");
+        return true;
+      }
+      else
+      {
+        mooseError(
+            "OpenMCCellAverageProblem was found but does not have an 'xml_directory' parameter.");
```
meltawila: I'm not sure what would trigger this mooseError. `xml_directory` is an optional parameter and defaults to current directory. so if the user omits it the parameter should still exist in the object params and resolve to the current directory. If the xml files were missing I think a different error will show from OpenMC side not from this check, right? I want to suggest removing this check but please correct me if I'm wrong

meltawila:
`tutorials/pincell_multiphysics.openmc` failure is not related to this PR. I see it was failing in the moose submodule update commit from this morning. we should check this separately

src/actions/OpenMCInitAction.C
JoffreyDorville: I agree with you @meltawila! I used it during development but I realized that it was not necessary after my tests. I will remove it then.

JoffreyDorville: Done

meltawila (approved):
Looks good to me!

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,100 @@
...
+      "Initializes OpenMC when an OpenMCCellAverageProblem is present in the input.");
+  return params;
+}
+
+OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(parameters) {}
+
+void
+OpenMCInitAction::act()
+{
+  // Check if OpenMCCellAverageProblem is present and return xml_directory if it is the case
```
aprilnovak: I feel like this comment and function name are misleading (or at least they confused me for a bit). Just reading the code here, I'm led to think that `if (!get_openmc_problem_type_xml_directory(xml_directory))` is checking if a directory exists, but that's not really what this is doing - it's just checking to see if `OpenMCCellAverageProblem` exists in the input file. Can we delete the `get_openmc_problem_type_xml_directory` function and just put the code content into here instead?

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,100 @@
...
+  std::string xml_directory;
+  if (!get_openmc_problem_type_xml_directory(xml_directory))
+    return;
+
+  init_openmc(xml_directory);
+}
+
+bool
+OpenMCInitAction::get_openmc_problem_type_xml_directory(std::string & xml_directory) const
```
aprilnovak: In Cardinal (and all MOOSE applications), the convention is to use camel case (beginning with a lowercase letter) for all class methods

aprilnovak: `getOpenMCProblemTypeXMLDirectory`

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,100 @@
...
+
+  for (const auto * action : problem_actions)
+  {
+    // If an action has the OpenMCCellAverageProblem type
+    if (action->getParam<std::string>("type") == "OpenMCCellAverageProblem")
+    {
+      // Retrieve the associated xml_directory_parameter
+      const InputParameters & obj_params = action->getObjectParams();
+
+      if (obj_params.have_parameter<FileName>("xml_directory"))
```
aprilnovak: Does this check need to exist? I think `xml_directory` has a default value no matter what the user does in the input file, so I think this if-statement could just be replaced by simpler code

src/actions/OpenMCInitAction.C
```diff
@@ -0,0 +1,100 @@
...
+        xml_directory = obj_params.get<FileName>("xml_directory");
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+void
+OpenMCInitAction::init_openmc(const std::string & xml_directory)
```
aprilnovak: same comment about camel case for functions, this can be `initOpenMC`

JoffreyDorville: Changed for `initOpenMC`

src/actions/OpenMCInitAction.C
JoffreyDorville: This test is not necessary. I just removed the if-statement.

src/actions/OpenMCInitAction.C
JoffreyDorville: Changed for `isOpenMCCellAverageProblemRequested`

src/actions/OpenMCInitAction.C
JoffreyDorville: I agree with you that the function name and documentation are misleading. I changed it to `isOpenMCCellAverageProblemRequested`.  The reason I would like to keep it as a separate function is that in my next PR, there is another function that performs a similar action but for mesh generators. Separating the code into two functions is clearer in that anticipated context, but I agree that it seems a bit too much for the context of this PR.

JoffreyDorville:
Thanks for the review @aprilnovak !

I just updated the code based on your feedback.

Is it possible that a user would want to declare more than one `OpenMCCellAverageProblem`, via multiapps for example?

With the current state of this PR, if there are multiple occurrences of `OpenMCCellAverageProblem` in the actions warehouse we just consider the first occurrence and retrieve the corresponding `xml_directory`. By just reading the code, I believe that the current code base without this PR will fail if there are more than one `OpenMCCellAverageProblem`  because it would attempt to initialize OpenMC more than once, but I have not tested it. The current PR will not fail in this case and will also not consider a different `xml_directory` than the first one it found.

Do we want to put a safeguard  in the current PR to avoid the case where multiple `OpenMCCellAverageProblem` are requested? Or we can also make a check on all retrieved `xml_directory` values so that we accept to initialize OpenMC as long as they are identical (this is something I am doing for the following PR with the `OpenMCMeshGenerator`).

aprilnovak:
Thanks @JoffreyDorville! Cardinal will not work properly if there's more than one `OpenMCCellAverageProblem` across multiapps - OpenMC uses some global variables that get scrambled if more than one OpenMC instance is initialized in memory. If `OpenMCCellAverageProblem` shows up more than once in a single input file, MOOSE merges those blocks - and then throws an error because `type = OpenMCCellAverageProblem` is specified more than once in the same block. 

We don't currently check for the multi-app scenario but the code will fail and prevent you from going further (although perhaps not with a useful error message to debug what went wrong).

aprilnovak:
I am good to merge this, thanks! @meltawila I take it you're separately looking into the test failure in the tutorials and the apptainer issue?

meltawila:
@aprilnovak yep!

Issue #1312, opened 2026-04-01
Enable output of all statepoint files in transient OpenMC solves.


TheBEllis:
## Reason
Currently, when performing transient OpenMC solves, the statepoint file output at each timestep will overwrite the one written out at the previous timestep. In some situations, for example shutdown dose rate calculations, data from multiple timesteps need to be kept of further analysis.  

## Design
Could potentially use edit `openmc::settings::path_output` on each timestep to force statepoint files to be written to different directories, preventing then being overwritten. This is potentially easier than changing the filenames of the statepoints, as `openmc_statepoint_write(std:string filename)` is called within `finalize_batch()`, so it's difficult to pass a different filename parameter without making changes to more fundamental parts of OpenMC.

Whether or not the statepoints from each timestep are kept should be set via a parameter to OpenMCProblemBase.

## Impact
This change will enable users to keep statepoint files from all timesteps, and more generally allow users to set `openmc::settings::path_output`. 

This feature could be extended to allow for specific timesteps to be output instead of all of them.

PR #1313, opened 2026-04-01
Added parameters to keep all statepoints in transient solves. 

TheBEllis:
Addresses issue #1312. Adds two parameters to OpenMCProblemBase, `statepoint_directory` and `keep_transient_statepoint`. 

`statepoint_directory` allows the user to set the directory they want the statepoint file to be written to. In `OpenMCProblemBase` this sets `openmc::settings::path_output`. 

`keep_transient_statepoint` creates a directory for each timestep's statepoint file to be saved to. If the user has already set `statepoint_directory`, this value is used as a base directory name, and a suffix is appended to indicate the timestep. 

I've added a couple of tests in, just to check the statepoints get output to the expected directories.


aprilnovak:
@TheBEllis can you rebase your branch off `devel`? It looks like there's a bunch of other commits coming along for the ride.

TheBEllis:
Whoops my bad, should be rebased now!

include/base/OpenMCProblemBase.h
```diff
@@ -445,6 +445,10 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
            std::to_string(_fixed_point_iteration) + ".h5";
   }
 
+  const std::string transientStatepointPath();
```
nuclearkevin, line 485: This function needs a doxygen comment. Something like the following should work.

```c++
/**
 * DESCRIPTION OF FUNCTION
 * @return DESCRIPTION OF RETURN VALUE
 */
```

include/base/OpenMCProblemBase.h
```diff
@@ -445,6 +445,10 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
...
+
+  const std::string formattedOutputPath(const std::string & path_output);
```
nuclearkevin: Also needs a doxygen comment.

include/base/OpenMCProblemBase.h
```diff
@@ -567,6 +571,12 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
   /// Directory in which OpenMC settings xml files are located
   const std::string & _xml_directory;
 
+  /// Directory to write statepoint file to
+  const std::string & _statepoint_directory;
+
+  /// Parameter determines whether statepoints from all timesteps should be saved in separtate directories to avoid them being overwritten
+  bool _keep_transient_statepoint;
```
nuclearkevin: I believe this should be marked as constant (it isn't modified anywhere else).
```suggestion
  const bool _keep_transient_statepoint;
```

src/base/OpenMCProblemBase.C
```diff
@@ -16,6 +16,8 @@
 /*                 See LICENSE for full restrictions                */
 /********************************************************************/
 
+#include "openmc/settings.h"
```
nuclearkevin: All includes need to be inside the `#ifdef ENABLE_OPENMC_COUPLING` (especially OpenMC includes) to avoid build failures when users only run with NekRS. Applies to line 20 as well.

src/base/OpenMCProblemBase.C
```diff
@@ -233,6 +248,15 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
     catchOpenMCError(err, "set the number of batches");
   }
 
+  if (isParamSetByUser("statepoint_directory") && !_keep_transient_statepoint)
+  {
+    openmc::settings::path_output = formattedOutputPath(_statepoint_directory);
+    if (!std::filesystem::is_directory(openmc::settings::path_output))
+    {
```
nuclearkevin:
```suggestion
```

Simple single line if statements don't need braces (MOOSE style guide).

src/base/OpenMCProblemBase.C
```diff
@@ -233,6 +248,15 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
...
+      std::filesystem::create_directory(openmc::settings::path_output);
+    }
```
nuclearkevin:
```suggestion
```

See above comment.

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+  // Get path of current input file
+  std::filesystem::path running_path = std::filesystem::absolute(
+      std::filesystem::path(getMooseApp().getLastInputFileName()).parent_path());
+
+  std::filesystem::path transient_statepoint_path;
+
+  // If user has not set statepoint_directory parameter, or has defined it as './',
+  // use a default
+  if (std::filesystem::equivalent(_statepoint_directory, running_path))
+  {
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+    transient_statepoint_path = "./statepoint_folder";
+  }
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+  else
+  {
+
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+    // Removes trailing "/" from transient_statepoint_path, if user has left any, ready to append
+    // suffix
+    transient_statepoint_path = transient_statepoint_path.filename().empty()
+                                    ? transient_statepoint_path.parent_path()
+                                    : transient_statepoint_path;
+  }
+
+  std::string timestep_suffix = "_ts_" + std::to_string(timeStep()) + "/";
+
+  // std::string transient_statepoint_path_str = transient_statepoint_path.string() +
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+  // timestep_suffix;
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+  if (p.is_relative())
+  {
+    std::filesystem::path input_file_path = std::filesystem::absolute(
+        std::filesystem::path(getMooseApp().getLastInputFileName()).parent_path());
+
+    p = std::filesystem::weakly_canonical(input_file_path / p);
+  }
+
+  if (p.string().back() != '/')
+  {
```
nuclearkevin:
```suggestion
```

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1060,72 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+    p += "/";
+  }
```
nuclearkevin:
```suggestion
```

test/tests/neutronics/openmc_statepoint_transient/openmc.i
```diff
@@ -0,0 +1,27 @@
...
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  keep_transient_statepoint = true
+[]
+
+[Executioner]
+  type = Transient
+  num_steps = 2 
+  batches = 50 # this number needs to match the value in settings.xml for the purpose of the test
```
nuclearkevin: `batches` belongs in the `[Problem]` block.

test/tests/neutronics/openmc_statepoint_transient/openmc_user_dir.i
```diff
@@ -0,0 +1,28 @@
...
+[Problem]
+  type = OpenMCCellAverageProblem
+  keep_transient_statepoint = true
+  statepoint_directory = "./mystatepoints/"
+[]
+
+[Executioner]
+  type = Transient
+  num_steps = 2 
+  batches = 50 # this number needs to match the value in settings.xml for the purpose of the test
```
nuclearkevin: `batches` belongs in the `[Problem]` block.

include/base/OpenMCProblemBase.h
```diff
@@ -567,6 +571,12 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
   /// Directory in which OpenMC settings xml files are located
   const std::string & _xml_directory;
 
+  /// Directory to write statepoint file to
+  const std::string & _statepoint_directory;
+
+  /// Parameter determines whether statepoints from all timesteps should be saved in separtate directories to avoid them being overwritten
```
nuclearkevin, line 628: Is there any way we can modify the statepoint file name instead of creating a new directory for every unique statepoint? Creating a collection of folders, each with a single file, feels rather cumbersome compared to adding a suffix to the statepoint file...

src/base/OpenMCProblemBase.C
```diff
@@ -233,6 +248,15 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
     catchOpenMCError(err, "set the number of batches");
   }
 
+  if (isParamSetByUser("statepoint_directory") && !_keep_transient_statepoint)
+  {
+    openmc::settings::path_output = formattedOutputPath(_statepoint_directory);
+    if (!std::filesystem::is_directory(openmc::settings::path_output))
```
nuclearkevin: Are we catching this if statement with a test? From what I can see, none of the tests try to create a statepoint directory with the same name as a file.

nuclearkevin (changes requested):
Thanks for opening this PR @TheBEllis! I have one question regarding the design and I think another test should be added to catch an edge case, but otherwise I don't have any major complaints.

src/base/OpenMCProblemBase.C
TheBEllis: Apologies, my lsp setup is extremely liberal when it comes to automatically inserting includes

include/base/OpenMCProblemBase.h
TheBEllis: I completely agree this would be a far cleaner method, and `openmc_statepoint_write(const char* filename, bool* write_source)` does accept a filename parameter. However, this method gets called in `finalize_batch()` at line 440 of `simulation.cpp`, making passing this parameter in difficult without underlying modifications to openmc.

nuclearkevin:
It appears as if the include statements are still outside of the `#ifdef ENABLE_OPENMC` guard, which is why the NekRS test is failing. 

~~I don't know off the top of my head where the cascading OpenMC/DAGMC test failures are coming from.~~ The test timeouts appear to be an issue on the CIVET side of things as they're common to all PRs.

src/base/OpenMCProblemBase.C
TheBEllis: I've added a RunException for this exact scenario, let me know if you think that's enough or whether I should be catching it in code:)

nuclearkevin: This works for me!

src/base/OpenMCProblemBase.C
```diff
@@ -16,6 +16,7 @@
 /*                 See LICENSE for full restrictions                */
 /********************************************************************/
 
+#include <filesystem>
```
nuclearkevin: This should also be inside the `#ifdef ENABLE_OPENMC_COUPLING`.

src/base/OpenMCProblemBase.C
```diff
@@ -1030,4 +1058,69 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
...
+      formattedOutputPath(transient_statepoint_path.string());
+
+  return transient_statepoint_path_str;
+}
+
+const std::string
+OpenMCProblemBase::formattedOutputPath(const std::string & output_path)
+{
+  std::filesystem::path p = output_path;
+  _console << p.string() << std::endl;
```
nuclearkevin: Why is this getting printed to console?

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -14,4 +14,28 @@
...
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
+    # in the fission bank on a process
+    max_parallel = 8
+    requirement = "A statepoint file must be written to the parameter defined directory when the batches parameter matches the batches "
+                  "in settings.xml and they are both specified."
+    capabilities = 'openmc'
+  []
+
+
+  [check_statepoint_exists_in_directory_w_existing_filenmae]
```
nuclearkevin:
```suggestion
  [check_statepoint_exists_in_directory_w_existing_filename]
```

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -14,4 +14,28 @@
...
+    type = RunException
+    input = openmc_custom_dir_filename.i
+    expect_err = 'filesystem error: cannot create directory: File exists'
```
nuclearkevin: Does this trace back to `OpenMCCellAverageProblem` through a `mooseError()`? if not, it will be difficult for the user to debug the error message.

nuclearkevin (changes requested):
A few more comments from me, but I think this is almost ready to go.

Terribly sorry for the inconvenience dealing with the regression suite being non-deterministic - we're trying to figure out what's going on but can't find any patterns.

src/base/OpenMCProblemBase.C
TheBEllis: Clangd is getting a stern word from me right now. Apologies for this coming up again

test/tests/neutronics/openmc_statepoint/tests
TheBEllis: I've moved the logic around a bit so that a `mooseError()` is thrown now. There are some slightly annoying lines in there, mostly related to removing trailing "/". The value passed to `openmc::settings::path_output` must contain a trailing "/", but when checking is a file with the same name exists using `is_regular_file()`, this has to be removed.

src/base/OpenMCProblemBase.C
TheBEllis: This has been removed

nuclearkevin (approved):
LGTM!

nuclearkevin:
I'll rerun the tests until they pass. The random `RunException` failures are unrelated.

include/base/OpenMCProblemBase.h
```diff
@@ -567,6 +581,12 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
...
+  const bool _keep_transient_statepoint;
```
aprilnovak:
```suggestion
  const bool & _keep_transient_statepoint;
```

aprilnovak: capturing by reference is recommended by MOOSE convention when possible because it'll be compatible with controls and other systems

src/base/OpenMCProblemBase.C
```diff
@@ -98,6 +98,17 @@ OpenMCProblemBase::validParams()
...
+
+  params.addParam<FileName>(
+      "statepoint_directory",
+      "./",
+      "The directory to write statepoint files to. Sets openmc::settings::path_output.");
+
+  params.addParam<bool>("keep_transient_statepoint",
+                        false,
+                        "Whether or not statepoints from all timesteps should be kept, and written "
+                        "to seperate directories.");
```
aprilnovak:
```suggestion
                        "to separate directories.");
```

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -14,4 +14,28 @@
...
+    type = CheckFiles
+    input = openmc_custom_dir.i
+    check_files = 'custom_dir/statepoint.50.h5'
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
+    # in the fission bank on a process
+    max_parallel = 8
+    requirement = "A statepoint file must be written to the parameter defined directory when the batches parameter matches the batches "
+                  "in settings.xml and they are both specified."
+    capabilities = 'openmc'
+  []
```
aprilnovak, line 32: please make sure each test has a `issues` and `design` parameter

src/base/OpenMCProblemBase.C
```diff
@@ -233,6 +246,22 @@ OpenMCProblemBase::OpenMCProblemBase(const InputParameters & params)
...
+  if (isParamSetByUser("statepoint_directory") && !_keep_transient_statepoint)
+  {
+    /// path_output must end with a "/", otherwise statepoint will not output correctly
+    openmc::settings::path_output = formattedOutputPath(_statepoint_directory);
+
+    /// Need to remove trailing "/" to do "is_regular_file"
+    std::filesystem::path p = openmc::settings::path_output;
+    p = p.filename().empty() ? p.parent_path() : p;
+
+    if (std::filesystem::is_regular_file(p))
```
aprilnovak, line 268: Apologies for the long delay in reviewing this @TheBEllis ! Things were busy with Physor last month and end of semester for us.

I am curious about the reason why we need to throw a hard error here - when we run OpenMC as usual, the output files are overwritten by the most recent run. Could that not also be done here, such that if the directory already exists we just proceed as normal and overwrite whatever is contained within? In other words, create the directory if it does not exist, but if it already exists then just allow OpenMC to overwrite itself as usual?

Or is there a design reason why we would need to manually delete the directory each time we'd run the Cardinal model?

src/base/OpenMCProblemBase.C
```diff
@@ -375,6 +404,19 @@ OpenMCProblemBase::externalSolve()
   // update tallies as needed before starting the OpenMC run
   executeEditors();
 
+  if (_keep_transient_statepoint)
+  {
+    openmc::settings::path_output = transientStatepointPath();
+
+    /// Need to remove trailing "/" to do "is_regular_file"
```
aprilnovak: Could this chunk of code be encapsulated in a function? I think it is repeated above in the other if-statement in the constructor.

test/tests/neutronics/openmc_statepoint_transient/tests
```diff
@@ -0,0 +1,30 @@
...
+                  "in settings.xml and they are both specified."
+    capabilities = 'openmc'
+  []
+
+  [check_statepoint_exists_custom_dir]
+    type = CheckFiles
+    input = openmc_user_dir.i
+    issues = '#1312'
+    design = 'OpenMCCellAverageProblem.md'
+    check_files = 'mystatepoints_ts_1/statepoint.50.h5 mystatepoints_ts_2/statepoint.50.h5'
```
aprilnovak, line 23:
```suggestion
    check_files = 'mystatepoints_ts_1/statepoint.50.h5 mystatepoints_ts_2/statepoint.50.h5'
    cli_args = 'Problem/statepoint_directory="./mystatepoints/"'
```

test/tests/neutronics/openmc_statepoint_transient/tests
```diff
@@ -0,0 +1,30 @@
...
+    # in the fission bank on a process
+    max_parallel = 8
+    requirement = "A statepoint file must be written for each timestep into the correct directory when the batches parameter matches the batches "
+                  "in settings.xml and they are both specified."
+    capabilities = 'openmc'
+  []
+
+  [check_statepoint_exists_custom_dir]
+    type = CheckFiles
+    input = openmc_user_dir.i
```
aprilnovak:
```suggestion
    input = openmc.i
```

test/tests/neutronics/openmc_statepoint_transient/tests
aprilnovak: if there's only a few lines of diffs between test files, we prefer to pass in parameters on the command line with `cli_args` rather than creating a new file itself. This helps lower technical debt by helping reduce number of syntax changes in the input files when names/parameters change. With the changes I've proposed here, we can delete now the `openmc_user_dir.i` file.

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -14,4 +14,32 @@
...
     capabilities = 'openmc'
   []
+
+  [check_statepoint_exists_in_directory]
+    type = CheckFiles
+    input = openmc_custom_dir.i
+    issues = '#1312'
+    design = 'OpenMCCellAverageProblem.md'
+    check_files = 'custom_dir/statepoint.50.h5'
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
```
aprilnovak, line 26:
```suggestion
    cli_args = 'Problem/statepoint_directory="./existing_file"'
    # This test has very few particles, and OpenMC will error if there aren't enough source particles
```

aprilnovak: same idea here, we can now delete `openmc_custom_dir.i`

test/tests/neutronics/openmc_statepoint/tests
```diff
@@ -14,4 +14,32 @@
                   "in settings.xml and they are both specified."
     capabilities = 'openmc'
   []
+
+  [check_statepoint_exists_in_directory]
+    type = CheckFiles
+    input = openmc_custom_dir.i
```
aprilnovak:
```suggestion
    input = openmc.i
```

src/base/OpenMCProblemBase.C
TheBEllis: No need to apologise at all @aprilnovak! Especially given it has taken me two weeks to get back to any of your comments.

The current implementation functions as you suggest (I believe?). If a directory with the user supplied name already exists, that's fine and we just overwrite whatever statepoint file is written inside the directory. The error is only thrown if the user asks a directory to be made that has the same name as an existing file.

I put this in just because I could imagine it being frustrating if a user was to accidentally delete their results file because they overwrote it with a directory of the same name. I don't know how or why this would happen, given that  I'd hope most users wouldn't name a directory "results.e", but you never know. If you don't think it necessary I can get rid:)

test/tests/neutronics/openmc_statepoint/tests
TheBEllis: I believe I have addressed all the tests I added

test/tests/neutronics/openmc_statepoint_transient/tests
TheBEllis: This makes a lot more sense that what I was doing

test/tests/neutronics/openmc_statepoint_transient/tests
```diff
@@ -0,0 +1,31 @@
...
+    type = CheckFiles
+    input = openmc.i
+    issues = '#1312'
+    design = 'OpenMCCellAverageProblem.md'
+    check_files = 'mystatepoints_ts_1/statepoint.50.h5 mystatepoints_ts_2/statepoint.50.h5'
+    cli_args = 'Problem/statepoint_directory="./mystatepoints/"'
+    # This test has very few particles, and OpenMC will error if there aren't enough source particles
+    # in the fission bank on a process
+    max_parallel = 8
+    requirement = "A statepoint file must be written for each timestep into the correct user defined directory."
```
aprilnovak, line 28: Just so I'm clear, what is the indended behavior here? When I run this test, it seems to be rather just changing the name of the statepoint directories, rather than placing each statepoint directory inside some other top-level directory.

```
(openmc) ajnovak2@ajnovak2-serv-01:~/may27b/cardinal/test/tests/neutronics/openmc_statepoint_transient$ ls
geometry.xml        mystatepoints_ts_2  settings.xml            summary.h5
materials.xml       openmc.i            statepoint_folder_ts_1  tests
mystatepoints_ts_1  openmc_out.csv      statepoint_folder_ts_2
```

From the wording here though, I would have expected we'd be creating a folder called `mystatepoints` which would then hold inside it `statepoint_folder_ts_1`, etc.

src/base/OpenMCProblemBase.C
aprilnovak: Ah, I see it now! Yes, that behaves like how I'd expect - thanks for clarifying

test/tests/neutronics/openmc_statepoint_transient/tests
TheBEllis: Sorry for the delay April, I had some leave booked in! So it is currently working as intended, my idea was to give the user an option to replace the default "statepoint_folder" prefix for the directory naming. This may be slightly unnecessary, I can't imagine too many scenarios where the user would need a custom name for the directory, but I thought it worth adding. Let me know if you think it should function differently.

src/base/OpenMCProblemBase.C
TheBEllis: I have added a `changeDirectoryToFile` function in the latest commit!

test/tests/neutronics/openmc_statepoint_transient/tests
aprilnovak: Thanks for the explanation! My preference would be to put the files nested inside a new directory - the docstring for `statepoint_directory` is "The directory to write statepoint files to" -- but that's not actually how the parameter is used, it's more like a prefix. So, I'd suggest either of the following (I prefer the first one, but if you prefer the second one that's ok by me!)

- Change the use of `statepoint_directory` to be consistent with the docstring and how OpenMC uses `output_path`, e.g. it would create a new directory named `my_custom_directory` and put all the statepoints inside there.
- Change `statepoint_directory` to `statepoint_prefix` parameter name and remove reference to `output_path` in the docstring.

aprilnovak:
Hi @TheBEllis, if you'd like I can tidy up this branch so it can get merged? I'm sure you are busy at the moment with the Physor NSE paper.

If you'd like me to help, can you give me permission to push to your `statepoint_timestamps` branch?

TheBEllis:
Hi April, sorry this has taken so long, you were right about me being busy with the PHYSOR NSE paper! I have changed the parameter name as mentioned in my reply to your comment, but I have foolishly synced my fork instead of doing a rebase so now there is a big merge commit, I will let the tests run but I need to fix that before merging!

test/tests/neutronics/openmc_statepoint_transient/tests
TheBEllis: I agree option 1 would be better, but I've unfortunately had to go with option 2. Option 1 is currently not viable due to all the statepoint files having the same name, so in a transient case they would all overwrite each other. While `openmc_statpoint_write` does have a filename parameter, it is called in finalize_batch, which itself doesn't accept an input filename parameter and instead passes a nullptr, prompting every statepoint file to have the default name. To change this behaviour would therefore require an OpenMC PR, so for now I'm going with the easy route:)

PR #1319, opened 2026-04-04
Allow for the use of relaxation with a `MeshTally` and the `MoabSkinner`

### LINKED ISSUE #1318: Relaxation does not work with mesh tallies when using the `MoabSkinner`
## Bug Description
Changes to Cardinal (namely the addition of the tally system) resulted in a regression where unstructured mesh tallies are not longer able to use relaxation when the `MoabSkinner` is active. Cell tallies should not be allowed to use relaxation when skinning is active (as cell tally bins change when the geometry is skinned); mesh tally bins do not run into the same issue.

Aside from blocking a relaxation use case, this regression broke the MSFR input deck (which we don't test due to the size of the NekRS case). 

## Steps to Reproduce
Run the OpenMC portion of the MSFR tutorial. 

## Impact
Broken tutorial, inability to use relaxation with mesh tallies and skinning.

nuclearkevin:
I also took the liberty of tagging a bunch of relaxation and `MoabSkinner` tests for SQA purposes.

Closes #1318


nuclearkevin:
It appears as if Cardinal tests using `RunException`s have a chance of randomly timing out. I'll ask some MOOSE folk if they know what the cause could be - doesn't seem to be impacting the tests in any other MOOSE apps or the framework proper.

nuclearkevin:
@aprilnovak got the tests to pass. This should be ready for a review now.

Unfortunately, Guillaume has no idea what's going on with the random `RunException` test failures. It doesn't appear to be related to particular machines on CIVET. Cardinal itself isn't hanging when exiting (at least, not from what I've seen when running in gdb & valgrind). Current working theory is it has something to do with our MOOSE submodule, or the CIVET recipe.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -275,7 +275,7 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
   // guarantee that the tallies from iteration to iteration correspond to exactly
   // the same number of bins or to exactly the same regions of space, so we must
   // disable relaxation.
-  if (_need_to_reinit_coupling && _relaxation != relaxation::none)
+  if ((_use_displaced || _has_adaptivity) && _relaxation != relaxation::none)
```
aprilnovak, line 279: is this the right way to approach it? we would still want to avoid relaxation when using the skinner, correct? (using the skinner is encompassed in `_need_to_reinit_coupling`). Why not add `_has_adaptivity` to `_need_to_reinit_coupling`?

nuclearkevin: We can use relaxation with the skinner, but only when a non-adaptive mesh tally is being used. Banning relaxation outright when skinning resulted in the regression this PR fixes (breaking the MSFR tutorial). That's the behaviour we get here if I swap this for `_need_to_reinit_coupling`.

aprilnovak: oh I see - ok, all these booleans and the venn diagrams was confusing me

nuclearkevin: Yeah...

As we've added more and more features to Cardinal, the control flow has gotten progressively more difficult to parse. Probably worth it to do a general code review of the OpenMC problems at some point.

src/base/OpenMCCellAverageProblem.C
aprilnovak, line 195:
```suggestion
    _need_to_reinit_coupling(_has_adaptivity || _using_skinner /* use_displaced also added to this condition later in constructor */),
```

aprilnovak: I feel like this helps me read the code, because I kept coming back up here and forgetting why use_displaced was not showing up (it just does, later in the constructor). Perhaps `_need_to_reinit_coupling` doesn't really belong in the member initialization list since it's not a const or null/void thing.

test/tests/neutronics/dagmc/cell_tallies/with_relaxation.i
```diff
@@ -0,0 +1,58 @@
+[Mesh]
+  [file]
+    type = FileMeshGenerator
+    file = ../mesh_tallies/slab.e
+  []
+  allow_renumbering = false
+[]
+
+[Functions]
```
aprilnovak: Can we just set temperature to a constant value? Do we need to set it to a function to reach the desired test failure? If not, I suggest we just do

```
[ICs]
  [temp]
    type = ConstantIC
    variable = temp
    value = 500.0
[]
```

test/tests/neutronics/dagmc/mesh_tallies/tests
```diff
@@ -60,3 +63,17 @@
...
 []
+
+[Tests]
+  issues = '#1318'
+  design = 'MeshTally.md OpenMCCellAverageProblem.md'
+
+  [mesh_tally_relaxation_skin]
+    type = CSVDiff
+    input = with_relaxation.i
+    csvdiff = with_relaxation_out.csv
```
aprilnovak, line 74: do the results in the CSV file actually capture that relaxation was utilized? wouldn't the tally average value be the same regardless of whether relaxation is on/off? If I'm not misunderstanding, I suggest we make that postprocessor a `PointValue` or something which would trigger different values if for some reason relaxation stopped working silently

test/tests/neutronics/dagmc/mesh_tallies/with_relaxation.i
```diff
@@ -0,0 +1,84 @@
+[Mesh]
+  [file]
+    type = FileMeshGenerator
+    file = slab.e
+  []
+[]
+
+[AuxVariables]
```
aprilnovak: Does `cell_temperature` need to exist here or can it be removed? simpler = better :)

test/tests/neutronics/dagmc/mesh_tallies/with_relaxation.i
```diff
@@ -0,0 +1,84 @@
...
+[AuxKernels]
+  [cell_temperature]
+    type = CellTemperatureAux
+    variable = cell_temperature
+  []
+[]
+
+[Functions]
+  [temp]
+    type = ParsedFunction
```
aprilnovak, line 10: Also for this test, the tally only changes between two Picard iterations from statistical noise, not from the changing bins themselves. That's probably fine to still view the effect of relaxation, but in case we wanted the response to be stronger we could make this a function of time

aprilnovak:
Thanks for tackling this fix @nuclearkevin!

test/tests/neutronics/dagmc/mesh_tallies/tests
nuclearkevin: Good point. I was originally coming at this from the standpoint of making sure that the error message isn't thrown as opposed to checking that relaxation is actually being applied. I'll swap over to a `PointValue` post-processor.

test/tests/neutronics/dagmc/mesh_tallies/with_relaxation.i
nuclearkevin: Nope!

test/tests/neutronics/dagmc/mesh_tallies/tests
nuclearkevin: I've added kappa_fission_raw to the gold file as well to ensure relaxation is being applied.

test/tests/neutronics/dagmc/mesh_tallies/with_relaxation.i
nuclearkevin: After adding the raw kappa_fission tally value and the relaxed tally value to the gold file, it's obvious that relaxation is being applied.

PR #1330, opened 2026-05-06
Update notion of tolerance for CriticalitySearch

### LINKED ISSUE #1323: Notion of tolerance in criticality search does not match up with BrentsMethod
## Bug Description

After doing some criticality searches of my own, I'm fairly certain that our notion of tolerance in Cardinal does not match up with the way tolerance is used in MOOSE's [BrentsMethod::Root](https://github.com/idaholab/moose/blob/6c007fc62d6f6bffb5a62b809f8c4c06fa8f660a/modules/fluid_properties/src/utils/BrentsMethod.C#L66). While not technically causing any errors, I believe this should be classified as a bug, as users will not really be getting outputs in line with expectations based on the current documentation for certain choices of tolerance, dependent on the actual worth curves of their model.

The tolerance that BrentsMethod uses is on the root `r` itself, not on `f(r)`. So currently in the documentation, we specify that the search tolerance is on `k(r)`m but it's actually on `r`. Here's an example `CriticalitySearch` output where I specify a tolerance of `0.01` for a rotating drum
```txt
-----------------------------------------------------------------
| Iteration | Rotation [degrees] |   k (mean)   |  k (std dev)  |
-----------------------------------------------------------------
|         0 |       0.000000e+00 | 9.727946e-01 |  3.900132e-04 |
|         1 |       1.800000e+02 | 1.033793e+00 |  3.863450e-04 |
|         2 |       8.028030e+01 | 1.000894e+00 |  4.037180e-04 |
|         3 |       7.772526e+01 | 1.000369e+00 |  4.217397e-04 |
|         4 |       7.595457e+01 | 9.999968e-01 |  4.117775e-04 |
|         5 |       7.596970e+01 | 9.999806e-01 |  3.488482e-04 |
|         6 |       7.684748e+01 | 9.992673e-01 |  4.235318e-04 |
|         7 |       7.743126e+01 | 1.000430e+00 |  4.350114e-04 |
|         8 |       7.713937e+01 | 9.998898e-01 |  4.351874e-04 |
|         9 |       7.719894e+01 | 9.999999e-01 |  4.283253e-04 |
|        10 |       7.720394e+01 | 1.000605e+00 |  4.143592e-04 |
-----------------------------------------------------------------
```
It's clear that after a few iterations, the `k` is well within an absolute tolerance of 0.01 of it's target `k=1`, but the search continues until `Rotation [degrees]` is within this tolerance with the last guess. The good news is that the root finding is returning eigenvalues that are very close to the target, but it's not really possible to specify a tolerance on the actual eigenvalue directly if we're using `BrentsMethod`.

There's two places in Cardinal that I believe need to be changed to reflect this. We do actually check that `abs(k-k_target)<tol` after the search is done, but I believe this check should be removed because it's possible that `BrentsMethod` converges the root within the tolerance but not `k`. This could be the case in which the slope of our function is very steep. For case with a very steep function near the root, the user would simply need a stricter tolerance on the root to get `k` closer to the target.
https://github.com/neams-th-coe/cardinal/blob/044a9336920389ac8f57de1640b462ef44ad1c86/src/criticality/CriticalitySearchBase.C#L120-L123

Also, inside our generic `func = [&](Real x)` method in `src/criticality/CriticalitySearchBase.C`, we check to make sure the tolerance is not too tight compared to the actual tally statistics. As I've said, this comparison is a comparison of the root tolerance with the function eval tolerance, so I think that is a pathway to trouble.
https://github.com/neams-th-coe/cardinal/blob/044a9336920389ac8f57de1640b462ef44ad1c86/src/criticality/CriticalitySearchBase.C#L108-L113

I think it would be okay to modify this check by adding a new parameter for the user to specify that they want to confirm their uncertainty in `k` is less than a certain amount/throw a warning to use more particles if that ends up not being the case. We could also live without it, and it could be up to the user to specify enough particles to get a meaningful result.

## Steps to Reproduce

Not really necessary, but you can verify with any Cardinal criticality search test by changing the tolerance and seeing the effect on the search.

## Impact

The impact is just that the docs are misleading. The actual outputs probably would give most people what they want, but depending on certain problems, the search results might be confusing if the user assumes tolerance is on k.

I will issue a PR to address this. Tagging @gonuke

lewisgross1296:
Now that we know that the tolerance passed to `BrentsMethod` is actually imposed on the root and not the eigenvalue, we should require the user to specify a `root_tol` and `k_tol` to make sure the search satisfies their needs. Closes #1323.

Updated test syntax (converted all `tolerance` -> `k_tol` and tried to give reasonable values for `root_tol`). I'm working on getting a Docker image with these changes so I can test the searches locally and confirm, but it's giving me a bit of trouble.

@aprilnovak please let me know when the new docs are live. I plan to tackle docs and my build issues tomorrow.


aprilnovak:
working on it! fyi, you can build docs locally if you need to render them.

src/criticality/CriticalitySearchBase.C
```diff
@@ -36,10 +36,12 @@ CriticalitySearchBase::validParams()
       "Maximum for values to search over; the root must occur at a value smaller than the maximum");
   params.addRangeCheckedParam<Real>(
       "target", 1.0, "target > 0.0", "Target value of k effective to search for");
+  params.addRequiredRangeCheckedParam<Real>(
```
magnoxemo, line 39: I have just a clarifying question:

I am just wondering if `root_tol` is a really good parameter name to expose to users. And correct me if I am understanding it wrong, for `RotationSearch` root_tolerance is basically imposed on `theta` or rotation angle right? Will it be more helpful to name it something `theta_tol` RotationSearch.

Ig as it's the base class, `root_tol` makes more sense. But a few more words in the param doc string will be helpful for example what does `root_tol` mean in the context of critical boron search or rotation rotation search.

src/criticality/CriticalitySearchBase.C
```diff
@@ -105,22 +108,23 @@ CriticalitySearchBase::searchForCriticality()
...
+  BrentsMethod::root(func, _minimum, _maximum, _root_tol);
 
   // check if the method converged
-  if (abs(kMean(_estimator) - _target) >= _tolerance)
-    mooseError("Failed to converge criticality search! This may happen if your tolerance is too "
-               "tight given the statistical error in the computation of k.");
+  if (abs(kMean(_estimator) - _target) >= _k_tol)
+    mooseWarning("The eigenvalue produced by the criticality search was not within "
+                 "specified k_tol of the target! This could occur if k_tol is too "
+                 "tight for simulation's statistical error.");
```
gonuke: This can also occur for a system with very large differential worth: dk/dx, such that converging adequately on x does not converge adequately on k

gonuke (changes requested):
Minor comment on error msg

src/criticality/CriticalitySearchBase.C
lewisgross1296: > Ig as it's the base class, root_tol makes more sense. But a few more words in the param doc string will be helpful for example what does root_tol mean in the context of critical boron search or rotation rotation search.

There are other searches that are not related to changing an angle, so theta is too control drum specific for those. Also, if you look at `searchForCriticality`, the method we call is actually named root  `BrentsMethod::root(func, _minimum, _maximum, _root_tol);` so I'm pretty happy with calling it `root_tol`. 

Each type of search has docs pages with an example, so I think the user should be okay to understand the `k_tol` and `root_tol` meanings

src/criticality/CriticalitySearchBase.C
lewisgross1296: I'll add a note in the error message

doc/content/source/actions/AddCriticalitySearchAction.md
```diff
@@ -8,14 +8,13 @@ criticality search include:
...
 a postprocessor named `critical_value`.
 
 !alert warning
-If the convergence `tolerance` is too small (on the order of or smaller than the statistical standard deviation),
-convergence may not be possible. If the search fails to converge, use a looser `tolerance` or
-increase the number of particles.
+There are two tolerances that must be specified `root_tol` (the tolerance on the root the search is looking for) and `k_tol`(the tolerance on the actual tallied k-eigenvalue).  If the `root_tol` is too large, the solver might stop short of the actual critical value. If the `k_tol` is too small (on the order of or smaller than the statistical standard deviation), convergence may not be possible. If the search fails to converge, use a looser `k_tol` or increase the number of particles. Both of these tolerances are absolute tolerances, so it is important, especially on `root_tol` to input something reasonable for the problem at hand. For example, specifying a `root_tol = 0.001` is a very strict tolerance for finding a critical drum angle, which may span from 0 to 180. Conversely, a `root_tol` of 50 would likely be too large of a tolerance.
```
aprilnovak:
```suggestion
There are two tolerances that must be specified `root_tol` (the tolerance on the root the search is looking for, e.g. a critical drum angle or boron appm concentration) and `k_tol`(the tolerance on the actual tallied k-eigenvalue).  If the `root_tol` is too large, the solver might stop short of the actual critical value. If the `k_tol` is too small (on the order of or smaller than the statistical standard deviation), convergence may not be possible. If the search fails to converge, use a looser `k_tol` or increase the number of particles. Both of these tolerances are absolute tolerances, so it is important, especially on `root_tol` to input something reasonable for the problem at hand. For example, specifying a `root_tol = 0.001` is a very strict tolerance for finding a critical drum angle, which may span from 0 to 180. Conversely, a `root_tol` of 50 would likely be too large of a tolerance for drum angle but may be acceptable for critical boron concentration.
```

src/criticality/CriticalitySearchBase.C
```diff
@@ -36,10 +36,12 @@ CriticalitySearchBase::validParams()
       "Maximum for values to search over; the root must occur at a value smaller than the maximum");
   params.addRangeCheckedParam<Real>(
       "target", 1.0, "target > 0.0", "Target value of k effective to search for");
+  params.addRequiredRangeCheckedParam<Real>(
+      "root_tol", "root_tol > 0", "Absolute tolerance to converge root that yields target.");
```
aprilnovak:
```suggestion
      "root_tol", "root_tol > 0", "Absolute tolerance to converge root that yields target value of the multiplication factor.");
```

src/criticality/CriticalitySearchBase.C
```diff
@@ -105,22 +108,24 @@ CriticalitySearchBase::searchForCriticality()
...
 
   // check if the method converged
-  if (abs(kMean(_estimator) - _target) >= _tolerance)
-    mooseError("Failed to converge criticality search! This may happen if your tolerance is too "
-               "tight given the statistical error in the computation of k.");
+  if (abs(kMean(_estimator) - _target) >= _k_tol)
+    mooseWarning("The eigenvalue produced by the criticality search was not within "
+                 "specified k_tol of the target! This could occur if k_tol is too "
+                 "tight for simulation's statistical error. It can also occur if "
+                 "`root_tol` is too loose and the worth curve is steep near the target");
```
aprilnovak:
```suggestion
                 "'root_tol' is too loose and the worth curve is steep near the target");
```

src/criticality/CriticalitySearchBase.C
```diff
@@ -105,22 +108,24 @@ CriticalitySearchBase::searchForCriticality()
...
-  BrentsMethod::root(func, _minimum, _maximum, _tolerance);
+  BrentsMethod::root(func, _minimum, _maximum, _root_tol);
 
   // check if the method converged
-  if (abs(kMean(_estimator) - _target) >= _tolerance)
-    mooseError("Failed to converge criticality search! This may happen if your tolerance is too "
-               "tight given the statistical error in the computation of k.");
+  if (abs(kMean(_estimator) - _target) >= _k_tol)
+    mooseWarning("The eigenvalue produced by the criticality search was not within "
+                 "specified k_tol of the target! This could occur if k_tol is too "
```
aprilnovak:
```suggestion
                 "specified 'k_tol' of the target! This could occur if 'k_tol' is too "
```

aprilnovak:
Thanks @lewisgross1296! Looks good to me, just some minor comments for clarity/examples.

test/tests/criticality/material_density/tests
```diff
@@ -19,16 +19,16 @@
...
     requirement = 'The system shall warn if the selected tolerance might cause failure to converge due to high statistical noise.'
     capabilities = 'openmc'
   []
   [fail_converge]
     type = RunException
     input = openmc.i
-    cli_args = 'Problem/CriticalitySearch/tolerance=1e-6'
-    expect_err = "Failed to converge criticality search!"
+    cli_args = 'Problem/CriticalitySearch/k_tol=2e-3 Problem/particles=100000 --error'
+    expect_err = "The eigenvalue produced by the criticality search was not within specified k_tol of the target!"
```
aprilnovak:
```suggestion
    expect_err = "The eigenvalue produced by the criticality search was not within specified 'k_tol' of the target!"
```

aprilnovak: sorry, broke your test!

lewisgross1296: No worries! I didn't double check the changes :)

PR #1334, opened 2026-05-14
Add an `AuxKernel` to compute various figures of merit

### LINKED ISSUE #1333: Add an `AuxKernel` to compute various figures of  merit
## Reason
It is often useful to compute a Figure of Merit (FoM) for Monte Carlo algorithms to enable objective comparisons between techniques. The definition of a FoM depends on the algorithms under evaluation. For variance reduction in Monte Carlo transport simulations the following FoM is often used:

```math
\text{FoM}_{VR,i} = \frac{1}{T R_{i}^{2}}
```

where $T$ is the total simulation time (time spent running the Monte Carlo calculation and performing variance reduction), and $R_{i}$ is the stochastic relative error in tally bin $i$. When evaluating Monte Carlo tally AMR algorithms, I've proposed the following FoM:

```math
\text{FoM}_{AMR,i} = \frac{1}{T R_{j,i}}\left(\frac{|u_{j,i} - u_{0,i}|}{u_{0,i}}\right)
```

where $u_{j,i}$ is the tally value at AMR cycle $j$ in tally bin $i$, $u_{0,i}$ is the tally value on the initial (non-adapted) calculation, and $R_{j,i}$ is the tally stochastic relative error at AMR cycle $j$ in tally bin $i$. 

We should add an auxkernel to compute these FoMs for users. This would enable comparisons between stochastic algorithms without requiring external post-processing or the specification of many `ParsedAux` auxkernels (which is quite tedious).

## Design
Addition of an `FoMAux` which computes either the variance reduction or AMR FoM base on the field variables provided by the user.

## Impact
New feature, quality of life improvement for users running large parameter sweeps to evaluate the performance of variance reduction or AMR algorithms in a Monte Carlo setting.

nuclearkevin:
Closes #1333


src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+registerMooseObject("CardinalApp", FoMAux);
+
+InputParameters
+FoMAux::validParams()
+{
+  auto params = OpenMCAuxKernel::validParams();
+  params.addClassDescription("An auxkernel which computes a figure of merit for a tally.");
+  params.addCoupledVar("tally_value",
+                       "The variable containing the value of the tally. FoMAux "
+                       "assumes this is a volumetric quantity.");
```
aprilnovak: can you elaborate more on the "volumetric quantity" part here? What if the OpenMC model was 3-D, but a 2-D mesh is used for AMR? Does that violate this "volumetric quantity" part here, or is this moreso referring to a variable that only exists on a side (or something else?)?

src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+      "tally_value_init",
+      "The variable containing the tally value on the initial adaptivity step. FoMAux "
+      "assumes this is a volumetric quantity. This is required for the AMR "
+      "figure of merit.");
+  params.addRequiredCoupledVar(
+      "tally_rel_error",
+      "The variable containing the statistical relative error of the tally. FoMAux "
+      "assumes this is a volumetric quantity.");
+
+  params.addRequiredParam<PostprocessorName>("sim_time", "The cumulative simulation time.");
```
aprilnovak, line 42: Can the documentation string give a recommendation on a specific postprocessor to use?

src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+
+  params.addRequiredParam<MooseEnum>(
+      "fom_type",
+      MooseEnum("VR AMR"),
+      "The type of Figure of Merit (FoM) to compute. Options are the standard "
+      "variance reduction FoM (VR) or the adaptive mesh refinement FoM (AMR).");
```
aprilnovak, line 48: How strongly do you feel about "VR" and "AMR" as the enum values? What about `variance_reduction` and `adaptive_mesh`? I tend to prefer things which convey more information in their name out of clarity.

src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+
+FoMAux::FoMAux(const InputParameters & parameters)
+  : OpenMCAuxKernel(parameters),
+    _tally_val(isCoupled("tally_value") ? &coupledValue("tally_value") : nullptr),
+    _tally_val_init(isCoupled("tally_value_init") ? &coupledValue("tally_value_init") : nullptr),
+    _tally_val_rel_err(coupledValue("tally_rel_error")),
+    _sim_time(getPostprocessorValue("sim_time")),
+    _fom_type(getParam<MooseEnum>("fom_type").getEnum<FoMType>())
+{
+  if (_var.feType() != FEType(libMesh::CONSTANT, libMesh::MONOMIAL))
```
aprilnovak, line 61: is it not possible to call `errorCoupledConstMonomial` here instead of repeating this code? Or is it not possible to get the variable name internally?

src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+{
+  switch (_fom_type)
+  {
+    case FoMType::VR:
+      return 1.0 / _sim_time / _tally_val_rel_err[0] / _tally_val_rel_err[0];
+    case FoMType::AMR:
+      return std::abs((*_tally_val)[0] - (*_tally_val_init)[0]) / _sim_time /
+             _tally_val_rel_err[0] / (*_tally_val_init)[0];
+    default:
+      return 0.0;
```
aprilnovak, line 89: why not a `mooseError` here to catch unhandled enum?

src/auxkernels/FoMAux.C
nuclearkevin: What I mean by "volumetric quantity" here is that it is divided by the volume of the tally bin (e.g. W -> W/cm3). 

Also, we cannot use a 2D libMesh tally in OpenMC, an error will be thrown.

src/auxkernels/FoMAux.C
```diff
@@ -0,0 +1,106 @@
...
+    paramError("variable",
+               "FoMAux only supports CONSTANT MONOMIAL shape functions. Please "
+               "ensure that 'variable' is of type MONOMIAL and order CONSTANT.");
+
+  errorCoupledConstMonomial("tally_value");
+  errorCoupledConstMonomial("tally_value_init");
+  errorCoupledConstMonomial("tally_rel_error");
+
+  // Error-check the different FoM options.
+  switch (_fom_type)
```
aprilnovak: I don't feel strongly so you don't have to do this, but elsewhere in Cardinal the convention is to do error checking just based on each enum alone, since it's sort of a different behavior than handling some type of behavior that all enum options _must_ do.

```
if (_fom_type == FoMType::AMR)
{
  // check stuff
}
```

src/auxkernels/OpenMCAuxKernel.C
```diff
@@ -46,6 +46,18 @@ OpenMCAuxKernelTempl<ComputeValueType>::mappedElement()
   return !(cell_info.first == OpenMCCellAverageProblem::UNMAPPED);
 }
 
+template <typename ComputeValueType>
+void
+OpenMCAuxKernelTempl<ComputeValueType>::errorCoupledConstMonomial(const std::string & param_name)
+{
+  if (this->isCoupled(param_name))
```
aprilnovak, line 53: should an error be thrown here if the variable is not coupled in?

test/tests/neutronics/foms/amr.i
```diff
@@ -0,0 +1,75 @@
+[Mesh]
+  [sphere]
+    type = FileMeshGenerator
+    file = ../meshes/sphere.e
+  []
+  [solid1]
```
aprilnovak: do we need to renumber subdomain to 100?

test/tests/neutronics/foms/amr.i
```diff
@@ -0,0 +1,75 @@
...
+    []
+  []
+[]
+
+[AuxVariables]
+  [AMR_fom]
+    order = CONSTANT
+    family = MONOMIAL
+  []
+  [kappa_fission_init]
```
aprilnovak, line 32: How does this work in practice? I would have thought that the initial value would be the value from adaptivity cycle 0. Is an initial condition being manually set here just for the sake of the test being able to run without the AMR system? If so, maybe a comment here could be helpful for future users to make it clear this is just how the test is configured.

test/tests/neutronics/foms/amr.i
```diff
@@ -0,0 +1,75 @@
...
+
+[AuxKernels]
+  [AMR_fom]
+    type = FoMAux
+    variable = 'AMR_fom'
+    fom_type = 'AMR'
+    tally_value = 'kappa_fission'
+    tally_value_init = 'kappa_fission_init'
+    tally_rel_error = 'kappa_fission_rel_error'
+    sim_time = 2.0
```
aprilnovak, line 56: same comment here, presumably this would come from a postprocessor but actually using simulation time in a test would make it impossible to match gold files

src/auxkernels/FoMAux.C
nuclearkevin: At present, there's no good post-processor to use :( 

MOOSE does not provided a PP that returns simulation walltime. For the PDF rejection sampling studies, I used a `Receiver` that was manually populated by a custom problem. For all of the other FoM studies using actual OpenMC simulations, the cumulative walltime was computed manually and provided to the `FoMAux` as a constant post-processor. 

I'd like to add a wall time post-processor, but decided to submit this first as things can take a while to get into the framework (especially given our issues with submodule updates as of late).

src/auxkernels/FoMAux.C
nuclearkevin: My personal preference as a user is to keep things reasonably short on the enumeration side of things as it is less tedious to type out repeatedly. If you feel strongly about this, I can change it.

test/tests/neutronics/foms/vr.i
```diff
@@ -0,0 +1,64 @@
+[Mesh]
+  [sphere]
+    type = FileMeshGenerator
+    file = ../meshes/sphere.e
+  []
+  [solid1]
```
aprilnovak: remove subdomain ID customization?

src/auxkernels/FoMAux.C
nuclearkevin: We cannot call it here as the variable acted on by an AuxKernel does not go through the `Couplable` interface.

src/auxkernels/FoMAux.C
aprilnovak: Right, forgot about 2D mesh constraints.

So, I suppose like FET tallies would violate the "volumetric quantity" part, as an example? Could we clarify in the doc strings that "volumetric quantity" means "assumes this field has 1/volume units" or something like that?

src/auxkernels/FoMAux.C
aprilnovak: I'm okay with your design choice!

src/auxkernels/FoMAux.C
nuclearkevin: I tend to default to switch statements with enums, but in this case I definitely agree that a single if statement is a better way to go about this.

src/auxkernels/OpenMCAuxKernel.C
nuclearkevin: With how I've designed this throwing an error on the other branch would always result in an error getting thrown for the VR FoM (as it doesn't require the tally value or the old tally value).

nuclearkevin: (e.g. this would no longer work for optional coupled variables)

aprilnovak: would it be better to place the three calls to `errorCoupledConstMonomial` inside an `if (AMR FOM)` block instead? I am thinking down the future of a scenario where other classes might call this `errorCoupledConstMonomial` and fail to catch that a variable is not const monomial because it is not coupled to the class (as opposed to truly being valid variable type/order).

On that note, are there other places in Cardinal where we can leverage this function to replace some code?

test/tests/neutronics/foms/amr.i
nuclearkevin: In practice, picking out the value at the initial adaptivity cycle is a massive pain. MOOSE provides no explicit way to do this, so you need to use a projection aux to copy the value of the variable on cycle 0 into a separate variable. Then, you disable the projection aux with the controls system to make sure the variable doesn't get overriden.

An initial condition was chosen instead to avoid needing to jump through those hoops for a simple test. I plan on adding a FoM section to the AMR tutorial at some point to go over this process.

test/tests/neutronics/foms/amr.i
nuclearkevin: I'll add a comment to these tests to mention that this is a proxy, same with the initial adaptivity cycle variable in the other test.

src/auxkernels/OpenMCAuxKernel.C
nuclearkevin: I scanned over the OpenMC AuxKernels and none of them can use this at present.

PR #1337, opened 2026-05-16
Ensure coupling is reinitialized on criticality searches + improve criticality search performance

### LINKED ISSUE #1336: Improvements to multiphysics criticality searches
## Reason
When Cardinal performs a criticality search it modifies the underlying CSG geometry or materials without regenerating the cell-to-element (and dual) maps. This causes erroneous behaviour when using cell tallies which will have their tallied values stored in the elements mapped at the beginning of the Picard iteration instead of elements which map to the new cell position. 

Additionally, multiphysics criticality searches are quite slow at present as all mapped tallies are computed during the search process (even though the intermediate tallies are thrown out). We should add an option to criticality search objects to disable the tallies added by Cardinal to improve performance.

## Design
A fix for this problem is conceptually straight-forward. `OpenMCCellAverageProblem` should re-generate the cell to element (and dual) maps after the criticality search is performed using the in-memory critical state. Then, a final OpenMC calculation should be performed on the critical state to obtain the critical state tallies.

## Impact
Correcting erroneous behaviour, performance improvement for multiphysics criticality searches.

nuclearkevin:
This PR reinitializes the cell->element (and dual) maps on every criticality search iteration, which ensures OpenMC cells receive the correct temperatures and densities to converge to the correct critical state in a Picard iterate. This also ensures cell tally results are correctly mapped to MOOSE mesh elements.

I also took the liberty of adding a performance enhancement to criticality searches where tallies are disabled during the search process (if requested), and a final "critical state" calculation can be run with tallies enabled again. Open to opinions on how this performance enhancement can be tested, especially given our inability to output each individual criticality search iteration due to MOOSE output restrictions.

Closes #1336

### DIFF
--- a/include/criticality/OpenMCMaterialDensity.h
+++ b/include/criticality/OpenMCMaterialDensity.h
@@ -31,7 +31,6 @@ public:
   OpenMCMaterialDensity(const InputParameters & parameters);
 
   virtual void updateOpenMCModel(const Real & input) override;
-  virtual bool changingGeometry() const override { return false; }
 
 protected:
   virtual std::string quantity() const override
--- a/include/criticality/OpenMCMaterialDensity.h
+++ b/include/criticality/OpenMCMaterialDensity.h
@@ -31,6 +31,7 @@ public:
   OpenMCMaterialDensity(const InputParameters & parameters);
 
   virtual void updateOpenMCModel(const Real & input) override;
+  virtual bool changingGeometry() const override { return false; }
 
 protected:
   virtual std::string quantity() const override
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -2638,7 +2638,7 @@ OpenMCCellAverageProblem::reinitCoupling()
         // we don't need to send anything in to OpenMC, so we can leave.
         importProperties();
         _console << "Skipping " << incoming_transfer
-                  << " transfer into OpenMC because 'initial_properties = hdf5'" << std::endl;
+                 << " transfer into OpenMC because 'initial_properties = hdf5'" << std::endl;
         return;
       }
       case coupling::moose:
@@ -2651,7 +2651,7 @@ OpenMCCellAverageProblem::reinitCoupling()
         // if we're just using whatever temperature and density are already in the XML
         // files, we don't need to send anything in to OpenMC, so we can leave.
         _console << "Skipping " << incoming_transfer
-                  << " transfer into OpenMC because 'initial_properties = xml'" << std::endl;
+                 << " transfer into OpenMC because 'initial_properties = xml'" << std::endl;
         return;
       }
       default:
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -28,6 +28,7 @@
 #include "SetupMGXSAction.h"
 #include "OpenMCVolumeCalculation.h"
 #include "CreateDisplacedProblemAction.h"
+#include "CriticalitySearchBase.h"
 
 #include "openmc/constants.h"
 #include "openmc/cross_sections.h"
@@ -526,6 +527,9 @@ OpenMCCellAverageProblem::initialSetup()
   // Coupling re-initialization should be triggered if we have cells transforms which can happen
   // even if the mesh isn't moving or adaptive.
   _need_to_reinit_coupling |= hasCellTransform();
+  // The criticality search may modify the geometry.
+  if (_criticality_search)
+    _need_to_reinit_coupling |= _criticality_search->changingGeometry();
 
   if (!_needs_to_map_cells)
     checkUnusedParam(parameters(),
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -523,7 +523,7 @@ OpenMCCellAverageProblem::initialSetup()
                  "[Mesh] will move, but the underlying OpenMC geometry will remain unchanged. "
                  "Unexpected behavior may occur.");
 
-  // Coupling re-initialization should be triggered if we have cells transofrms which can happen
+  // Coupling re-initialization should be triggered if we have cells transforms which can happen
   // even if the mesh isn't moving or adaptive.
   _need_to_reinit_coupling |= hasCellTransform();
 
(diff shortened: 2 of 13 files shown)


include/criticality/OpenMCMaterialDensity.h
```diff
@@ -31,6 +31,7 @@ class OpenMCMaterialDensity : public OpenMCMaterialSearch
   OpenMCMaterialDensity(const InputParameters & parameters);
 
   virtual void updateOpenMCModel(const Real & input) override;
+  virtual bool changingGeometry() const override { return false; }
```
gonuke: Couldn't this live in the `OpenMCMaterialSearch` class and only be implemented once, instead of here & `BoratedWater`?

gonuke (changes requested):
Thanks @nuclearkevin!

Just one suggestion to move a function definition up the inheritance tree...

include/criticality/OpenMCMaterialDensity.h
nuclearkevin: Whoops, thanks for catching this! I'll shuffle it up one level

lewisgross1296:
Followed up on slack but this solved my issue. Also of note, I tried out both disabling tallies during search and doing the critical state solve. My critical answers didn't change by much but the heating tally now shows up properly (I.e. remapping seems to have resolved my concern). 

Thanks for getting to this so quickly @nuclearkevin!!

gonuke (approved):
This LGTM - thanks @nuclearkevin - I'll hold on merging to let @aprilnovak take a look

gonuke:
> This LGTM - thanks @nuclearkevin - I'll hold on merging to let @aprilnovak take a look

I can't merge anyway 😀

aprilnovak:
@nuclearkevin  could you please rebase this branch? New conflict in documentation after merging another PR

src/criticality/CriticalitySearchBase.C
```diff
@@ -48,6 +51,17 @@ CriticalitySearchBase::validParams()
       "are used, statistical noise may require many criticality calculations to converge.");
   params.addParam<MooseEnum>(
       "estimator", getEigenvalueEnum(), "Type of eigenvalue estimator to use");
+  params.addParam<bool>(
+      "run_critical_state",
+      true,
+      "Whether a final k-eigenvalue calculation should be performed on the critical model state.");
```
aprilnovak, line 57: Is this option strictly referring to Cardinal/accessing the last OpenMC run via Cardinal? Or does OpenMC apply the last criticality manuever (e.g. change boron concentration) and then *not* run a final k-eigenvalue calculation?

src/criticality/CriticalitySearchBase.C
```diff
@@ -64,19 +78,28 @@ CriticalitySearchBase::CriticalitySearchBase(const InputParameters & parameters)
...
+    _run_critical_state(getParam<bool>("run_critical_state")),
+    _tally_during_search(getParam<bool>("tally_during_search")),
     _target(getParam<Real>("target"))
 {
   if (_minimum >= _maximum)
     paramError("minimum",
                "The 'minimum' value (" + std::to_string(_minimum) +
                    ") must be less than the 'maximum' value (" + std::to_string(_maximum) + ").");
 
+  if (!_tally_during_search && !_run_critical_state)
```
aprilnovak, line 90: Could the code be modified to always run tallies on the last step instead of having the user required to set the parameters like this? Or is the issue that we don't know when to re-enable tallies?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2536,90 +2540,8 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
...
-        // Clear the MGXS manager.
-        openmc::data::mg = {};
-        // Reload the MGXS data.
-        openmc::data::mg.read_header(openmc::settings::path_cross_sections);
-        openmc::put_mgxs_header_data_to_globals();
-        openmc::finalize_cross_sections();
-        openmc::settings::verbosity = initial_verbosity;
-      }
+      // Reinitialize the MOOSE -> OpenMC coupling.
+      reinitCoupling();
```
aprilnovak: Can we split this function up into smaller pieces? My only concern is with the naming - reinitializing coupling makes sense as a descriptor for setting up the geometric maps, but sending new temperatures and densities into OpenMC is *not* done on initialization of the problem in the same manner, but rather is only called within `syncSolutions`.

aprilnovak: or maybe just a different name would work - what about `reinitCouplingAndApplyFeedback` or something?

src/criticality/CriticalitySearchBase.C
lewisgross1296: I think this would prevent the current behavior, which is to not run a multi physics iteration solve at the critical guess. In some cases, the final critical guess is very close to the second to last guess and temp/dens change marginally. Though with certain worth curves and search tolerances, it would be important to rerun at the critical guess for correctness.

I'm not necessarily opposed to forcing the critical state to run, as it's higher fidelity, but I guess some people might not _need_ it and decide to skip the critical state solve to save on runtime.

src/criticality/CriticalitySearchBase.C
nuclearkevin: Most root finder implementations (including Brent's method in MOOSE) return the final iteration which satisfies the root-finding constraints & algorithm. They do not execute the nonlinear function on the final guess as the goal is to find the root, not evaluate the function at the root.

This option allows users to run the final iteration if they want to (the critical state calculation). This often won't be all that different compared to the second-to-last iteration (the final Monte Carlo calculation executed in the Brent's method function), but some users (e.g. @lewisgross1296) want to perform this final iteration for the sake of completeness. I think this should be an option as opposed to being forced as it incurs the performance penalty of another iteration (which can be fairly substantial if few iterations are required for convergence).

src/criticality/CriticalitySearchBase.C
nuclearkevin: We cannot modify it so Brent's method only runs tallies on the final iteration, as we don't know what the final iteration will be.

src/base/OpenMCCellAverageProblem.C
nuclearkevin: My preference would be to rename this function as opposed to splitting it.

aprilnovak:
LGTM!

PR #1340, opened 2026-05-27
V26 update

### LINKED ISSUE #1345: Update Cardinal to v26 NekRS
## Reason
NekRS is now on version v26; update Cardinal to this version.

## Design
Update code base to v26.

- [ ] Add tests where possible to better cover the field-file writing aspects of Cardinal (e.g. Cardinal writes field files, we can do better than `CheckFiles` type tests, such as by loading initial conditions, to make sure these don't break and don't require manual intervention going forward.

## Impact
Breaking changes for all NekRS-related code, tests, and examples.

aprilnovak:
This PR required updates to every NekRS test, due to breaking changes in V26. Below is a summary of *major* and *minor* changes performed for each test.  Closes #1345

## Source Code Changes (aside from syntax updates)

- In order to support some tests which used to set known functions in the udf and then pass that into the usrwrk array, I added the `NekVolumetricData` object to accomplish the same thing but compatible now with v26.
- NekRS modified how it numbers the passive scalars. If temperature is not present, then the first passive scalar is saclar00 (0-indexed). Some source code changes are related to robustifying these indices.

## Minor Changes to Tests
Minor changes encompass routine updates to syntax. Since I was reviewing every test, I also took the liberty to simplify tests when possible - for example, if the actual solution fields were not used in NekRS, why not eliminate extraneous passive scalars and application of boundary conditions? Similarly, unused files were deleted (e.g., NekRS now supports the udf = "file.udf", so we could delete some extraneous files). It also seems that, at least when testing this locally, that some precompilation is no longer needed for the stochastic sampling runs, so I have removed some of those prereq test recipes.

Since the vast majority of the test updates fall into this category, I simply will not list a specific test if it falls in this category and let the file diffs speak for themselves.

## Major Changes to Tests
Major changes required re-golds. Below is a description of each case for your review.

-	`cht/nondimensional`: If you let these files run to the pseudo steady state, they blow up from unreasonable CFL after the first data transfer from MOOSE. This test previously passed I suppose because we only ran one time step to reach the gold file – the case was set up with Re=3000, which is turbulent but uses a fairly coarse mesh. Presumably many years ago, this case did run to the pseudo-steady state for some coincidental reason, since it does not make sense for such a coarse mesh with a turbulent flow to have been stable. So, this inconsistency has been fixed while simultaneously making the test smaller to allow us to run more time steps to hopefully insulate ourselves in the future (one change you will see is a compression vertically in the dimension to allow fewer axial elements in the heat conduction mesh). In addition, while looking at this test I realized that the ../sfr_pincell test, which this non-dimensional version was supposed to compare against as a dimensional reference, was set up with different conditions. Perhaps at some time in the past they were consistent with one another, but the test description which states that the nondimensional results were compared to the dimensional results was now clearly incorrect. So, I updated the ../sfr_pincell test to match the conditions in the nondimensional folder, and re-assessed equivalence between the dimensional and non-dimensional results. Even with these updates, the test displays non-determinism in parallel. I confirmed that the same non-determinism existed prior to v26 – we just never really saw it in the test suite because we have `min_parallel = 6`, which effectively meant we always ran this test with 6 ranks because none of the recipes launch with > 3 ranks. The diffs are on the order of 1e-5 or smaller – and I confirmed that these diffs do not grow in time by running a very long pseudo-transient.
-	`cht/sfr_pincell`: For consistency with the ../nondimensional case, I updated this test to use identical settings and separately confirmed that it matches the (scaled) nondimensional test data. The tests now run faster as well.
- re-golded the `test/tests/lowMach` test. Small diffs, only explainable as being caused by differences in the problem/NekRS solve itself, are the culprit. The purpose of this test is not to act as a NekRS CI case itself, but rather as a Cardinal case and it still meets that purpose. The postprocessors in Cardinal are functioning correctly, based on looking at the max/min console output from NekRS (the diffs in this test were from the `max_p` and `min_p` postprocessors).
- deleted the `test/tests/nek_standalone/ethier` test. This test was performing an identical operation as the `lowMach` test so it is not necessary to have the redundancy (later we plan to have the CI cases run through Cardinal anyways, so we plan to add the *specific* things of `ethier` in `nek_ci`). It is easier to delete since a re-gold would also have been necessary, since for some reason temperature was removed as a solution field in v26's `ethier` case relative to the v23 versions of these files.
- ADDED new tests in `nek_file_output/usrwrk`: this test previously only checked the existence of certain file outputs. I added some tests which then read in those files to confirm they also hold the correct data. This insulates us in the future from breakages where the files are created, but don't hold the correct data.
- ADDED new tests in `postprocessors/nek_point_value`: previously, we interpolated all scalars to query at a point value, and then indexed into them in a rigid way. The new test ensures that the indexing functions as well when temperature is not present but other scalars are.
- ADDED new tests in `postprocessors/nek_volume_integral`: this fully covers the new scalar index API 
- re-golded the `userobjects/side/nondimensional` test: this test was showing failures on CIVET, but not locally. After further testing, I believe this was due to a small floating point indeterminism in which bin of the userobject an element falls into, since the elements in the original mesh were nearly perfectly straddling one of the boundaries of the userobject edges. I recreated the mesh to have better quality and re-golded, but still confirmed that the values look identical to the prior test and still match the dimensional version for which the purpose of this test exists.

### Tutorials

The tutorials were all updated, both in terms of the input files as well as the associated documentation. For my own tracking purposes, below is a list of the completed tutorials and any notable changes (beyond just updating the syntax):

- `load_from_exodus`: this test used the `ethier` NekRS case in the past, but this is unecessarily brittle since it's a complicated case and had huge diffs between syntax in v26 and v23. I instead opted to just copy in a simple Nek case that we already had which is totally adequate to demonstrate what the tutorial is for.
- `nek_stochastic`
- `pebble_1`: it was easy to add a test to cover this case, which I included to insulate us from future large code changes.
- `pebble_cht`
- `pincell_multiphysics`
- `standalone`
- `subchannel`: the test was previously just a RunApp test, but it is easy to extend to make a `CSVDiff` type test which will eliminate the need to open up the output files to view and confirm correct behavior for future updates.
- `turbulence`


doc/content/source/postprocessors/field_specs.md
```diff
@@ -10,7 +10,7 @@ The field is specified with the `field` parameter, which may be one of:
 - `velocity_y_squared`: $y$-component of velocity, squared
 - `velocity_z_squared`: $z$-component of velocity, squared
 - `velocity_component`: velocity vector dotted against another vector
-- `scalar01`: first passive scalar
-- `scalar02`: second passive scalar
-- `scalar03`: third passive scalar
+- `scalar01`: first non-temperautre passive scalar
```
nandu90: minor typo

src/base/NekInterface.C
```diff
@@ -159,49 +241,46 @@ endControlNumSteps()
 bool
 hasTemperatureVariable()
 {
-  nrs_t * nrs = (nrs_t *)nrsPtr();
-  return nrs->Nscalar ? platform->options.compareArgs("SCALAR00 IS TEMPERATURE", "TRUE") : false;
+  return Nscalar() ? platform->options.compareArgs("SCALAR00 NAME", "TEMPERATURE") : false;
```
nandu90, line 244: Do we want to mandate that temperature is scalar00? We can change this to have temperature be at any location, as specified by the user in par file. 

Let me know if you recommend this change.

src/base/NekInterface.C
```diff
@@ -159,49 +241,46 @@ endControlNumSteps()
...
 
 bool
 hasHeatSourceKernel()
 {
-  return static_cast<bool>(udf.sEqnSource);
+  return static_cast<bool>(nrs->userSource);
 }
 
 bool
 isInitialized()
```
nandu90, line 269: check

src/base/NekInterface.C
aprilnovak: For now, I am thinking we leave it as-is and assume Cardinal always puts temperatures as scalar 00. I've added this point (that now temperature can have a general position in the scalars) as an issue to track here after the first v26 update goes in (in the interest of keeping this PR more dedicated to the bare minimum needed to update Cardinal to v26).

https://github.com/neams-th-coe/cardinal/issues/1296

nandu90: Sounds good

src/base/NekInterface.C
nandu90: The intent of this function seems to be to check if `nrs->mesh` exists (https://github.com/aprilnovak/cardinal/blob/v26-update/src/mesh/NekRSMesh.C#L74) . The function should be modified to do that, instead of returning `nrs` object.

nandu90 (commented):
The updates look good to me overall, besides the minor specific comments. I have not run the CI suite on my end. I assume the Nek related tests are passing.

src/base/NekInterface.C
aprilnovak: Not quite, the purpose is to check that the `NekRSMesh` object has been used in an input file with a `NekRSProblem`. The intent is to check if `nrs` itself exists (I can't replace it by a check to `nrs->mesh` because the error this is checking is if `nrs` exists in the first place, so checking for `nrs->mesh` seg faults). This is the only way I know how to test this, because the mesh gets created first before the `[Problem]` block gets executed (whose presence causes `NekInitAction` to call and initialize nek).

aprilnovak:
The deformation tests have been skipped, as agreed with @dshaver14. There will be a dedicated effort in the future to port over these tests, and we feel it is more important to get v26 update merged than to delay on these tests.

aprilnovak:
Major thanks to @loganharbour for his help with changes on the testing side to support this!

nandu90:
Reviewed the changes. Looks good. Ready to merge.

PR #1342, opened 2026-05-28
Remove fluid property app tests

loganharbour:
These tests aren't particularly useful and at the moment due to how MOOSE tests cardinal with the new way of getting the fluid property apps, these tests fail because they use the old syntax.

See https://civet.inl.gov/job/3860773/, in which these tests are old (and not tested in Cardinal).


aprilnovak:
Thanks @loganharbour ! should we also remove these as submodules at the same time, and replace by the proper recursive pull inside MOOSE?

loganharbour:
I think it's probably best to handle all of the app stuff at the same time. Which I realized isn't as straight forward of an easy task as this.

At the moment my priority is just getting the auto MOOSE updates. Once that's in place, we can fine tune these a bit more.

aprilnovak (approved):

PR #1344, opened 2026-05-30
update frontier

aprilnovak:
This updates the job script recommendation for Frontier; compatible with v26 in Cardinal.

@AyaHegazy22 can you please take a look and confirm that these instructions are compatible with your IBM code? (of course, additional steps are needed to compile the IBM work but if you can confirm these instructions are the correct "starting point" that would be great!)

@MariaReji03 please confirm if you can build with these instructions

### DIFF
--- a/doc/content/hpc.md
+++ b/doc/content/hpc.md
@@ -169,7 +169,7 @@ export CARDINAL_DIR=$HOME_DIRECTORY_SYM_LINK/cardinal
 
 ## Frontier
 
-Last updated on 02/17/2024
+Last updated on 03/25/2026 (thanks to M. Dalinger for his contributions)
 
 Frontier is an [!ac](HPC) system at [!ac](ORNL) with 9408 AMD compute nodes, each with
 4 AMD MI250X, each with 2 Graphics Compute Dies (GCDs), which you can think of as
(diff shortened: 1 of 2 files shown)


MariaReji03:
I tested the updated Frontier instructions. The bashrc commands and build steps are correct and worked after one additional change.

Issue: The build was failing when trying to build the GPU version of HYPRE with the error:
`g++: error: language hip not recognized`

Solution: I modified `/contrib/nekRS/CMakeLists.txt` to disable HYPRE GPU support:
Changed from:
`set(ENABLE_HYPRE_GPU ON CACHE BOOL "Enable HYPRE GPU support")`
To:
`set(ENABLE_HYPRE_GPU OFF CACHE BOOL "Enable HYPRE GPU support")`

aprilnovak:
Thanks @MariaReji03  - can you also please confirm that you're able to run a case with the Frontier job script?

MariaReji03:
The jobscript works fine after making the necessary adjustments. Two suggestions:

1. To the usage instructions, add 
"Run `chmod +x job_frontier_v26` to make the script executable"
2. If you disabled HYPRE GPU capabilities during the Cardinal build (by setting `ENABLE_HYPRE_GPU=OFF` in `/contrib/nekRS/CMakeLists.txt`), you must add `export ENABLE_HYPRE_GPU=OFF` to the jobscript. Otherwise, the job will fail with linker errors for `libnekrs-hypre.so`.

PR #1347, opened 2026-05-30
fix compile warning

aprilnovak:
This code throws compile warning on Mac with conda environment.

### DIFF
--- a/src/transforms/OpenMCCellTransformBase.C
+++ b/src/transforms/OpenMCCellTransformBase.C
@@ -100,7 +100,7 @@ OpenMCCellTransformBase::transform(const MooseEnum & transform_type, const Point
                              << ") degrees." << std::endl;
     }
     else
-      _moose_object.mooseError("In transformation, unknown transformation type: " + transform_type);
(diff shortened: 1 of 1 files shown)


nuclearkevin (approved):
LGTM. I guess the clang implementation on Apple hardware is more strict about casting c-strings to `std::string`s.

loganharbour:
I'm working on adding clang recipes to cardinal testing which should allow this to get picked up in CI.

I think currently we don't add all of the extra warning -> error flags in the cardinal builds in CI. Probably should.

nuclearkevin:
> I'm working on adding clang recipes to cardinal testing which should allow this to get picked up in CI.
> 
> I think currently we don't add all of the extra warning -> error flags in the cardinal builds in CI. Probably should.

We should definitely be compiling with `-Werror`, though I don't think we should add the flag to the NekRS part of the build (if possible). NekRS doesn't use standard versioning and it would be a pain to get that software stack to be `-Werror` compliant. Less of an issue with OpenMC/DAGMC - they're really good about merging small PRs reasonably quickly.

PR #1348, opened 2026-05-30
Apply black format to python

loganharbour:
This is up to you guys if you want to switch to this. This is a part of my bring-cardinal-up-to-date-with-other-apps push.

Black format is the equivalent for python that clang format is for c++. We use this on moose and a few of the applications now. If interested, we'd add a check to civet to check for formatting. The `black` utility is installed automatically in the conda environment, but can be installed outside of it with `pip install` or `conda install`. There's also vscode plugins for auto formatting code on save.

I use claude to describe briefly why you would want this:

> Black is an opinionated Python code formatter that enforces a consistent style automatically.Black is an opinionated Python code formatter that enforces a consistent style automatically. Here's why it's valuable in a repo:
> 
> - Eliminates style debates. Black makes formatting decisions for you — there's no arguing about line length, quote style, or whitespace. Teams stop bikeshedding and focus on logic instead.
> 
> - Consistent codebase. Every file looks the same regardless of who wrote it. This reduces cognitive load when reading unfamiliar code, since the style is always predictable.
> 
> - Cleaner diffs and PRs. When Black handles formatting, code reviews don't get cluttered with nitpicks about spacing or style. Diffs reflect actual logic changes, not reformatting noise.
> 
> - Removes formatting burden from developers. You can write messy code as you think, then run Black (or have it run on save) and it's instantly clean. No mental overhead maintaining style manually.
> 
> - Easy CI enforcement. Running black --check in CI ensures nothing merges without being formatted, keeping the repo consistently clean without relying on developer discipline.
> 
> - Low friction onboarding. New contributors don't need to learn a style guide — they just run Black and it handles everything.
> 
> The main tradeoff is that Black is intentionally unconfigurable (with minor exceptions like line length). You accept its style wholesale, which is actually the point — it removes all the decisions. For teams that can accept that, it dramatically reduces formatting friction.

If we do this, I'd attribute the formatting commit to a maintenance account. I don't need code contributions for formatting files.


nuclearkevin:
I'm on board with enforcing Black formatting with a CIVET check. My only concern is if different version of Black (installed locally with pip) will yield different formatting changes than the formatting check on CIVET. Would it be possible to get Black style patches from @moosebuild in a similar manner to clang-format to get around this?

lewisgross1296:
Big +1 on this. Agree that Kevin's suggestion would be super useful. I'm sure it's possible somehow. 

There's one thing Claude mentioned above that I think is worth discussing
> The main tradeoff is that Black is intentionally unconfigurable (with minor exceptions like line length). You accept its style wholesale, which is actually the point — it removes all the decisions. For teams that can accept that, it dramatically reduces formatting friction.

In my experience sometimes black aggressively breaks things in to multiple lines making python files a bit longer and/or re-formats into slightly less readable code. In my own code bases with theses scenarios, we use 
```
# fmt: off
line that black over aggressively formats
# fmt: on
```

It's kind of against the philosophy, but sometimes it feels more "right." This could potentially be mitigated by using a longer line length default. I'm specifically thinking of some OpenMC model builder code that we use `# fmt: off/on` in. Would be a case by case basis if this was decided to be allowed.

PR #1353, opened 2026-06-04
Internally convert non-TET4/10 mesh elements in MoabSkinner

### LINKED ISSUE #1352: Allow usage of HEX8 meshes with the skinner
## Reason
Currently MoabSkinner only accepts tetrahedral meshes (TET4 and TET10). Any mesh containing HEX8 elements triggers an error. This forces users who want to couple their apps. that use hexahedral elements with Cardinal and want to use the skinner to add `ElementsToTetrahedronsConverter` into their meshing to be able to use it.

This also forces Cardinal to tally OpenMC results on the tetrahedral mesh only.

## Design
We can detect non-tetrahedral elements inside MoabSkinner and perform the conversion to TET4 internally. where the MOOSE mesh is cloned and HEX8 elements are converted to TET4 using `MooseMeshElementConversionUtils::convert3DMeshToAllTet4()`. The original MOOSE mesh is never modified. Then a map between the clone elements and their parent HEX8 elements is made. MOAB geometry and all skinning logic operate on the internal converted tet mesh. Temperature, density, and subdomain binning are evaluated on the original MOOSE mesh where auxiliary variables live, and the resulting bin assignments are propagated to child tet IDs via the parent map.

## Impact
Simplify workflows. No changes to existing input file syntax. Pure TET4/TET10 workflows are unaffected.

meltawila:
closes #1352 

I used here the same set of tests we already use in /userobjects/moab_skinner/ with TET meshes to make sure they work as expected with HEX8 meshes and will add a test the mixes HEX8 with TET4 elements in the mesh (I'm thinking with `AnnularMeshGenerator`).


include/userobjects/MoabSkinner.h
```diff
@@ -231,6 +233,18 @@ class MoabSkinner : public GeneralUserObject
...
+  bool _requires_tet_conversion;
+
+  /// True when buildTetMesh() is called
+  bool _tet_mesh_built;
+
+  /// TET4 clone of the MOOSE mesh. Present only when _requires_tet_conversion is true.
+  std::unique_ptr<ReplicatedMesh> _tet_mesh;
+
+  /// Maps every _tet_mesh element ID to the original getMooseMesh() element ID it was produced from
+  std::map<dof_id_type, dof_id_type> _tet_to_original_elem;
```
GiudGiud:
```suggestion
  std::unordered_map<dof_id_type, dof_id_type> _tet_to_original_elem;
```
for the lookup scaling

but ideally you'd use a vector of size max_id

in the tet mesh, you should just keep the IDs contiguous

include/userobjects/MoabSkinner.h
```diff
@@ -245,8 +259,22 @@ class MoabSkinner : public GeneralUserObject
...
-  /// Moose mesh
+  /// Get the MooseMesh (displaced or not, depending on _use_displaced)
   MooseMesh & getMooseMesh();
+
+  /**
+   * Return the mesh used for DAGMC geometry construction.
+   * Returns *_tet_mesh when _requires_tet_conversion && _tet_mesh,
+   * otherwise returns getMooseMesh().getMesh().
+   */
+  MeshBase & getGeometryMesh();
```
GiudGiud: add DAGMC to the name?

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +248,89 @@ MoabSkinner::getMooseMesh()
               : _fe_problem.mesh());
 }
 
+MeshBase &
+MoabSkinner::getGeometryMesh()
+{
+  if (_requires_tet_conversion && _tet_mesh)
+    return *_tet_mesh;
```
GiudGiud: can you have one and not the other ?
I 'd mooseassert that case if not

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +248,89 @@ MoabSkinner::getMooseMesh()
...
+  // Decide whether conversion is needed and validate element types
+  bool has_tet10 = false;
+  for (const auto * elem : source.active_element_ptr_range())
+  {
+    if (elem->type() == HEX8)
+      _requires_tet_conversion = true;
+    else if (elem->type() == TET10)
+      has_tet10 = true;
+    else if (elem->type() != TET4)
+      mooseError("MoabSkinner only supports HEX8, TET4, and TET10 meshes, but found element type '",
```
GiudGiud: why the limitation? We worked a bunch on making `convert3DMeshToAllTet4` work with every element type, you could take advantage of that

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +248,89 @@ MoabSkinner::getMooseMesh()
...
+                 "'.");
+  }
+
+  if (!_requires_tet_conversion)
+    return;
+
+  // convert3DMeshToAllTet4() only handles linear elements, so a mesh that mixes HEX8 with
+  // quadratic TET10 elements cannot be converted.
+  if (has_tet10)
+    mooseError("MoabSkinner cannot convert a mesh that mixes HEX8 and TET10 elements. "
```
GiudGiud: you could "all_first_order" it before running the "all_tet" operation

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +248,89 @@ MoabSkinner::getMooseMesh()
...
+               "Use an all-HEX8, all-TET4, or all-TET10 mesh.");
+
+  if (_verbose)
+    _console << "MoabSkinner: non-tetrahedral elements detected. "
+             << "Building internal TET4 copy of the mesh..." << std::endl;
+
+  _tet_mesh = std::make_unique<ReplicatedMesh>(source.comm());
+  _tet_mesh->copy_nodes_and_elements(source);
```
GiudGiud: I think we might have a clone routine instead?

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +248,89 @@ MoabSkinner::getMooseMesh()
...
+  const subdomain_id_type tmp_remove_sid = *sids.rbegin() + 1;
+
+  // convert3DMeshToAllTet4() contracts and re-prepares the mesh internally.
+  std::vector<dof_id_type> converted_ids;
+  MooseMeshElementConversionUtils::convert3DMeshToAllTet4(
+      *_tet_mesh, elems_to_process, converted_ids, tmp_remove_sid, true);
+
+  // Read the stamped original ID back from every surviving tet element.
+  for (const auto * tet : _tet_mesh->active_element_ptr_range())
+    _tet_to_original_elem[tet->id()] = tet->get_extra_integer(orig_id_tag);
```
GiudGiud: delete the extra integer after?

OR
delete this map and just use the extra integers everywhere?

include/userobjects/MoabSkinner.h
meltawila: I removed the map entirely per your comment later. The original element ID now is in `orig_id` element extra integer on the converted mesh

src/userobjects/MoabSkinner.C
meltawila: yea that was redundant. I dropped `_requires_tet_conversion` everywhere

src/userobjects/MoabSkinner.C
meltawila: I removed the restriction and now call `all_first_order()` on the clone

src/userobjects/MoabSkinner.C
meltawila: yep. I replaced this with to `source.clone()`

src/userobjects/MoabSkinner.C
meltawila: I went with the second option!

meltawila:
I think I was initially overrestricting this since now the parent element ID is in `orig_id` element extra integer on the converted mesh with no need to map elements IDs differently for each element type. So this should support HEX, PYRAMID, PRISM, ~~C0POLYHEDRON~~ (not yet) through `convert3DMeshToAllTet4()`. Also `all_first_order()` allows for higher order elements as well. I'm adding a test with mixed HEX8 and PRISM6 elements (and also HEX27 and PRISM18 with `ElementOrderConversionGenerator`) using:
```
[Mesh]
  [cylinder]
    type = AnnularMeshGenerator
    nr = 10
    nt = 20
    rmin = 0
    rmax = 1
    quad_subdomain_id = 2
    tri_subdomain_id = 3
    growth_r = -1.2
  []
  [extrude]
    type = AdvancedExtruderGenerator
    input = cylinder
    heights = '10'
    num_layers = '10'
    direction = '0 0 1'
  []
[]
```
I think any edge cases would be handled through `convert3DMeshToAllTet4` error checks

doc/content/source/userobjects/MoabSkinner.md
```diff
@@ -26,6 +26,16 @@ which to track particles.
...
 
+### Supported Mesh Element Types
+
+`MoabSkinner` natively supports meshes consisting of TET4 and TET10 elements, since DAGMC
+geometry requires triangulated surfaces (i.e. skins of tetrahedra). Meshes containing HEX, PYRAMID, PRISM
+elements are also supported: when these elements are detected, an internal all-TET4 copy of
+the mesh is built automatically, and all skinning
+and geometry operations are performed on that copy. The original MOOSE mesh is never
+modified; temperature, density, and subdomain binning are always evaluated on the original
+mesh where the auxiliary variables live. Higher orders of these elements types are also supported.
```
nuclearkevin:
```suggestion
mesh where the auxiliary variables live. Higher order versions of these element types are also supported.
```

include/userobjects/MoabSkinner.h
```diff
@@ -245,13 +253,25 @@ class MoabSkinner : public GeneralUserObject
     Sense sense;
   };
 
-  /// Moose mesh
+  /// Get the MooseMesh (displaced or not, depending on _use_displaced)
   MooseMesh & getMooseMesh();
+
+  /**
+   * Return the mesh used for DAGMC geometry construction.
+   * Returns *_tet_mesh when the source mesh contained non-tetrahedral elements
```
nuclearkevin:
```suggestion
   * @return *_tet_mesh when the source mesh contained non-tetrahedral elements
```

src/userobjects/MoabSkinner.C
```diff
@@ -244,9 +246,74 @@ MoabSkinner::getMooseMesh()
...
+MeshBase &
+MoabSkinner::getDAGMCGeometryMesh()
+{
+  if (_tet_mesh)
+    return *_tet_mesh;
+  return getMooseMesh().getMesh();
+}
+
+void
+MoabSkinner::buildTetMesh()
```
nuclearkevin, line 258: I'm not a fan of needing to create a TET4 copy of the full MOOSE mesh prior to performing skinning due to the memory penalty. We effectively store three meshes in memory now, which will not scale well when using larger problems with the skinner (e.g. assembly bowing or core flowering problems).

I think the more elegant way to go about this would be to manually triangulate square element faces while skinning instead of creating a full mesh copy. This is obviously a lot of extra work, so I'd accept an issue being opened to track this and coming back to this later. This mirrors what we did for AMR (which originally made a mesh copy as a stopgap while waiting for features to get merged in OpenMC).

test/tests/userobjects/moab_skinner/hex-mesh/tests
```diff
@@ -0,0 +1,78 @@
...
+  []
+  [check_skins_step1]
+    type = Exodiff
+    prereq = convert_to_gmsh_step1
+    input = read_skins1.i
+    exodiff = read_skins1_in.e
+    cli_args = '--mesh-only'
+    requirement = "The system shall properly skin a MOAB mesh and create new MOAB surface meshes bounding bin regions. The bins shall be re-generated on each time step from a hex mesh."
+  []
+  [hex_sphere]
```
nuclearkevin: What is the purpose of this test? It appears to be almost a carbon copy of `bins`, just with a spherical mesh instead of a slab mesh.

src/userobjects/MoabSkinner.C
```diff
@@ -576,27 +646,66 @@ MoabSkinner::sortElemsByResults()
...
+  else
   {
-    const Elem * const elem = getMooseMesh().queryElemPtr(e);
-    if (!elem)
-      continue;
+    // aux vars live on the original mesh; _elem_bins
+    // must hold tet-mesh IDs (keyed the same as _id_to_elem_handles).
+
+    // compute bin index for every original element.
+    std::map<dof_id_type, unsigned int> orig_id_to_bin;
```
nuclearkevin:
```suggestion
    std::unordered_map<dof_id_type, unsigned int> orig_id_to_bin;
```

`std::map` is sorted in key-space, so insertions are quite a bit slower. I'd recommend defaulting to `std::unordered_map` when possible unless you need a sorted data structure.

nuclearkevin (changes requested):

test/tests/userobjects/moab_skinner/hex-mesh/tests
meltawila: This test used to be in `/moab_skinner/errors/tests` as a `RunException` and I wasn't sure if I should just remove the test after the limitation is lifted. But I agree it's not doing anything new compared to the previous tests. I can remvoe it

src/userobjects/MoabSkinner.C
meltawila: I created the issue https://github.com/neams-th-coe/cardinal/issues/1357

nuclearkevin (approved):
LGTM!

PR #1358, opened 2026-06-10
Use superlu_dist LU solver for more stable testing

meltawila:
`tutorials/pincell_multiphysics.openmc` test began failing after the MOOSE submodule update in Cardinal which bumped PETSc. The failure appears to come from very small PETSc version dependent differences in the temperature solution. These differences are below the displayed precision when inspecting the temperature results but they are amplified by OpenMC’s temperature dependent cross section treatment producing a visible k_eff shift and causing the csvdiff failure. 

This PR switches the MOOSE-side solve from hypre/boomeramg to direct LU with superlu_dist making the test less sensitive to PETSc version differences. I see this is also consistent with `tests/bison_coupling` tests. other coupled tests seem to be unaffected with version changes so far


aprilnovak:
Test failure is unrelated.

PR #1360, opened 2026-06-11
Assign OpenMC materials with the skinner

### LINKED ISSUE #1359: Assign OpenMC materials to DAGMC geometries in MoabSkinner
## Reason
`MoabSkinner` currently cannot assign material names to subdomains when generating DAGMC geometry when coupled with OpenMCCellAveragedProblem. In standalone mode `material_names` accepts a list that must match the internal `meshSubdomains()` order which is not obvious for the user making it easy to assign the wrong material to the wrong block. When coupled with OpenMCCellAverageProblem `material_names` is populated automatically by querying a preexisting DAGMC geometry. There is currently no way to generate a DAGMC geometry directly from a MOOSE mesh within Cardinal which is the end goal of this (along with #1286 and #1353)

## Design
Add `material_blocks` parameter to MoabSkinner to be able to use it as:
```
material_blocks = 'fuel clad coolant'
material_names  = 'UO2 Zircaloy H2O'
```
`material_blocks` accepts a list of subdomain names or IDs. `material_names` lists the corresponding OpenMC material names to add to the generated DAGMC geometry groups. The two parameters must be the same length and are used together when coupled with OpenMCCellAverageProblem.

## Impact
Current input syntax is unaffected (standalone use where `material_names` is only used or when coupled with OpenMCCellAverageProblem without using `material_names` at all). The new additions will allow users to update materials assignment from the MoabSkinner

meltawila:
closes #1359


src/userobjects/MoabSkinner.C
```diff
@@ -40,12 +40,16 @@ MoabSkinner::validParams()
...
+  params.addParam<std::vector<SubdomainName>>(
+      "material_blocks",
+      "List of mesh subdomain names for which to assign material names in the generated "
+      "DAGMC geometry. Must be the same length as 'material_names'.");
   params.addParam<std::vector<std::string>>(
       "material_names",
-      "List of names for each subdomain to use for naming the new volumes created in MOAB. "
-      "You only need to set this if using this skinner independent of OpenMC; otherwise, "
-      "these names are auto-deduced from OpenMC");
+      "Material names to assign to subdomains in the generated DAGMC geometry. "
```
aprilnovak: Can you check that the tables with `verbose = true` in `OpenMCCellAverageProblem` for the material <--> block mapping show the correct info? Let's confirm that the void materials show up properly. I'm referring to this code, which I'm not sure what it shows if we're manually setting void to some regions.

```
  if (_cell_to_elem.size())
  {
    _console
        << "\n ===================>  OPENMC SUBDOMAIN MATERIAL MAPPING  <====================\n"
        << std::endl;
    _console << "      Subdomain:  Subdomain name; if unnamed, we show the ID" << std::endl;
    _console << "       Material:  OpenMC material name(s) in this subdomain; if unnamed, we\n"
             << "                  show the ID. If N duplicate material names, we show the\n"
             << "                  number in ( ).\n"
             << std::endl;
    vt.print(_console);
    _console << std::endl;
  }
```

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
-        parameters(), "material_names", "using skinner independent of an OpenMC [Problem]");
-    _material_names = getParam<std::vector<std::string>>("material_names");
+    checkRequiredParam(parameters(), "material_names", "specifying 'material_blocks'");
+    auto block_names = getParam<std::vector<SubdomainName>>("material_blocks");
+    auto mat_names = getParam<std::vector<std::string>>("material_names");
 
-    if (_material_names.size() != _n_block_bins)
+    if (block_names.size() != mat_names.size())
+      paramError("material_names",
+                 "'material_names' must be the same length as 'material_blocks' (" +
```
aprilnovak: we can make the error even more explicit by also showing the length of `material_names` here. I find this helpful when input files are filled by `${}` type syntax where it might not be obvious what the length of something is.

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
+      unassigned_blocks.push_back(getMooseMesh().getSubdomainName(subdomain_id));
+
+  if (!unassigned_blocks.empty())
+    mooseWarning("The following mesh subdomains have no material assignment in MoabSkinner "
+                 "and will default to void in OpenMC:\n  ",
+                 Moose::stringify(unassigned_blocks, "\n  "));
+
+  if (_verbose)
+  {
+    VariadicTable<std::string, std::string, std::string> vt(
```
aprilnovak: I see this printing here now - is it possible to fuse this with the table already being printed in `OpenMCCellAverageProblem` about the materials? I think they are showing the same info, since if we are ultimately going to be creating a DAGMC model from Cardinal from scratch, then the entirety of the mesh represents the entirety of the OpenMC model (and hence is equivalent to the entirety of the cell <--> elem mapping we establish). LMK if I am overlooking something with this suggestion

test/tests/neutronics/dagmc/assign_materials/tests
```diff
@@ -0,0 +1,21 @@
+[Tests]
```
aprilnovak, line 1: Let's add `RunException` tests here for the various errors you throw.

Can you explain to me what these tests are doing? Perhaps you can add to the `requirement` flag exactly what is being tested? e.g. are you loading these models which get built by Cardinal, into some other software to inspect material assignments? Can we do better than a `RunApp` type test for `generate_dagmc`? Those are easy to miss regressions in the future if a failure mode is run-but-hidden-error

src/userobjects/MoabSkinner.C
```suggestion
      "Material names (or IDs) to assign to subdomains in the generated DAGMC geometry. "
```

aprilnovak: Can you please add a test to cover this?

src/userobjects/MoabSkinner.C
```diff
@@ -40,12 +40,16 @@ MoabSkinner::validParams()
   params.addParam<Real>("density_max", "Upper bound of density bins");
   params.addRangeCheckedParam<unsigned int>(
       "n_density_bins", "n_density_bins > 0", "Number of density bins");
-
+  params.addParam<std::vector<SubdomainName>>(
+      "material_blocks",
+      "List of mesh subdomain names for which to assign material names in the generated "
```
```suggestion
      "List of mesh subdomain names (or IDs) for which to assign material names in the generated "
```

aprilnovak: Can you add a test to cover this?

If we feel like we're repeating a lot of similar tests across different objects, we could add some helper functions to `CardinalProblem` or somewhere else appropriate, to reduce any notion of duplicate testing.

src/userobjects/MoabSkinner.C
meltawila: The test I have currently covers this with `material_names = 'h2o 1'` which is only because the .xml file I'm reusing from another test here only has material IDs for some materials with no names. Did you prefer separating the testing where one test only uses material names and the other only uses material IDs? or is this enough?

aprilnovak: this is enough! I didn't look super closely at your tests. It would be helpful to add something short to the `requirements` about the fact that this test is indeed covering the dual options for `material_names`

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
+                     Moose::stringify(block_names.size()) + ")");
+
+    for (std::size_t i = 0; i < block_names.size(); ++i)
+    {
+      if (!MooseMeshUtils::hasSubdomainName(getMooseMesh().getMesh(), block_names[i]))
+        paramError("material_blocks",
+                   "Subdomain '" + std::string(block_names[i]) + "' not found in the mesh");
+      auto id = getMooseMesh().getSubdomainID(block_names[i]);
+      auto it = _blocks.find(id);
+      _block_id_to_material_name[it->first] = mat_names[i];
```
GiudGiud, line 363: could also check here that the user has not passed a repeated entry for blocks

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
+    auto block_names = getParam<std::vector<SubdomainName>>("material_blocks");
+    auto mat_names = getParam<std::vector<std::string>>("material_names");
 
-    if (_material_names.size() != _n_block_bins)
+    if (block_names.size() != mat_names.size())
+      paramError("material_names",
+                 "'material_names' must be the same length as 'material_blocks' (" +
+                     Moose::stringify(block_names.size()) + ")");
+
+    for (std::size_t i = 0; i < block_names.size(); ++i)
```
GiudGiud:
```suggestion
    for (const auto i : index_range(block_names))
```

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
 
-  if (_standalone)
+  if (isParamValid("material_blocks"))
   {
-    checkRequiredParam(
-        parameters(), "material_names", "using skinner independent of an OpenMC [Problem]");
-    _material_names = getParam<std::vector<std::string>>("material_names");
+    checkRequiredParam(parameters(), "material_names", "specifying 'material_blocks'");
+    auto block_names = getParam<std::vector<SubdomainName>>("material_blocks");
+    auto mat_names = getParam<std::vector<std::string>>("material_names");
```
GiudGiud:
```suggestion
    const auto & block_names = getParam<std::vector<SubdomainName>>("material_blocks");
    const auto & mat_names = getParam<std::vector<std::string>>("material_names");
```
no need to copy them

src/userobjects/MoabSkinner.C
```diff
@@ -248,25 +252,81 @@ void
...
+
+  if (_verbose)
+  {
+    VariadicTable<std::string, std::string, std::string> vt(
+        {"Subdomain ID", "Subdomain Name", "Material"});
+    for (const auto & [subdomain_id, block_index] : _blocks)
+    {
+      auto name = getMooseMesh().getSubdomainName(subdomain_id);
+      auto mat_it = _block_id_to_material_name.find(subdomain_id);
+      auto mat = (mat_it != _block_id_to_material_name.end()) ? mat_it->second : "(void)";
```
GiudGiud:
```suggestion
      const auto & name = getMooseMesh().getSubdomainName(subdomain_id);
      const auto & mat_it = _block_id_to_material_name.find(subdomain_id);
      const auto & mat = (mat_it != _block_id_to_material_name.end()) ? mat_it->second : "(void)";
```

src/userobjects/MoabSkinner.C
meltawila: The current implementation with verbose=true in my test will print the below:
 

```
===================>  OPENMC SUBDOMAIN MATERIAL MAPPING  <====================

              Subdomain:  Subdomain name; if unnamed, we show the ID
               Material:  OpenMC material name(s) in this subdomain; if unnamed, we
                          show the ID. If N duplicate material names, we show the
                          number in ( ).

------------------------
| Subdomain | Material |
------------------------
| block_1   |  1       |
| block_2   |  2       |
------------------------


Postprocessor Values:
+----------------+----------------+----------------+
| time           | k              | k_std_dev      |
+----------------+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |
+----------------+----------------+----------------+

Skinning geometry into 4 temperature bins, 1 density bins, and 2 block bins... 

MoabSkinner updated material assignments:
--------------------------------------------
| Subdomain ID | Subdomain Name | Material |
--------------------------------------------
| 1            | block_1        | h2o      |
| 2            | block_2        | 1        |
--------------------------------------------


...
...

 ===================>  OPENMC SUBDOMAIN MATERIAL MAPPING  <====================

              Subdomain:  Subdomain name; if unnamed, we show the ID
               Material:  OpenMC material name(s) in this subdomain; if unnamed, we
                          show the ID. If N duplicate material names, we show the
                          number in ( ).

------------------------
| Subdomain | Material |
------------------------
| block_1   |  h2o     |
| block_2   |  1       |
------------------------

Sending temperature to OpenMC cells... 
Setting cell id 1, instance 0 (of 1) [1 contained cells] to temperature (K): 575 
Setting cell id 2, instance 0 (of 1) [1 contained cells] to temperature (K): 675 
 Running OpenMC with 1000 particles per batch...
 Maximum neutron transport energy: 20000000 eV for U235
 Initializing source particles...

 ====================>     K EIGENVALUE SIMULATION     <====================
```
I'll add a test that assigns void material to make sure it's consistent, but otherwise I think this console printout is useful to see where material assignment is getting updated instead of relying on the subdomain to material mapping only

src/userobjects/MoabSkinner.C
meltawila: The current implementation with verbose=true in my test will print the below:

```
===================>  OPENMC SUBDOMAIN MATERIAL MAPPING  <====================

              Subdomain:  Subdomain name; if unnamed, we show the ID
               Material:  OpenMC material name(s) in this subdomain; if unnamed, we
                          show the ID. If N duplicate material names, we show the
                          number in ( ).

------------------------
| Subdomain | Material |
------------------------
| block_1   |  1       |
| block_2   |  2       |
------------------------


...

Skinning geometry into 4 temperature bins, 1 density bins, and 2 block bins... 

MoabSkinner updated material assignments:
--------------------------------------------
| Subdomain ID | Subdomain Name | Material |
--------------------------------------------
| 1            | block_1        | h2o      |
| 2            | block_2        | 1        |
--------------------------------------------


...
...

 ===================>  OPENMC SUBDOMAIN MATERIAL MAPPING  <====================

              Subdomain:  Subdomain name; if unnamed, we show the ID
               Material:  OpenMC material name(s) in this subdomain; if unnamed, we
                          show the ID. If N duplicate material names, we show the
                          number in ( ).

------------------------
| Subdomain | Material |
------------------------
| block_1   |  h2o     |
| block_2   |  1       |
------------------------

Sending temperature to OpenMC cells... 
Setting cell id 1, instance 0 (of 1) [1 contained cells] to temperature (K): 575 
Setting cell id 2, instance 0 (of 1) [1 contained cells] to temperature (K): 675 
 Running OpenMC with 1000 particles per batch...
 Maximum neutron transport energy: 20000000 eV for U235
 Initializing source particles...

 ====================>     K EIGENVALUE SIMULATION     <====================
```
I think this console printout can be useful to see where material assignment is getting updated but in principle yes the subdomain to material mapping table is showing the same information

src/userobjects/MoabSkinner.C
meltawila: Added. this part is now:
```
      auto id = getMooseMesh().getSubdomainID(block_names[i]);
      if (seen_ids.count(id))
        paramError("material_blocks",
                   "Subdomain '" + std::string(block_names[i]) +
                       "' refers to the same mesh subdomain as another entry in 'material_blocks'");
      seen_ids.insert(id);
      auto it = _blocks.find(id);
      _block_id_to_material_name[it->first] = mat_names[i];
```

test/tests/neutronics/dagmc/assign_materials/tests
meltawila: I added the RunException tests for the different scenarios. generate_dagmc.i is FEProblem/solve=false input file and is only used to generate an initial .h5m geometry file (no externally generated .h5m geometry is provided to this test) so it's a first step for the openmc.i file which will modify the material assignment in that geometry again before running openmc on it.

What I'm planning to do in another PR is to add a command line argument that could allow running `-i openmc.i --generate-dagmc-only` which would bypass the Problem block and run the skinner only which is effectively what I'm doing here with the two input files.

src/userobjects/MoabSkinner.C
aprilnovak: > I think this console printout is useful to see where material assignment is getting updated instead of relying on the subdomain to material mapping only

Can you elaborate more on what *different* information is shown by the new table compared to what is already printed? From this example it seems to show the same information

meltawila: mainly reporting that _MoabSkinner updated material assignments_ part. maybe I should only keep this statement without the table?

aprilnovak: if there's not any new info, then I lean towards reducing console output (to avoid de-sensitizing the user to loads of things printing). Can we make it print out just

`MoabSkinner updating material assignments...`

include/userobjects/MoabSkinner.h
```diff
@@ -478,4 +478,7 @@ class MoabSkinner : public GeneralUserObject
 
   /// Tolerance to use for comparing values to bin bounds
   const Real BIN_TOLERANCE = 1e-6;
+
+  /// Map from mesh SubdomainID to DAGMC material name
```
aprilnovak:
```suggestion
  /// Map from mesh SubdomainID to OpenMC material name
```

src/userobjects/MoabSkinner.C
```diff
@@ -41,12 +41,16 @@ MoabSkinner::validParams()
...
+      "List of mesh subdomain names (or IDs) for which to assign material names in the generated "
+      "DAGMC geometry. Must be the same length as 'material_names'.");
   params.addParam<std::vector<std::string>>(
       "material_names",
-      "List of names for each subdomain to use for naming the new volumes created in MOAB. "
-      "You only need to set this if using this skinner independent of OpenMC; otherwise, "
-      "these names are auto-deduced from OpenMC");
+      "Material names (or IDs) to assign to subdomains in the generated DAGMC geometry. "
+      "Must be the same length as 'material_blocks' and listed in the same order. "
+      "Any subdomain not listed in 'material_blocks' will have no material assignment "
```
aprilnovak: I feel like this "Any subdomain..." sentence should go in the docstring for `material_blocks` since it's referring to subdomains.

src/userobjects/MoabSkinner.C
```diff
@@ -85,6 +89,8 @@ MoabSkinner::validParams()
                         "Whether the skinned mesh should be generated from a displaced mesh ");
   params.addClassDescription("Re-generate the OpenMC geometry on-the-fly according to changes in "
                              "the mesh geometry and/or contours in temperature and density");
+  ExecFlagEnum & exec_enum = params.set<ExecFlagEnum>("execute_on", true);
+  exec_enum = {EXEC_TIMESTEP_BEGIN};
```
aprilnovak: how does this interact with `OpenMCCellAverageProblem`, which would otherwise control when this object gets executed? Can you add a comment here describing why this is needed? In general, setting these exec flags manually can be somewhat error-prone

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +321,75 @@ MoabSkinner::initialize()
...
   findBlocks();
+  _block_id_to_material_name.clear();
 
-  if (_standalone)
+  if (isParamValid("material_blocks"))
   {
-    checkRequiredParam(
-        parameters(), "material_names", "using skinner independent of an OpenMC [Problem]");
-    _material_names = getParam<std::vector<std::string>>("material_names");
+    checkRequiredParam(parameters(), "material_names", "specifying 'material_blocks'");
```
aprilnovak: the way this is written, I think you're not covering the possibility that the user provides `material_names` but NOT `material_blocks`. Suggest you have this error check outside this if-statement, like

```
if (isParamValid("material_blocks") != isParamValid("material_names"))
  // error

if (isParamValid("material_blocks"))
  // stuff
```

You might need to put the first if-statment combined with an `if (_standalone)` for this to be correct?

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +321,75 @@ MoabSkinner::initialize()
...
+      if (!MooseMeshUtils::hasSubdomainName(getMooseMesh().getMesh(), block_names[i]))
+        paramError("material_blocks",
+                   "Subdomain '" + std::string(block_names[i]) + "' not found in the mesh");
+      auto id = getMooseMesh().getSubdomainID(block_names[i]);
+      if (seen_ids.count(id))
+        // A repeated entry for blocks (including if the same block is passed by its name then later
+        // by its ID) should error here
+        paramError("material_blocks",
+                   "Subdomain '" + std::string(block_names[i]) +
+                       "' refers to the same mesh subdomain as another entry in 'material_blocks'");
```
```suggestion
                       "' was already listed in 'material_blocks' (either by name or ID). Each block can only be listed one time.");
```

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +321,75 @@ MoabSkinner::initialize()
     buildTetMesh();
 
   findBlocks();
+  _block_id_to_material_name.clear();
 
-  if (_standalone)
+  if (isParamValid("material_blocks"))
```
aprilnovak: should there be an unused-parameter warning if `_standalone` is true, but `material_blocks` is provided?

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +321,75 @@ MoabSkinner::initialize()
...
+    _block_id_to_material_name.clear();
+    for (const auto & [subdomain_id, block_index] : _blocks)
+      _block_id_to_material_name[subdomain_id] = mat_names[block_index];
   }
   else
   {
-    // the external class should have set the value of _material_names
+    // .h5m was prepared outside of Cardinal; _material_names was set by setMaterialNames().
+    // Build the same existing map.
     checkUnusedParam(parameters(),
```
aprilnovak: do we need an unused parameter check for `material_blocks`?

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +321,75 @@ MoabSkinner::initialize()
...
                      "material_names",
-                     "using the skinner in conjunction with an OpenMC [Problem]");
+                     "using the skinner in conjunction with an OpenMCCellAverageProblem "
+                     "without 'material_blocks' overrides");
+    _block_id_to_material_name.clear();
```
aprilnovak: lines 378-380 here are the same as lines 366-368 - can you condense the code to reduce the duplication?

src/userobjects/MoabSkinner.C
```diff
@@ -806,7 +863,18 @@ MoabSkinner::materialName(const unsigned int & block,
...
+    if (block_index == block)
+    {
+      auto it = _block_id_to_material_name.find(subdomain_id);
+      if (it != _block_id_to_material_name.end())
+        return "mat:" + it->second;
+
+      return "mat:" + _material_names.at(block);
+    }
+  }
+  mooseError("materialName: could not find subdomain for block index ", block);
```
aprilnovak: is it possible to write a test to trigger this? or is this just additional error checking to be safe?

test/tests/neutronics/dagmc/assign_materials/tests
```diff
@@ -0,0 +1,59 @@
...
+    expect_err = "Subdomain 'nonexistent' not found in the mesh"
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+    requirement = "The system shall error if a block listed in 'material_blocks' does not exist in the mesh"
+  []
+  [material_blocks_duplicate]
+    type = RunException
+    input = generate_dagmc.i
+    cli_args = "UserObjects/moab/material_blocks='1 block_1' UserObjects/moab/material_names='1 2'"
+    expect_err = "Subdomain 'block_1' refers to the same mesh subdomain as another entry in 'material_blocks'"
```
aprilnovak: if you accept the change in the source code, this error message will need to change text slightly

test/tests/neutronics/dagmc/assign_materials/tests
```diff
@@ -0,0 +1,59 @@
+[Tests]
```
aprilnovak, line 1: Let's add a RunException type test (with `cli_args = '--error'`) for the warning that gets printed if there is a subdomain not listed in `material_blocks`, to confirm that warning about void fill gets printed

test/tests/neutronics/dagmc/assign_materials/generate_dagmc.i
```diff
@@ -0,0 +1,61 @@
...
+    input = block_1
+    combinatorial_geometry = 'x >= 37.5'
+    block_id = 2
+    block_name = block_2
+    epsilon = 1e-3
+  []
+  allow_renumbering = false
+[]
+
+[AuxVariables]
```
aprilnovak, line 25: for this test to be effective, do we need to have temperature binning? if we can make it simpler, that'll be easier to convey to people how this object is intended to be used.

test/tests/neutronics/dagmc/assign_materials/generate_dagmc.i
```diff
@@ -0,0 +1,61 @@
...
+  [temp]
+    order = CONSTANT
+    family = MONOMIAL
+    [AuxKernel]
```
aprilnovak: instead of this, why not

```
[AuxVariables]
  [temp]
    order = CONSTANT
    family = MONOMIAL
    initial_condition = 200
  []
[]
```

test/tests/neutronics/dagmc/assign_materials/tests
```diff
@@ -0,0 +1,59 @@
...
+  [generate_dagmc]
+    type = RunApp
+    input = generate_dagmc.i
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+    requirement = "The system shall assign materials to MOOSE mesh blocks and generate a new DAGMC geometry"
+                  "with the correct material group tags"
+  []
+  [run_updated_geometry]
+    type = CSVDiff
```
aprilnovak, line 68: Can you explain to me in more detail what this test is doing? You generate the `moab_mesh_0.h5m` file from the test above. Then, this gets used as geometry in this test. Instead of testing for k-effective, why not do two `PointValue` postprocessors that fetch the value of https://cardinal.cels.anl.gov/source/auxkernels/CellMaterialIDAux.html? That would give the material ID at that location, which I think would be perfect for ensuring correct behavior? (since we can only infer correct behavior through k-effective, which might be harder to come back to in the future and understand what this test was doing).

test/tests/neutronics/dagmc/assign_materials/tests
```diff
@@ -0,0 +1,59 @@
...
+    prereq = generate_dagmc
+    input = openmc.i
+    csvdiff = openmc_out.csv
+    cli_args = 'Outputs/hide="cell_instance"'
+    requirement = "The system shall update material assignments with 'material_blocks' and 'material_names' "
+                  "(using names or IDs) and run OpenMC with the regenerated geometry"
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+  []
+[]
```
aprilnovak, line 88: The tests don't seem to cover the case where a name is provided in `material_names`. Can you just modify the `generate_dagmc.i` case to use a name for material 2? (and edit the `materials.xml`)

test/tests/neutronics/dagmc/assign_materials/tests
aprilnovak: and, can we add a test that specifically checks that void goes into unlisted subdomains?

src/userobjects/MoabSkinner.C
meltawila: When coupled with `OpenMCCellAverageProblem`, the problem drives this object directly by calling `initialize()` and `update()` when needed and flags it as `_standalone = false` in `makeDependentOnExternalAction()` during `initialSetup()`. so we are not relying on the execution flags to drive skinning. updating this default was my attempt to control the redundant executions coming from the regular MOOSE UserObject execution order. But I agree both TIMESTEP_END or TIMESTEP_BEGIN default would trigger a redundant `MoabSkinner::initialize()` similarly and doesn't solve the issue (I mistakenly thought it did earlier while adding some print statements). I was also trying to update `_execute_enum` in `makeDependentOnExternalAction` but that didn't seem to change the behavior either:
```
  virtual void makeDependentOnExternalAction()
  {
    _standalone = false;
    const_cast<ExecFlagEnum &>(_execute_enum) = EXEC_NONE;
  }
```
I'll remove these two lines since they don't solve the redundant execution issue but it still persists unless you have a solution for this?

meltawila:
```
ExecFlagEnum & exec_enum = params.set<ExecFlagEnum>("execute_on", true);
exec_enum = EXEC_NONE;
```
would fix this but will affect the current standalone behavior

src/userobjects/MoabSkinner.C
meltawila: `material_names` without `material_blocks` is currently accepted in standalone mode. I restructured this if statement to address following comments, but still kept this possibility to keep the current behavior unless you'd like to change that. I believe all standalone tests provides `material_names` but not `material_blocks`. will push this change shortly

src/userobjects/MoabSkinner.C
meltawila: I believe the current code does the right thing when `_standalone` is true and `material_blocks` is provided. this is pretty much what `generate_dagmc` test is doing. in standalone if `material_blocks` is provided it'll be used, if not, subdomain IDs in ascending order are used (the current behavior)

src/userobjects/MoabSkinner.C
meltawila: I don't think so unless I'm missing something. when it exists it's always used, and when it doesn't exist the map is either built similar to the previous step using setMaterialNames() or following IDs ascending order (both are the current code behavior)

src/userobjects/MoabSkinner.C
meltawila: I changed this to mooseAssert. I don't think I can trigger this error

test/tests/neutronics/dagmc/assign_materials/generate_dagmc.i
meltawila: Temperature binning can't be removed entirely (`temperature`, `temperature_max`, `n_temperature_bins` are required params)

aprilnovak: perhaps in the future it would make sense to have these parameters be optional - it's kind of clunky syntax the way you are using the skinner here.

meltawila: In fact after updating my test to use CellMaterialIDAux (which apparently according to documentation was intended for fluids only) I had to add density blocks and density binning to make it work in my tests.
"_If a MOOSE element is not fluid or did not map at all to an OpenMC cell, then this auxiliary kernel returns −1_"

test/tests/neutronics/dagmc/assign_materials/tests
meltawila: Updated the test to use CellMaterialIDAux but I had to add density binning to make it work

aprilnovak: yeah, that makes sense - I still think this test is a bit more robust. Thanks!

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +318,78 @@ MoabSkinner::initialize()
...
+      _block_id_to_material_name[it->first] = mat_names[i];
+    }
   }
   else
   {
-    // the external class should have set the value of _material_names
-    checkUnusedParam(parameters(),
-                     "material_names",
-                     "using the skinner in conjunction with an OpenMC [Problem]");
+    if (_standalone)
```
aprilnovak, line 371: should there be a 

```
checkUnusedParam(parameters(), "material_blocks", "using skinner independent of an OpenMCCellAverageProblem");
```

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
...
+      "OpenMCCellAverageProblem"
+      "where material names are deduced from OpenMC.");
   params.addParam<std::vector<std::string>>(
       "material_names",
-      "List of names for each subdomain to use for naming the new volumes created in MOAB. "
-      "You only need to set this if using this skinner independent of OpenMC; otherwise, "
-      "these names are auto-deduced from OpenMC");
-
+      "Material names (or IDs) to assign to subdomains in the generated DAGMC geometry. "
+      "Must be provided together with 'material_names' where both have the same length"
```
```suggestion
      "Must be provided together with 'material_blocks' where both have the same length"
```

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
   params.addParam<Real>("density_max", "Upper bound of density bins");
   params.addRangeCheckedParam<unsigned int>(
       "n_density_bins", "n_density_bins > 0", "Number of density bins");
-
+  params.addParam<std::vector<SubdomainName>>(
+      "material_blocks",
+      "List of mesh subdomain names (or IDs) for which to assign material names in the generated "
+      "DAGMC geometry. Must be provided together with 'material_names' where both have the same "
+      "length."
```
```suggestion
      "length and are listed in the same order."
```

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
...
+      "Any subdomain not listed in 'material_blocks' will have no material assignment and OpenMC "
+      "will"
+      "default to void for that region. This parameter is optional when coupled with "
```
```suggestion
      "default to void for that region. This parameter is optional when used in combination with "
```

src/userobjects/MoabSkinner.C
meltawila: Ops sorry

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
...
+      "OpenMCCellAverageProblem"
+      "where material names are deduced from OpenMC.");
```
```suggestion
      "because in this case, the default is to auto-detect the materials in the original h5m file and re-apply them during skinning.");
```

aprilnovak: Feel free to adjust these requested changes - just trying to make this documentation a bit clearer

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
...
   params.addParam<std::vector<std::string>>(
       "material_names",
-      "List of names for each subdomain to use for naming the new volumes created in MOAB. "
-      "You only need to set this if using this skinner independent of OpenMC; otherwise, "
-      "these names are auto-deduced from OpenMC");
-
+      "Material names (or IDs) to assign to subdomains in the generated DAGMC geometry. "
+      "Must be provided together with 'material_names' where both have the same length"
+      "and listed in the same order. This parameter is optional when coupled with "
```
```suggestion
      "and listed in the same order. This parameter is optional when used in combination with "
```

src/userobjects/MoabSkinner.C
```diff
@@ -41,13 +41,22 @@ MoabSkinner::validParams()
...
+      "OpenMCCellAverageProblem where material names are deduced from OpenMC.");
```
```suggestion
      "OpenMCCellAverageProblem because in this case, the default is to auto-detect the materials in the original h5m file and re-apply them during skinning.");
```

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +324,75 @@ MoabSkinner::initialize()
     buildTetMesh();
```
aprilnovak, line 327: Can we add a more detailed description of the new options to `MoabSkinner.md`? I'm thinking of an explicit explanation like the following.

```
## Material Names and Blocks

For the vast majority of use cases, you will be using the `MoabSkinner` in tandem with `OpenMCCellAverageProblem`. In this case, you can rely on OpenMC to auto-detect the mapping between materials and subdomains, so that when the geometry is re-skinned, OpenMC will re-apply the original materials corresponding to the regions of space. However, if you'd like to instead force the use of a _different_ material from what was present in the original `.h5m` file, you can provide `material_names` and `material_blocks`, to have the skinner apply specific settings of materials on each block.

The `MoabSkinner` is used separately of `OpenMCCellAverageProblem` only for testing purposes; in this case, the user must provide `material_blocks` and `material_names`.
```

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +324,75 @@ MoabSkinner::initialize()
...
 
-  if (_standalone)
+  const bool has_material_blocks = isParamValid("material_blocks");
+  const bool has_material_names = isParamValid("material_names");
+  if (has_material_blocks != has_material_names)
   {
-    checkRequiredParam(
-        parameters(), "material_names", "using skinner independent of an OpenMC [Problem]");
-    _material_names = getParam<std::vector<std::string>>("material_names");
+    if (has_material_blocks)
```
aprilnovak: We can combine these two together like

```
mooseError("'material_blocks' and 'material_names' must both be provided if specifying materials for the skinner; you have provided only one of these parameters");
```

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +324,75 @@ MoabSkinner::initialize()
...
-    checkUnusedParam(parameters(),
-                     "material_names",
-                     "using the skinner in conjunction with an OpenMC [Problem]");
+    // Neither 'material_blocks' nor 'material_names' was supplied in the input.
+    // This is only valid when the skinner is controlled externally, e.g. an .h5m was
+    // prepared outside of Cardinal and material names is set in setMaterialNames().
+    if (_standalone)
+      paramError("material_blocks",
+                 "Standalone MoabSkinner requires both 'material_blocks' and "
+                 "'material_names'. Omit both parameters only when DAGMC geometry is"
```
aprilnovak: Correct me if I'm mistaken - but I thought if `_standalone` is true, that we need both material_blocks and material_names? So I don't understand the "Omit both..." instruction

meltawila: you're correct. the "Omit both ..." instruction here assumes what you mentioned earlier that for the vast majority of use cases, a user will be using the `MoabSkinner` in tandem with `OpenMCCellAverageProblem` and Cardinal will auto detect material names from the provided .h5m. In that case if a user still provided `material_blocks`, I'd want to tell them to remove the parameter entirely instead of telling them to list all `material_names` and `material_blocks`. at least that's what I thought when I wrote this

meltawila: Aaah nevermind this is inside `if (_standalone)` :|

aprilnovak: I'd remove the `Omit both...` thing here from this error message, and instead rely on the `MoabSkinner.md` page to help explain all the use cases. I generally default to throwing error messages that only pertain to the use case where the error was trigger, vs. explaining all the methods by which the parameters can be combined.

src/userobjects/MoabSkinner.C
ssgudala-bwxt: This might be just an issue with semantics, but I am using MoabSkinner to generate the *initial* DAGMC mesh from an Exodus mesh, before including it in an OpenMCCellAverageProblem. In this case, the "standalone" mode is not purely for testing purposes.

meltawila: yes I agree. this is also what I'm doing here in `test/tests/neutronics/dagmc/assign_materials/generate_dagmc.i‎` in this PR. after this one gets merged I'd like to have a command line argument that overrides the [Problem] and generate the dagmc only. so a user doesn't need to have 2 input files with the same geometry where one is generating the initial dagmc and the other runs an OpenMCCellAverageProblem

ssgudala-bwxt: Ok sounds good. That would be the ideal workflow as a user.

src/userobjects/MoabSkinner.C
meltawila: this check would be a dead code with the most recent commits. we check material_blocks/material_names are both provided above before getting here

src/userobjects/MoabSkinner.C
```diff
@@ -315,27 +326,66 @@ MoabSkinner::initialize()
...
   {
-    // the external class should have set the value of _material_names
-    checkUnusedParam(parameters(),
-                     "material_names",
-                     "using the skinner in conjunction with an OpenMC [Problem]");
+    // Neither 'material_blocks' nor 'material_names' was supplied in the input.
+    // This is only valid when the skinner is controlled externally, e.g. an .h5m was
+    // prepared outside of Cardinal and material names is set in setMaterialNames().
+    if (_standalone)
+      paramError("material_blocks",
```
aprilnovak: ideally this would be two separate paramError, so that we point to the faulty parameter (it could be confusing if `material_blocks` is specified by the other one is missing

aprilnovak:
Thanks @meltawila! I'll be ready to merge after that one `paramError` is made into two (just in the interest of clarity). Otherwise looks great

src/userobjects/MoabSkinner.C
```diff
@@ -365,13 +365,18 @@ MoabSkinner::initialize()
...
     // This is only valid when the skinner is controlled externally, e.g. an .h5m was
     // prepared outside of Cardinal and material names is set in setMaterialNames().
     if (_standalone)
+    {
       paramError("material_blocks",
-                 "Standalone MoabSkinner requires both 'material_blocks' and "
-                 "'material_names'.");
+                 "'material_blocks' is required when using the MoabSkinner independent of an "
+                 "OpenMCCellAverageProblem");
+      paramError("material_names",
```
aprilnovak: you'll want to only throw the paramError if that parameter has been neglected though - right now, this would still throw both even if the user forgot just 1?

aprilnovak:
```
if (forgot material_names)
  throw_error_for_material_names
if (forgot material_blocks)
  throw_error_for_material_blocks
```

meltawila: Actually looking back at this I think I should ~~revert back to my original paramError before the last commit~~ have only one mooseError that says both parameters are missing? the short version of this part is:
```
  ...
  const bool has_material_blocks = isParamValid("material_blocks");
  const bool has_material_names = isParamValid("material_names");
  if (has_material_blocks != has_material_names)
    ERROR -- both are needed together

  if (has_material_blocks)
  {
    ...
  }
  else
  {
    // >>> If we get here this means we know that has_material_blocks == has_material_names and also 
    // >>> that !has_material_blocks -- so we should error for material blocks parameters
    if (_standalone)
      ERROR -- neither is provided
  }
```
there's no scenario where a user can reach the forgot material_names error:
- if provided both parameters -> normal execution
- if provided only one -> `mooseError("'material_blocks' and 'material_names' must both be provided if specifying "
               "materials for the skinner; you have provided only one of these parameters");`
- if provided neither -> `paramError("material_blocks", "Standalone MoabSkinner requires both 'material_blocks' and 'material_names'.");`.  [doesn't need to specify which one is missing because both are]

aprilnovak: Oh, I see - yeah you are right. Sorry about that! Getting confused with all the branches

aprilnovak (approved):

PR #1365, opened 2026-06-22
Openmc mesh generator

JoffreyDorville:
Hi everyone!

This PR adds a new OpenMC mesh generator to create a mesh mirror from an OpenMC regular mesh knowing its OpenMC mesh ID.

### Context

I am using Cardinal to transfer temperatures to OpenMC with a custom temperature field (regular mesh + vector of temperature values) on the OpenMC side. The multiphysics coupling seems to work well with my local prototype so I am working on cleaning the different features I have to submit them as PRs:

1. OpenMC initialization action (#1302)
2. OpenMC mesh generator (this PR)
3. Temperature transfer to OpenMC temperature field (TBD)

### Implementation

A new `OpenMCMeshGenerator` has been implemented. It relies on the generic MOOSE mesh generator, with the subgenerator logic, to create the mesh mirror. It only takes the mesh ID declared in the OpenMC model to dynamically cast it and retrieve the geometric data needed for its reconstruction on the Cardinal side (dimension, lower left and upper right). An `xml_directory` argument is available to select a custom OpenMC model path. A `scaling` argument has also been added to accommodate different units/scale between Cardinal and OpenMC.

### Limitations

It is currently limited to the regular mesh type of OpenMC, but could be extended to other structured types if needed. For unstructured meshes, we could probably just read the same input file that OpenMC is reading.


nuclearkevin (changes requested):
Thanks for implementing this mesh generator @JoffreyDorville! Just a few notes before I start reviewing the changes more thoroughly:

- `ElementIDAux` and `OpenMCMeshGenerator` are missing documentation markdown files.
- Cardinal's SQA process requires that every PR implementing a major feature or bugfix has a corresponding issue. 

Once you've added documentation and opened a feature issue, the tests added here need to be tagged with `design = 'CLASSNAME.md` and `issue = #ISSUE_NUMBER`. See https://github.com/neams-th-coe/cardinal/blob/devel/test/tests/neutronics/foms/tests for an example of how to go about this.

doc/content/source/auxkernels/ElementIDAux.md
```diff
@@ -0,0 +1,34 @@
+# ElementIDAux
+
+!syntax description /AuxKernels/ElementIDAux
+
+## Description
+
+Returns the current element ID from the mesh.
```
aprilnovak: It doesn't *have* to happen in this PR, but IMO this object makes more sense to be at the MOOSE framework level. Any objections to opening a PR there for `ElementIDAux`? The advantage is that the testing (and maintaining the testing) for this object would be handled by the MOOSE team.

src/auxkernels/ElementIDAux.C
```diff
@@ -0,0 +1,37 @@
...
+  params.addClassDescription("Displays the current mesh element ID.");
+  return params;
+}
+
+ElementIDAux::ElementIDAux(const InputParameters & parameters) : AuxKernel(parameters) {}
+
+Real
+ElementIDAux::computeValue()
+{
+  return static_cast<Real>(_current_elem->id());
```
aprilnovak: do you need the cast? It would happen implicitly when returning the function anyways

doc/content/source/auxkernels/ElementIDAux.md
```diff
@@ -0,0 +1,34 @@
...
+
+Associated with a `PointValueSampler`, it allows to determine which mesh element
```
aprilnovak: This sentence sounds like it is referring to a workflow that includes `ElementIDAux` as one piece - I don't think it's needed here to solely explain `ElementIDAux` though, so I'd suggest to remove this sentence.

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,98 @@
...
+  params.addClassDescription(
+      "A class that generates a mesh from an available OpenMC regular mesh.");
+  params.addRequiredParam<unsigned int>(
+      "mesh_id", "The id of the OpenMC mesh that this class should provide to Cardinal.");
+  params.addRangeCheckedParam<Real>(
+      "scaling",
+      1.0,
+      "scaling > 0.0",
+      "Scaling factor used to scale the OpenMC mesh dimensions for the creation of the mesh");
+  params.addParam<FileName>(
```
aprilnovak, line 42: Does this need to be in this class if it's already in the problem class/actions?

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
...
+## Description
+
+This mesh generator is used to create a mirror mesh from a regular mesh loaded with OpenMC at runtime.
+
+The OpenMC mesh used by `OpenMCMeshGenerator` is identified via the `mesh_id` parameter
+and must correspond to the valid ID number of a regular mesh in the OpenMC model. The path to the
+folder containing the OpenMC model can be declared using the `xml_directory` parameter.
+
+`OpenMCMeshGenerator` retrieves the coordinates of the lower left and upper right corners of the OpenMC
+regular mesh and scales them down according to a user-defined scaling factor (defaulting to 1)
```
aprilnovak:
```suggestion
regular mesh and divides them by a user-defined scaling factor (defaulting to 1)
```

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Example
+
+Here is an example showing how to set up an `OpenMCMeshGenerator` from a regular mesh declared with
+an ID of 1 in the OpenMC model:
+
+!listing test/tests/meshgenerators/openmc_mesh_generator/openmc_mesh_only.i
+
+In this example, a scaling factor of 10 is used. This means that the dimensions retrieved from OpenMC
+will be divided by 10 to construct the mesh on the Cardinal side.
```
aprilnovak:
```suggestion
will be divided by 10 to construct the mesh on the Cardinal side (since OpenMC always uses units of centimeters, this means that the Cardinal mesh would have units of decimeters).
```

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
...
+
```
aprilnovak, line 41: For the `scaling` parameter, I'm wondering if this needs to exist at all in this object? `OpenMCCellAverageProblem` also has a `scaling` parameter, which it uses to adjust the `[Mesh]` block (through a division). I think the current design in this PR would double-count this because the `[Mesh]` would be re-scaled into, e.g. meters, but then the `[Problem]` class would also do its own scaling, to units of hectometers?

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,98 @@
...
+               "A mesh with the id " + Moose::stringify(_mesh_id) +
+                   " does not exist in the OpenMC model! Please make sure the mesh has been "
+                   "added in the OpenMC model and you've supplied the correct mesh id.");
+
+  _mesh_index = openmc::model::mesh_map.at(_mesh_id);
+  _mesh = openmc::model::meshes[_mesh_index].get();
+
+  if (_mesh->n_dimension_ != 3)
+  {
+    mooseError("3D mesh only!");
```
aprilnovak: What is the reason for this limitation? Hypothetically, we could build 1D and 2D meshes, too. Is it something fundamental, or just not done at this stage? If it's not a fundamental barrier, I'd suggest we edit the error message to something like "At this time, this object only supports 3D meshes. Please contact the Cardinal developer team if you need 2D or 1D meshes."

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,98 @@
...
+  else
+  {
+    mooseError("Only implemented for openmc::RegularMesh!");
+  }
+}
+
+std::unique_ptr<MeshBase>
+OpenMCMeshGenerator::generate()
+{
+  return std::move(*_build_mesh);
```
aprilnovak, line 96: The way that most mesh generators are structured is that the mesh building step happens in this `generate` function. Is there a reason why we can't move the constructor code into here? For one thing, it would eliminate us needing to have a member variable `_build_mesh` and would increase similarity to other objects.

include/meshgenerators/OpenMCMeshGenerator.h
```diff
@@ -18,45 +18,34 @@
...
-  /// OpenMC cell IDs to which the translation will be applied
-  const std::set<int32_t> _cell_ids;
+  const unsigned int _mesh_id;
 
-private:
-  /// The MooseObject owning this transform
-  const MooseObject & _moose_object;
+  const Real & _scaling;
+
+  openmc::Mesh * _mesh = nullptr;
```
aprilnovak: `MeshGenerator` already has a member variable named `_mesh` - can you use a different name here? Also, if everything gets moved into `generate()` we can probably remove the two mesh data members here.

src/meshgenerators/OpenMCMeshGenerator.C
JoffreyDorville: I discovered that there is a restriction in MOOSE to call a mesh subgenerator (via `addMeshSubgenerator`) only during a specific action. If I move the mesh building step in the `OpenMCMeshGenerator` constructor instead, I end up with the following error: "Can only call addMeshSubgenerator() during MeshGenerator construction". I followed the example of this mesh generator to make it work:  `cardinal/contrib/moose/framework/src/meshgenerators/SideSetExtruderGenerator.C`.

include/meshgenerators/OpenMCMeshGenerator.h
JoffreyDorville: Done. I also modified `_mesh_index` to `_openmc_mesh_index` so that it is clearly identified.

src/meshgenerators/OpenMCMeshGenerator.C
JoffreyDorville: It is essentially because I am limiting my tests to 3D at the moment. I will edit the error message as suggested.

doc/content/source/auxkernels/ElementIDAux.md
JoffreyDorville: Done. I agree that this was unclear!

src/auxkernels/ElementIDAux.C
JoffreyDorville: Done. The cast was not needed here.

src/meshgenerators/OpenMCMeshGenerator.C
JoffreyDorville: I was thinking that both `scaling` and `xml_directory` were good to keep in both objects (`OpenMCCellAverageProblem` and `OpenMCMeshGenerator`) as they can be used independently in principle. I am not sure if running an OpenMCMeshGenerator alone would be that useful, but with the current state of the code, it is possible. Please let me know what you think!

doc/content/source/auxkernels/ElementIDAux.md
JoffreyDorville: I agree that this object is generic enough to be proposed as a PR to MOOSE directly. Do you know what would be the best way to continue with this PR if I were to propose ElementIDAux in a separate MOOSE PR? Should we keep ElementIDAux in this PR and wait until we have access to it through MOOSE before removing it (maybe via a followup PR)?

src/meshgenerators/OpenMCMeshGenerator.C
JoffreyDorville: Done.

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
JoffreyDorville: I will need additional testing to make sure I understand it correctly. I'll answer back when I have more elements!

doc/content/source/auxkernels/ElementIDAux.md
aprilnovak: I propose that a PR be opened into MOOSE now - it'll take me a bit more time to review your PR here in full, so likely by the time it's in MOOSE it'll be time for your PR to get merged. They're usually quick to review.

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
aprilnovak: A quick first check is to look at the mesh generated without the `--mesh-only` flag (i.e. as the result of a simulation), when `scaling` /= 1 is used. Does `scaling` get applied 2x?

doc/content/source/auxkernels/ElementIDAux.md
JoffreyDorville: I just created the MOOSE PR for the new auxkernel: https://github.com/idaholab/moose/pull/33250

src/meshgenerators/OpenMCMeshGenerator.C
aprilnovak: Hmm, that's a good point... when the [Mesh] is not in units of centimeters, the way we currently approach this is that the scaling is handled entirely inside `OpenMCCellAverageProblem`. I think the same design should be undertaken here, by removing `scaling` from the mesh object itself. 

You can confirm that, if you had scaling in both places, that OpenMC is going to try to write data to a [Mesh] which has its coordinates divided by `scaling * scaling` instead of just `scaling`. You could devise a scenario to prove this by making an OpenMC model which is a cube of size 10 cm x 10 cm x 10 cm with a different material outside (extending to infinity). The cell IDs that you'll see on the Cardinal side will be different if the [Mesh] coordinates were divided by `scaling` vs. `scaling * scaling`

doc/content/source/auxkernels/ElementIDAux.md
aprilnovak: Thanks for doing this! Now that it's been merged, you can put a submodule update for MOOSE in this PR, then remove the duplicates files added here.

JoffreyDorville: Done.

test/tests/meshgenerators/openmc_mesh_generator/openmc_model.py
```diff
@@ -0,0 +1,68 @@
...
+tally_2.filters = [mesh_filter_2]
+tally_2.scores = ["flux"]
+
+# Register tally
+tallies = openmc.Tallies([tally_1, tally_2])
+
+# Register settings
+settings = openmc.Settings()
+settings.batches = 2
+settings.particles = 50
```
aprilnovak: Can you increase the number of particles? I think this is why some of your tests are failing when running in parallel - 50 is quite low, and when running with MPI there could be coincidentally zero fission sites banked on a particular rank

JoffreyDorville: Thanks for your comment! I was thinking the same when I saw the reasons why these tests were failing, but did not have the chance to modify them yet. I'll change that now!

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
JoffreyDorville: I added 3 test cases verifying that the scaling factors of `OpenMCMeshGenerator` and `OpenMCCellAverageProblem` work together consistently. In `OpenMCMeshGenerator`, the scaling factor is used to divide the coordinates retrieved from OpenMC to create a MOOSE mesh. From what I understand with the scaling factor from `OpenMCCellAverageProblem`, Cardinal applies the scaling factor (by multiplying) to the MOOSE mesh coordinates to evaluate the OpenMC cell that corresponds to the center of MOOSE mesh cells during mapping.

src/meshgenerators/OpenMCMeshGenerator.C
JoffreyDorville: Just quoting my previous comment as it also answers this comment.

> I added 3 test cases verifying that the scaling factors of OpenMCMeshGenerator and OpenMCCellAverageProblem work together consistently. In OpenMCMeshGenerator, the scaling factor is used to divide the coordinates retrieved from OpenMC to create a MOOSE mesh. From what I understand with the scaling factor from OpenMCCellAverageProblem, Cardinal applies the scaling factor (by multiplying) to the MOOSE mesh coordinates to evaluate the OpenMC cell that corresponds to the center of MOOSE mesh cells during mapping.

So in practice, a user would need to define a scaling factor of 100 for both objects if we want to have the MOOSE mesh in meter (knowing that OpenMC uses cm).

test/tests/meshgenerators/openmc_mesh_generator/openmc_model.py
JoffreyDorville: Done

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+
+  // Search for OpenMCMeshGenerators and verify that xml_directory parameters are consistent
+  bool found = false;
+  for (const auto * action : mesh_gen_actions)
+  {
+    if (action->getMooseObjectType() == "OpenMCMeshGenerator")
+    {
+      std::string xml_directory_temp = action->getObjectParams().get<FileName>("xml_directory");
+      if (found)
+      {
```
nuclearkevin, line 116:
```suggestion
```

By the MOOSE style guide, single-line if statements don't use braces.

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+        if (xml_directory_temp != xml_directory)
+        {
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+          mooseError(
+              "Inconsistent xml directories for OpenMC in the declared OpenMCMeshGenerators");
+        }
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+      }
```
nuclearkevin, line 128:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+      else
+      {
+        xml_directory = xml_directory_temp;
+        found = true;
+      }
+    }
+  }
+  if (found)
+  {
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -66,6 +86,41 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+    return true;
+  }
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,13 +38,32 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
 void
 OpenMCInitAction::act()
```
nuclearkevin, line 39: Thinking about the scaling parameter discussion above/below. What are your thoughts on checking (in `OpenMCInitAction::act()`) to make sure both the `OpenMCCellAverageProblem` and the `OpenMCMeshGenerator` use the same scaling parameter (similarly to `xml_dir`)? I don't think there's a case where you'd want different scaling parameters when using the classes in concert, and specifying two scaling parameters can be error-prone.

src/actions/OpenMCInitAction.C
```diff
@@ -37,13 +38,32 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+  std::string xml_directory_generator;
+  bool openmc_mesh_generator_requested = isOpenMCMeshGeneratorRequested(xml_directory_generator);
+
+  // if there is no need to initialize OpenMC, return
+  if (!openmc_problem_requested && !openmc_mesh_generator_requested)
     return;
 
+  // If both OpenMC problem and mesh generator are present, check xml_directory consistency
+  if (openmc_problem_requested && openmc_mesh_generator_requested)
+  {
```
nuclearkevin, line 59:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,13 +38,32 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+    if (xml_directory_problem != xml_directory_generator)
+    {
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,13 +38,32 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+      mooseError("Inconsistent xml directories for OpenMC in the mesh generator and the problem "
+                 "declarations.");
+    }
```
nuclearkevin:
```suggestion
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,13 +38,32 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+  }
```
nuclearkevin, line 71:
```suggestion
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,99 @@
...
+    paramError("mesh_id",
+               "A mesh with the id " + Moose::stringify(_mesh_id) +
+                   " does not exist in the OpenMC model! Please make sure the mesh has been "
+                   "added in the OpenMC model and you've supplied the correct mesh id.");
+
+  _openmc_mesh_index = openmc::model::mesh_map.at(_mesh_id);
+  _openmc_mesh = openmc::model::meshes[_openmc_mesh_index].get();
+
+  if (_openmc_mesh->n_dimension_ != 3)
+  {
```
nuclearkevin:
```suggestion
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,99 @@
...
+    mooseError("At this time, this object only supports 3D meshes. Please contact the Cardinal "
+               "developer team if you need 2D or 1D meshes!");
+  }
```
nuclearkevin:
```suggestion
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,99 @@
...
+    params.set<Real>("ymin") = regular_mesh->lower_left()[1] / _scaling;
+    params.set<Real>("ymax") = regular_mesh->upper_right()[1] / _scaling;
+    params.set<Real>("zmin") = regular_mesh->lower_left()[2] / _scaling;
+    params.set<Real>("zmax") = regular_mesh->upper_right()[2] / _scaling;
+    addMeshSubgenerator("GeneratedMeshGenerator", name() + "_openmc_sub_mesh", params);
+
+    _build_mesh = &getMeshByName(name() + "_openmc_sub_mesh");
+  }
+  else
+  {
```
nuclearkevin:
```suggestion
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,99 @@
...
+    mooseError("Only implemented for openmc::RegularMesh!");
+  }
```
nuclearkevin:
```suggestion
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,85 @@
+[Tests]
+  issues = "#1369"
+  design = "OpenMCMeshGenerator.md ElementIDAux.md"
```
nuclearkevin:
```suggestion
  design = "OpenMCMeshGenerator.md"
```
Since this is no longer in Cardinal, we don't need to include it :)

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,99 @@
...
+    params.set<Real>("ymax") = regular_mesh->upper_right()[1] / _scaling;
+    params.set<Real>("zmin") = regular_mesh->lower_left()[2] / _scaling;
+    params.set<Real>("zmax") = regular_mesh->upper_right()[2] / _scaling;
+    addMeshSubgenerator("GeneratedMeshGenerator", name() + "_openmc_sub_mesh", params);
+
+    _build_mesh = &getMeshByName(name() + "_openmc_sub_mesh");
+  }
+  else
+  {
+    mooseError("Only implemented for openmc::RegularMesh!");
```
nuclearkevin:
```suggestion
    mooseError("Only implemented for regular meshes in OpenMC!");
```

Users may not understand C++ namespaces. IMO, this is more clear.

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,85 @@
...
+                 "OpenMCMeshGenerators"
+    requirement = "OpenMCMeshGenerator must fail if there are two different xml_directory"
+                  "values"
+    capabilities = "openmc"
+  []
+  [wrong_mesh_type]
+    type = RunException
+    input = openmc_wrong_mesh_type.i
+    cli_args = "--mesh-only"
+    expect_err = "Only implemented for openmc::RegularMesh!"
```
nuclearkevin:
```suggestion
    expect_err = "Only implemented for regular meshes in OpenMC!"
```

nuclearkevin (changes requested):
Thanks @JoffreyDorville! I'm mostly happy where this is at design-wise aside from one comment (checking scaling between `OpenMCCellAverageProblem` and `OpenMCMeshGenerator`). The vast majority of these are me being the style police to make sure we conform to MOOSE's C++ guidelines.

src/actions/OpenMCInitAction.C
JoffreyDorville: I really like this idea! I would be interested to know what are your thoughts on this idea @aprilnovak ?

src/actions/OpenMCInitAction.C
JoffreyDorville: In this specific case, will it not create a "dangling else" problem with the following else statement?

nuclearkevin: Ah, you're very correct. I didn't notice that when reviewing the changes. I'll resolve the change comments.

src/actions/OpenMCInitAction.C
JoffreyDorville: Just added the consistency check. Thanks again for the idea @nuclearkevin !

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,86 @@
...
+    capabilities = "openmc"
+  []
+  [inconsistent_scaling_1]
+    type = RunException
+    input = openmc_mesh_scaling.i
+    cli_args = "Problem/scaling=1.0"
+    expect_err = "Inconsistent scaling factors for OpenMC in the mesh generator and problem "
+                 "declarations."
+    requirement = "Inconsistent scaling factors between OpenMCMeshGenerator (S1) and "
+                  "OpenMCCellAverageProblem (S2) must fail when S2 < S1"
```
nuclearkevin:
```suggestion
                  "OpenMCCellAverageProblem (S2) must fail when S2 is not equal to S1"
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,86 @@
...
+    capabilities = "openmc"
+  []
+  [inconsistent_scaling_2]
```
nuclearkevin: We can remove this test as it effectively tests the same requirement as `inconsistent_scaling_1` (scaling factors must be the same).

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,86 @@
...
+  []
+  [consistent_scaling]
+    type = CSVDiff
+    input = openmc_mesh_scaling.i
+    csvdiff = openmc_mesh_scaling_out.csv
+    requirement = "Consistent scaling factors between OpenMCMeshGenerator and "
+                  "OpenMCCellAverageProblem must work"
+    capabilities = "openmc"
+  []
+  [inconsistent_scaling_1]
```
nuclearkevin:
```suggestion
  [inconsistent_scaling]
```

nuclearkevin (changes requested):
IMO, this is almost ready to go! My only comment is that one of the tests is now a duplicate of another test due to the recent scaling check changes.

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done.

nuclearkevin (approved):
Thanks @JoffreyDorville, this looks good to me now! If @aprilnovak is happy with it, I'll merge tomorrow.

aprilnovak:
I'm out of the office today, but I'd like to take a look on Monday if we could wait to merge until then? Just want to be sure I understand how the `scaling` works now

doc/content/source/meshgenerators/OpenMCMeshGenerator.md
```diff
@@ -0,0 +1,44 @@
+# OpenMCMeshGenerator
+
+!syntax description /Mesh/OpenMCMeshGenerator
+
+## Description
+
+This mesh generator is used to create a mirror mesh from a regular mesh loaded with OpenMC at runtime.
```
aprilnovak:
```suggestion
This mesh generator is used to create a mirror mesh from a [regular mesh in OpenMC](https://docs.openmc.org/en/stable/pythonapi/generated/openmc.RegularMesh.html) loaded at runtime.
```

include/meshgenerators/OpenMCMeshGenerator.h
```diff
@@ -0,0 +1,51 @@
...
+
+  OpenMCMeshGenerator(const InputParameters & parameters);
+
+  std::unique_ptr<MeshBase> generate() override;
+
+private:
+  /// Generated mesh
+  std::unique_ptr<MeshBase> * _build_mesh;
+
+  const unsigned int _mesh_id;
```
aprilnovak: Do you need this to be a member variable, or would it be sufficient to just get a local copy of it in the constructor that then goes out of scope when no longer needed? For clarity, I try to avoid creating member variables unless they are used across multiple functions/accessed via other classes

include/meshgenerators/OpenMCMeshGenerator.h
```diff
@@ -0,0 +1,51 @@
...
+
+  const Real & _scaling;
```
aprilnovak, line 42:
```suggestion
  /// Scaling factor to divide the mesh coordinates in the RegularMesh in OpenMC to get into units used by Cardinal
  const Real & _scaling;
```

include/meshgenerators/OpenMCMeshGenerator.h
```diff
@@ -0,0 +1,51 @@
...
+
+  openmc::Mesh * _openmc_mesh = nullptr;
```
aprilnovak, line 45:
```suggestion
  /// pointer to underlying OpenMC mesh
  openmc::Mesh * _openmc_mesh = nullptr;
```

include/meshgenerators/OpenMCMeshGenerator.h
```diff
@@ -0,0 +1,51 @@
...
+
+  unsigned int _openmc_mesh_index;
```
aprilnovak, line 48:
```suggestion
  /// index of the OpenMC mesh in OpenMC's internal arrays
  unsigned int _openmc_mesh_index;
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,19 +38,47 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+  // if there is no need to initialize OpenMC, return
+  if (!openmc_problem_requested && !openmc_mesh_generator_requested)
     return;
 
+  // If both OpenMC problem and mesh generator are present
+  if (openmc_problem_requested && openmc_mesh_generator_requested)
+  {
+    // Check for xml_directory consistency
+    if (xml_directory_problem != xml_directory_generator)
+      mooseError("Inconsistent xml directories for OpenMC in the mesh generator and problem "
```
aprilnovak:
```suggestion
      mooseError("Inconsistent 'xml_directory' for OpenMC in the mesh generator and problem "
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,19 +38,47 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+                 "declarations.");
```
aprilnovak:
```suggestion
                 "declarations. Make sure the same value is used throughout the [Mesh] and [Problem] blocks.");
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,19 +38,47 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+
+    // Check for scaling consistency
+    if (scaling_problem != scaling_generator)
+      mooseError("Inconsistent scaling factors for OpenMC in the mesh generator and problem "
```
aprilnovak:
```suggestion
      mooseError("Inconsistent 'scaling' factors for OpenMC in the mesh generator and problem "
```

src/actions/OpenMCInitAction.C
```diff
@@ -37,19 +38,47 @@ OpenMCInitAction::OpenMCInitAction(const InputParameters & parameters) : Action(
...
+                 "declarations.");
```
aprilnovak:
```suggestion
                 "declarations. Make sure the same value is used throughout the [Mesh] and [Problem] blocks.");
```

src/actions/OpenMCInitAction.C
```diff
@@ -60,12 +89,52 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+    if (action->getMooseObjectType() == "OpenMCMeshGenerator")
+    {
+      std::string xml_directory_temp = action->getObjectParams().get<FileName>("xml_directory");
+      Real scaling_temp = action->getObjectParams().get<Real>("scaling");
+      if (found)
+      {
+        // Check for xml_directory consistency
+        if (xml_directory_temp != xml_directory)
+          mooseError(
+              "Inconsistent xml directories for OpenMC in the declared OpenMCMeshGenerators");
```
aprilnovak:
```suggestion
              "Inconsistent 'xml_directory' values found in the declared OpenMCMeshGenerators; the same choice for 'xml_directory' must be used across all OpenMCMeshGenerator objects");
```

src/actions/OpenMCInitAction.C
```diff
@@ -60,12 +89,52 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+
+        // Check for scaling consistency
+        if (scaling_temp != scaling)
+          mooseError(
+              "Inconsistent scaling factors for OpenMC in the declared OpenMCMeshGenerators");
```
aprilnovak:
```suggestion
              "Inconsistent 'scaling' factors for OpenMC in the declared OpenMCMeshGenerators; the same choice for 'scaling' must be used across all OpenMCMeshGenerator objects");
```

src/actions/OpenMCInitAction.C
```diff
@@ -60,12 +89,52 @@ OpenMCInitAction::isOpenMCCellAverageProblemRequested(std::string & xml_director
...
+      }
+      else
+      {
+        xml_directory = xml_directory_temp;
+        scaling = scaling_temp;
+        found = true;
+      }
+    }
+  }
+  if (found)
```
aprilnovak: You can simplify these three lines (133 - 135) and just do

```
return found;
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,95 @@
...
+
+  params.addClassDescription(
+      "A class that generates a mesh from an available OpenMC regular mesh.");
+  params.addRequiredParam<unsigned int>(
+      "mesh_id", "The id of the OpenMC mesh that this class should provide to Cardinal.");
+  params.addRangeCheckedParam<Real>(
+      "scaling",
+      1.0,
+      "scaling > 0.0",
+      "Scaling factor used to scale the OpenMC mesh dimensions for the creation of the mesh");
```
aprilnovak:
```suggestion
      "Scaling factor used to scale the OpenMC mesh dimensions for the creation of the mesh; OpenMC always uses units of centimeters. Setting 'scaling = 100' means that the [Mesh] created will have units of meters, for example.");
```

src/meshgenerators/OpenMCMeshGenerator.C
```diff
@@ -0,0 +1,95 @@
...
+    params.set<Real>("ymin") = regular_mesh->lower_left()[1] / _scaling;
+    params.set<Real>("ymax") = regular_mesh->upper_right()[1] / _scaling;
+    params.set<Real>("zmin") = regular_mesh->lower_left()[2] / _scaling;
+    params.set<Real>("zmax") = regular_mesh->upper_right()[2] / _scaling;
+    addMeshSubgenerator("GeneratedMeshGenerator", name() + "_openmc_sub_mesh", params);
+
+    _build_mesh = &getMeshByName(name() + "_openmc_sub_mesh");
+  }
+  else
+    mooseError("Only implemented for regular meshes in OpenMC!");
```
aprilnovak:
```suggestion
    mooseError("This object is currently only implemented for regular meshes (RegularMesh) in OpenMC! Please contact the Cardinal developer team if you require another structured mesh type.");
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+    requirement = "OpenMCMeshGenerator must fail if the requested mesh type is not an "
+                  "OpenMC RegularMesh"
+    capabilities = "openmc"
+  []
+  [two_meshes]
+    type = Exodiff
+    input = openmc_two_meshes.i
+    exodiff = openmc_two_meshes_in.e
+    cli_args = "--mesh-only"
+    requirement = "OpenMCMeshGenerator must accept at least two different calls from the "
```
aprilnovak:
```suggestion
    requirement = "OpenMCMeshGenerator will accept more than one mesh generator from the "
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+  [verify_point_sampling]
+    type = CSVDiff
+    input = openmc_mesh_sampling.i
+    csvdiff = openmc_mesh_sampling_out_sample_points_0001.csv
+    requirement = "Points generated inside the mesh produced by OpenMCMeshGenerator "
+                  "must correctly map to the mesh element IDs."
+    mesh_mode = "replicated"
+    capabilities = "openmc"
+  []
+  [verify_bounding_box_and_info]
```
aprilnovak: what is the difference between the `verify_bounding_box_and_info` test vs. the `verify_point_sampling` test? There's no difference in the `cli_args`, so these two tests are running the same thing? The better way to do this is to have separate CSV files listed (you can check multiple), like this

```
csvdiff = 'openmc_mesh_sampling_out_sample_points_0001.csv openmc_mesh_sampling_out.csv'
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
+[Tests]
+  issues = "#1369"
+  design = "OpenMCMeshGenerator.md"
+
```
aprilnovak, line 4: Some tests I think are missing:

- Add check for an invalid mesh ID (non-existent one) in OpenMC
- Only supports 3D meshes in OpenMC
- Check that an error is thrown if different `xml_directory` are used between the Problem vs. Mesh (I see the check for consistency in the meshgenerators themselves)
- Check that an error is thrown if different `scaling` is used across multiple mesh generators (I see the check for consistency between the Problem and Mesh, but didn't see it across the different meshes)

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+  [mesh_only]
+    type = Exodiff
+    input = openmc_mesh_only.i
```
aprilnovak: Since the `[Mesh]` block in the `openmc_mesh_only.i` file is identical to the one in the `openmc_mesh_sampling.i` file, we can delete `openmc_mesh_only.i` - when you use the `--mesh-only` flag, it ignores the entire input file except for the `[Mesh]` block anyways.

aprilnovak:
```suggestion
    input = openmc_mesh_scaling.i
```

aprilnovak: To do so, you'll just need to rename the `openmc_mesh_only_in.e` that you have in the `gold` folder to `openmc_mesh_scaling_in.e`

test/tests/meshgenerators/openmc_mesh_generator/openmc_mesh_only.i
```diff
@@ -0,0 +1,8 @@
+[Mesh]
```
aprilnovak: I think this file can be deleted, see comment in `tests`

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+    exodiff = openmc_mesh_only_in.e
+    cli_args = "--mesh-only"
+    requirement = "OpenMCMeshGenerator must produce a valid mesh from an OpenMC model"
```
aprilnovak:
```suggestion
    requirement = "The system shall be able to generate a structured mesh in MOOSE from an OpenMC RegularMesh and with scaling; this is checked by comparing node coordinates and element count against the internal OpenMC mesh"
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+    cli_args = "--mesh-only"
+    expect_err = "Inconsistent xml directories for OpenMC in the declared "
+                 "OpenMCMeshGenerators"
+    requirement = "OpenMCMeshGenerator must fail if there are two different xml_directory"
+                  "values"
+    capabilities = "openmc"
+  []
+  [wrong_mesh_type]
+    type = RunException
+    input = openmc_wrong_mesh_type.i
```
aprilnovak: We can delete this input file and instead accomplish the same thing with `cli_args` - this helps long-term for code maintenance.

aprilnovak:
```suggestion
    input = openmc_mesh_scaling.i
```

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+    cli_args = "--mesh-only"
```
aprilnovak, line 9:
```suggestion
    cli_args = "Mesh/openmc_mesh/mesh_id=2 --mesh-only"
```

test/tests/meshgenerators/openmc_mesh_generator/openmc_wrong_mesh_type.i
```diff
@@ -0,0 +1,7 @@
+[Mesh]
```
aprilnovak: This file can be deleted, see comment in `tests`

test/tests/meshgenerators/openmc_mesh_generator/openmc_two_meshes.i
```diff
@@ -0,0 +1,13 @@
...
+    type = OpenMCMeshGenerator
+    mesh_id = 1
+    scaling = 10
+  []
+  [openmc_mesh_2]
+    type = OpenMCMeshGenerator
+    mesh_id = 1
+    scaling = 10
+  []
+  final_generator = openmc_mesh_2
```
aprilnovak: I think this test would be more interesting/useful if instead of using the `final_generator` (which is effectively going to ignore the presence of the first mesh generator) we instead used a `CombinerGenerator` to combine the two created meshes and create a united mesh from them. I could see this being an actual use case, e.g. if someone has one mesh in fuel assemble A and a different one in fuel assembly B that they want to both be represented.

aprilnovak:
```suggestion
  [combine]
    type = CombinerGenerator
    inputs = 'openmc_mesh openmc_mesh_2'
  []
```

Except, I would do this with two actual unique meshes in OpenMC, not the same object.

test/tests/meshgenerators/openmc_mesh_generator/openmc_two_meshes.i
```diff
@@ -0,0 +1,13 @@
...
+[]
```
aprilnovak, line 16: This may be a comment for a future PR/implementation stage - but what do you want to happen if the user manipulates the mesh loaded in `OpenMCMeshGenerator`? For instance, are you relying on that mesh not being manipulated by the user in other mesh generators, for the sake of how data is going to be copied to/from that mesh? 

Would it be an error if the user did this, for instance - and translated a mesh to a new position in space? If this is not the desired behavior, we should add a check that no other mesh generator can consume the `OpenMCMeshGenerator.

```
[Mesh]
  [o]
    type = OpenMCMeshGenerator
    // ...
  []
  [translate]
    type = TransformGenerator
    input = o
    transform_type = 'translate'
    vector_value = '100 50 0'
  []
```

This would translate the mesh coordinates on the MOOSE side so that each original coordinate (x, y, z) becomes (x + 100, y + 50, z + 0).

test/tests/meshgenerators/openmc_mesh_generator/tests
```diff
@@ -0,0 +1,76 @@
...
+    type = Exodiff
+    input = openmc_mesh_only.i
+    exodiff = openmc_mesh_only_in.e
+    cli_args = "--mesh-only"
+    requirement = "OpenMCMeshGenerator must produce a valid mesh from an OpenMC model"
+    capabilities = "openmc"
+  []
+  [different_xml_directories]
+    type = RunException
+    input = openmc_different_xml_directories.i
```
aprilnovak: You can delete this file and accomplish the same thing with `cli_args` using the `openmc_two_meshes` file. Another reason to consolidate like this (in addition to code maintenance reasons) is that we won't have a bunch of erroneous input files showing up on the documentation pages (at the bottom of each code object page there's a list of all the input files that object shows up in -- it can be confusing to a user if some of those input files are actually for the purpose of `RunException` tests, since they might mistake those input files as examples-to-follow rather than input-files-that-only-exist-for-testing-errors).

test/tests/meshgenerators/openmc_mesh_generator/openmc_different_xml_directories.i
```diff
@@ -0,0 +1,14 @@
+[Mesh]
```
aprilnovak: File can be deleted, see comment in `tests`

aprilnovak:
Thanks @JoffreyDorville for accommodating all the requested changes! This was my first look at the tests - these look great, just added some easy-to-implement requests to consolidate the number of input files and address some extra tests I think we can add.

src/actions/OpenMCInitAction.C
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/openmc_different_xml_directories.i
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/openmc_wrong_mesh_type.i
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done

JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/openmc_mesh_only.i
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done

JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/openmc_two_meshes.i
JoffreyDorville: Done. I agree this is a better illustration of what it can be used for!

JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done

test/tests/meshgenerators/openmc_mesh_generator/tests
JoffreyDorville: Done. I merged the two tests together. Maybe something to note here, because we are using `ElementIDAux` for sampling, we need to use the `replicated` `mesh_mode` when running on distributed meshes. Do you think we should improve how we test the IDs for the distributed mode?

include/meshgenerators/OpenMCMeshGenerator.h
JoffreyDorville: Done. I agree this was not necessary and somewhat redundant with `_openmc_mesh_index`.

test/tests/meshgenerators/openmc_mesh_generator/openmc_two_meshes.i
JoffreyDorville: This is a really great question. In the temperature field transfer feature I am working on, I assume  that the mapping between MOOSE mesh cells and openmc temperature field cells is naive (same cell ordering) and non-modifiable (no update based on geometric consideration for example) for the `OpenMCMeshGenerator`. Any transformation of the mesh will break this assumption.

If we were to allow mesh transformation, we would need to have a dynamic mapping identification process, just like the one for MOOSE mesh cell to OpenMC cells. If we do not want to go that route, we would need to prevent `OpenMCMeshGenerator` from being consumed by other generators as you mentioned, except combiner generators maybe. I also wonder if the combiner generator case would not also mandate the use of a dynamic mapping identification process as well.

To summarize, I think we can keep the current implementation of  `OpenMCMeshGenerator` as it is just one of many ways of creating a MOOSE mesh for the moment and because the mapping to OpenMC cells is dynamic. But I will need to take this point into account for the next PR to correctly transfer data to the openmc mesh.

I am leaning toward dynamic mapping identification, even if that is another mapping to add, it would unlock the most complete set of features in the end. What are your thoughts on this point?

JoffreyDorville:
Thanks for the review @aprilnovak ! I updated the code to address your comments and replied to all conversations. Please let me know what you think!

test/tests/meshgenerators/openmc_mesh_generator/openmc_two_meshes.i
aprilnovak: Agree, whatever path is taken I think this can be a future PR because there's nothing technically wrong/uncovered by this PR because it's just building the mesh.

I think the CombinerGenerator would also need some kind of mapping, I suppose - the elements will be ordered according to the order of the meshgenerators passed into the `inputs` parameter. So, you'd need to know what that order is for the data transfer to work properly. 

What do you envision that dynamic mapping to be? I imagine it'd be different from what OpenMCCellAverageProblem does to identify the cell <--> element mapping? (because if it were the same thing, then what would the advantage of the `OpenMCMeshGenerator` be? One could just make a `GeneratedMeshGenerator` manually)

aprilnovak:
LGTM!

PR #1366, opened 2026-06-22
force tolerance in ValueDifferenceHeuristicUserObject to be positive

magnoxemo:
This PR forces `tolerance` param in `ValueDifferenceHeuristicUserObject` to be positive always

### DIFF
--- a/src/userobjects/ValueDifferenceHeuristicUserObject.C
+++ b/src/userobjects/ValueDifferenceHeuristicUserObject.C
@@ -7,8 +7,9 @@ ValueDifferenceHeuristicUserObject::validParams()
 {
 
   InputParameters params = ClusteringUserObjectBase::validParams();
-  params.addRequiredParam<Real>(
+  params.addRangeCheckedParam<Real>(
       "tolerance",
+      "tolerance > 0"
       "Maximum allowed absolute relative difference in values for elements to be clustered");
   params.addClassDescription(
       "Clusters elements whose relative value differences are less than the specified tolerance.");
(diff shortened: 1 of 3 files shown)


src/userobjects/ValueDifferenceHeuristicUserObject.C
```diff
@@ -19,6 +20,8 @@ ValueDifferenceHeuristicUserObject::ValueDifferenceHeuristicUserObject(
     const InputParameters & params)
   : ClusteringUserObjectBase(params), _tolerance(getParam<Real>("tolerance"))
 {
+  if (_tolerance < 0.0)
```
aprilnovak: The recommended way to do this is to instead use `addRequiredRangeCheckedParam`, so that MOOSE's built-in systems handle the error checkin - if you grep for this phrase, you'll find examples

aprilnovak: then, we don't add RunException tests for it, since we rely on MOOSE's test suite itself to cover the correct functionality

src/userobjects/ValueDifferenceHeuristicUserObject.C
```diff
@@ -7,8 +7,10 @@ ValueDifferenceHeuristicUserObject::validParams()
 {
 
   InputParameters params = ClusteringUserObjectBase::validParams();
-  params.addRequiredParam<Real>(
-      "tolerance", "Maximum allowed difference in values for elements to be clustered");
+  params.addRangeCheckedParam<Real>(
```
nuclearkevin:
```suggestion
  params.addRequiredRangeCheckedParam<Real>(
```

nuclearkevin (approved):
LGTM. The failure is unrelated and will be addressed in https://github.com/neams-th-coe/cardinal/pull/1358.

PR #1376, opened 2026-07-01
fix qprime to W/m. Refs #1375

### LINKED ISSUE #1375: Energy conservation in the subchannel tutorial
## Reason
The `openmc_subchannel` tutorial was accidentally setting `q_prime` (linear heat rate) to be used in the subchannel solve with units of W, when it should be W/m.

## Design
Fix the tutorial for physical consistency.

## Impact
Improved tutorial.

aprilnovak:
Closes #1375


nuclearkevin (changes requested):
The subchannel tutorial changes look good to me!

I like the performance warning when building DAGMC without Double-Down quite a bit. We should also add a note to `options.md` above the `export ENABLE_DOUBLE_DOWN=yes` line to let people know that Double-Down results in a large improvement in performance as well (its fairly easy to miss the warnings when building as the terminal gets polluted with a *lot* of output).

nuclearkevin (approved):
LGTM. Failure is unrelated.

PR #1377, opened 2026-07-01
Unset force vendored libs for latest Mac. 

aprilnovak:
On the latest Mac, for some reason the forcing of vendored libs in OpenMC is clashing with `fmt` elsewhere when using MOOSE's conda environment. At least 4 users have seen this and this change fixes it.


aprilnovak:
The test failures are unrelated.

nuclearkevin (approved):
LGTM. Its rather odd that this is a problem now. Could be related to recent changes to how OpenMC handles CMake dependencies: https://github.com/openmc-dev/openmc/pull/3653/changes

Edit: forgot to mention that the failure is unrelated.

Issue #1373, opened 2026-06-30
Documentation Improvements


aprilnovak:
## Reason
This is a catch-all issue for documentation improvements, including: (i) new tutorials, (ii) adding publications to the website, (iii) documentation fixes, and more.

## Design
Implement needed changes to documentation markdown files.

## Impact
Improved user experience.

PR #1379, opened 2026-07-06
add more description to check_tally_sum error. Refs #1373

aprilnovak:
Improving error message for the global tally check, inspired by a user who ran into this error and needed help with debugging

### DIFF
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -83,6 +83,12 @@ public:
   bool isHeatingScore(const std::string & score) const;
 
   /**
+   * The value used to normalize the tallies
+   * @return the source strength or power, depending on the run mode in OpenMC
+   */
+  const Real tallyNormalizationValue() const;
+
+  /**
    * Whether the random ray solver can accumulate a score or not.
    * @return whether the random ray solver can accumulate the score or not
    */
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -86,7 +86,7 @@ public:
    * The value used to normalize the tallies
    * @return the source strength or power, depending on the run mode in OpenMC
    */
-  const Real tallyNormalizationValue() const;
+  Real tallyNormalizationValue() const;
 
   /**
    * Whether the random ray solver can accumulate a score or not.
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -1079,4 +1079,10 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
     uo->setValue();
 }
 
+const Real
+OpenMCProblemBase::tallyNormalizationValue() const
+{
+  return _run_mode == openmc::RunMode::FIXED_SOURCE ? *_source_strength : *_power;
+}
+
 #endif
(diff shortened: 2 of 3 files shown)


src/tallies/TallyBase.C
```diff
@@ -783,16 +783,41 @@ TallyBase::checkTallySum(const unsigned int & score) const
...
+        << "' parameter you provided for normalizing the tallies will therefore correspond only to "
+           "those hits which Cardinal is seeing. If this is your intention, you can set "
+           "'normalize_by_global_tally' to false, which means that Cardinal will be normalizing "
+           "your tallies so that they correspond to a ";
+    std::replace(run_type.begin(), run_type.end(), '_', ' ');
+
+    msg << run_type << " of "
+        << Moose::stringify(_local_sum_tally[score] / _global_sum_tally[score] *
+                            _openmc_problem.tallyNormalizationValue())
+        << " (NOT the user-specified power of " << _openmc_problem.tallyNormalizationValue()
```
nuclearkevin:
```suggestion
        << " (NOT the user-specified power / source rate of " << _openmc_problem.tallyNormalizationValue()
```

src/tallies/TallyBase.C
```diff
@@ -783,16 +783,41 @@ TallyBase::checkTallySum(const unsigned int & score) const
...
+        << "). Or, if this relative difference is acceptable to you, you can loosen "
+           "'tally_sum_tol'.\n\n";
+
+    msg << "If this error is NOT expected, you need to debug your model to determine where some "
+           "tally hits are not being mapped to Cardinal. If you are using a cell tally, this "
+           "usually happens because some OpenMC cells are not hit by an element centroid in the "
+           "[Mesh]. Another possibility is that your OpenMC model is bigger than the [Mesh] and "
+           "some tally hits are happening outside the [Mesh]. To debug, you can try:\n\n"
```
nuclearkevin:
```suggestion
           "some tally hits are happening outside the [Mesh]. To debug, you can try to:\n\n"
```

include/base/OpenMCProblemBase.h
```diff
@@ -82,6 +82,12 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
    */
   bool isHeatingScore(const std::string & score) const;
 
+  /**
+   * The value used to normalize the tallies
+   * @return the source strength or power, depending on the run mode in OpenMC
+   */
+  const Real tallyNormalizationValue() const;
```
nuclearkevin: First `const` isn't necessary as the value of a simple primitive is being returned (unless you wanted to return a ref instead).
```suggestion
  Real tallyNormalizationValue() const;
```

src/base/OpenMCProblemBase.C
```diff
@@ -1079,4 +1079,10 @@ OpenMCProblemBase::sendNuclideDensitiesToOpenMC()
     uo->setValue();
 }
 
+const Real
```
nuclearkevin:
```suggestion
Real
```

nuclearkevin (changes requested):

src/tallies/TallyBase.C
aprilnovak: Good catch! I addressed this a different way, since I didn't want the user to be confused whether `power / source rate` is a ratio vs. a "or".

nuclearkevin (approved):
LGTM. Failure is unrelated.

Issue #1378, opened 2026-07-03
Allow distribcell instance lists to be specified from input file


lewisgross1296:
## Reason

I'm working with a TRISO depletion model that I'm currently trying to couple to Cardinal. One general roadblock we've encountered is that there are so many TRISO cell instances in the model that the XML that gets written is actually unreadable for most programs that read XML. See openmc-dev/openmc#3744 and openmc-dev/openmc#3838.

In order to do coupled multiphysics + depletion, I plan to do a multiphysics solve with updated `materials.xml` after each depletion run. The `materials.xml` is actually okay, as there's only O(200) materials tracked over the system. The issue is that in `geometry.xml` the cell instance list that describes which TRISO instance corresponds to which material is so long that it can't be read.

In my standalone OpenMC simulations, I get around this by writing the `cell.fill` to a `npy` file e.g.,
```python
np.save("fuel_fill_list.npy", TRISO_fuel_cell.fill)
```
Later, when I want to deplete this `model` with multiphysics properties loaded, I'll load a version of `geometry.xml` that is not zoned (i.e. the same material is inside every TRISO sphere). And then after that `model` is loaded, I will access that cell via `openmc.geometry.get_all_cells()` with the right cell id and assign it the `cell.fill` by reading in the list from file via `np.load`
```python
TRISO_fuel_cell_id = 1
TRISO_fuel_cell = cell_dict[TRISO_fuel_cell_id]
TRISO_fuel_cell.fill = np.load("fuel_fill_list.npy", allow_pickle=True)
```

## Design

Looking for advice on this. I think if we can get the data into h5 format (which can happen outside Cardinal), we should be able to take advantage of OpenMC's h5 processing and the extern method [‎openmc_cell_set_fill‎](https://github.com/openmc-dev/openmc/blob/df6f94300f6f9013100073962ffc417f4c32c754/src/cell.cpp#L1245). What I'm not sure about is if this needs to be a UserObject or if we can just set some variables in `OpenMCCellAverageProblem` for the cell ID and the filename of the cell fill list.

A feature like this might be nice for those seeking to set specific temperatures for TRISOs, e.g. those looking at reactivity effects of changing TRISO temp compared to the general fuel compact computed temperature.

## Impact

It will add something new and shouldn't impact the existing API. I'm not sure if XML capabilities will improve one day to be able to handle these XML file with extremely long individual sub-elements, but it may quicker to go through h5 in general. Reading from the binary format only takes order of minutes in my experience.

PR #1380, opened 2026-07-07
Cell modifier

aprilnovak:
Refs #1378

Another follow-on PR can provide same functionality, but with the material_ids given in a file


src/modelmodifiers/ModelModifiersBase.C
```diff
@@ -0,0 +1,48 @@
...
+
+#ifdef ENABLE_OPENMC_COUPLING
+
+#include "ModelModifiersBase.h"
+#include "UserErrorChecking.h"
+#include "VariadicTable.h"
+#include "BrentsMethod.h"
+
+// To disable tallies
+#include "openmc/tallies/tally.h"
```
gonuke: Are these necessary here?

include/modelmodifiers/ModelModifiersBase.h
```diff
@@ -0,0 +1,37 @@
...
+/*                With the U. S. Department of Energy               */
+/*                                                                  */
+/*                 See LICENSE for full restrictions                */
+/********************************************************************/
+
+#pragma once
+
+#include "MooseObject.h"
+#include "OpenMCBase.h"
+#include "CardinalEnums.h"
```
gonuke: Is this needed?

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
```diff
@@ -0,0 +1,18 @@
+# OpenMCCellMaterialFill
+
+## Description
+
+`OpenMCCellMaterialFill` is a [AddModelModifiers](AddAddModelModifiersAction.md) that
```
gonuke:
```suggestion
`OpenMCCellMaterialFill` is a [AddModelModifiers](AddModelModifiersAction.md) that
```

doc/content/source/actions/AddModelModifiersAction.md
```diff
@@ -0,0 +1,12 @@
+# AddModelModifiersAction
```
gonuke, line 1: Not sure the practice/precedent, but this seems like it makes more sense to be `ModelModifierBase`

gonuke (changes requested):
Thanks for the quick solution @aprilnovak - just a few comments here - haven't investigated failing tests

src/modelmodifiers/ModelModifiersBase.C
aprilnovak: nope! was moving kind of fast to get this implemented, I will remove the unecessary headers (got copied over from a file I started from)

include/modelmodifiers/ModelModifiersBase.h
aprilnovak: nope! also stray from a copy

doc/content/source/actions/AddModelModifiersAction.md
aprilnovak: do you mean not have the action class, or that the text here belongs in the documentation for `ModelModifierBase`?

gonuke: reviewing other examples more carefully. I'll make my comment in the `OpenMCCellMaterialFill` instead

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
```diff
@@ -0,0 +1,18 @@
+# OpenMCCellMaterialFill
+
+## Description
+
+`OpenMCCellMaterialFill` is a [AddModelModifiers](AddModelModifiersAction.md) that
```
gonuke: Looking at the way other actions are documented, I think this should say:

```suggestion
`OpenMCCellMaterialFill` is a [ModelModifier](AddModelModifiersAction.md) that
```

gonuke (changes requested):
A new comment in the `OpenMCCellMaterialFill`

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
aprilnovak: oh! I see - yes, totally agree

gonuke (approved):
Looks good to me.... Maybe not for now, but makes me wonder if some of the other de facto model modifiers that we use for criticality searches should be refactored as `ModelModifier`s?

aprilnovak:
> makes me wonder if some of the other de facto model modifiers that we use for criticality searches should be refactored as ModelModifiers?

Yes, the same thing went through my mind when implementing this - perhaps the design would be:

- Move all criticality search "modifier" stuff into modifier-derived classes
- Each criticality search takes as input the name of a modifier to use. Effectively reduces criticality search to just 1 class, moves all criticality-search-derived classes into modifiers.

```
[Problem]
  [ModelModifiers]
    [d]
       type = OpenMCMaterialDensity
       material = '2'
    []
  []

  [CriticalitySearch]
    model_modifier = d
    root_tol = 1e-2
    # other search params...
  []
[]
```

Although I suppose that not all model modifiers would qualify for criticality searches. For instance, the one added in this PR just does one operation, there's no independent variable/root to search for.

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
```diff
@@ -0,0 +1,18 @@
+# OpenMCCellMaterialFill
+
+## Description
+
+`OpenMCCellMaterialFill` is a [ModelModifier](AddModelModifiersAction.md) that
+changes a cell's material fill before executing Cardinal.
```
lewisgross1296:
```suggestion
`OpenMCCellMaterialFill` is a [ModelModifier](AddModelModifiersAction.md) that
changes an `openmc.Cell`'s fill before executing Cardinal. Note that while an 
`openmc.Cell` can normally also be filled with an `openmc.Universe` or an
`openmc.Lattice`, theis`ModelModifier` can only replace the fill with an
`openmc.Material` or a list of `openmc.Material`s, also referred to as a `distribmat`.
```

doc/content/source/actions/AddModelModifiersAction.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Overview
+
+The `AddModelModifiersAction` allows you to modify an OpenMC model prior to running
+a simulation with Cardinal. Examples of modifications that can be made includes:
+
+- [OpenMCCellMaterialFill](OpenMCCellMaterialFill.md), to change a cell's material fill
```
lewisgross1296:
```suggestion
- [OpenMCCellMaterialFill](OpenMCCellMaterialFill.md), to change an `openmc.Cell`'s fill to an `openmc.Material` or a list of `openmc.Material`s (referred to as a `distribmat`)
```

src/actions/AddModelModifiersAction.C
```diff
@@ -0,0 +1,54 @@
...
+#include "OpenMCCellAverageProblem.h"
+#include "ModelModifiersBase.h"
+
+registerMooseAction("CardinalApp", AddModelModifiersAction, "add_model_modifiers");
+
+InputParameters
+AddModelModifiersAction::validParams()
+{
+  auto params = MooseObjectAction::validParams();
+  params.addClassDescription("Adds a modifiers that changes the OpenMC model prior to execution");
```
lewisgross1296:
```suggestion
  params.addClassDescription("Adds one or more modifiers that change the OpenMC model prior to execution");
```

src/base/CardinalApp.C
```diff
@@ -235,6 +235,11 @@ CardinalApp::associateSyntaxInner(Syntax & syntax, ActionFactory & /* action_fac
   registerTask("add_criticality_search", false /* is required */);
   addTaskDependency("add_criticality_search", "init_problem");
 
+  // Add the [Problem/ModelModifiers] block
+  registerSyntaxTask("AddModelModifiersAction", "Problem/ModelModifiers/*", "add_model_modifiers");
```
lewisgross1296, line 239: Out of curiosity, does the * just add every ModelModifier specified?

src/modelmodifiers/OpenMCCellMaterialFill.C
```diff
@@ -0,0 +1,83 @@
...
+
+InputParameters
+OpenMCCellMaterialFill::validParams()
+{
+  auto params = ModelModifiersBase::validParams();
+  params.addRequiredParam<int32_t>("cell_id", "Cell ID to modify");
+  params.addRequiredParam<std::vector<int32_t>>(
+      "material_ids",
+      "IDs of the materials to fill into the cell; the length of this array must match the number "
+      "of instances of the cell");
```
lewisgross1296, line 36: In theory, with N cell instances, I think it's valid OpenMC to specify either one or N materials. We currently prohibit the case in which only one is specified in the constructor below, though I can't think of a use case where we'd want this. Probably fine as is, just figured I'd mention it.

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Example Input File Syntax
+
+The following input will modify the material in cell 2; there are three instances of cell
+2, and each is modified to material 1.
+
+!listing test/tests/model_modifiers/openmc_cell_material/openmc.i
```
lewisgross1296:
```suggestion
!listing test/tests/model_modifiers/cell_material_fill/openmc.i‎
```

test/tests/openmc_errors/incorrect_problem/openmc.i
lewisgross1296, line 1: Is this just something else unrelated you caught during this PR?

lewisgross1296 (commented):
Some small thoughts on wordings and some questions I had, nothing major. Also I think one of the !listing blocks had an old filename

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
```diff
@@ -0,0 +1,18 @@
...
+  block=Problem
+
+!syntax parameters /Problem/AddModelModifiers/OpenMCCellMaterialFill
+
+!syntax inputs /Problem/AddModelModifiers/OpenMCCellMaterialFill
```
lewisgross1296:
```suggestion
!syntax parameters /Problem/ModelModifiers/OpenMCCellMaterialFill

!syntax inputs /Problem/ModelModifiers/OpenMCCellMaterialFill
```
I think this might be the correct syntax based on a doc error from CIVET

test/tests/openmc_errors/incorrect_problem/openmc.i
aprilnovak: yep! It's not a bug or anything, but I copied this file to add the test in this PR, and noticed it had extra fluff which was not needed for the test's sake.

src/modelmodifiers/OpenMCCellMaterialFill.C
aprilnovak: True, we could relax the check as-in and allow either 1 or equal to the number of instances. I'll update that in this PR

src/base/CardinalApp.C
aprilnovak: yep!

doc/content/source/actions/AddModelModifiersAction.md
aprilnovak: I'd prefer some wording in-between the two suggestions, since the user doesn't actually provide the `openmc.Material` (so I don't think it makes sense to refer to the class name like that). How about:

```
[OpenMCCellMaterialFill](OpenMCCellMaterialFill.md), to set the material filling an OpenMC cell
```

doc/content/source/modelmodifiers/OpenMCCellMaterialFill.md
aprilnovak: same comment, I don't think it's necessary to refer to class names here - the user doesn't need that level of detail for this object. What about:

```
`OpenMCCellMaterialFill` is a [ModelModifier](AddModelModifiersAction.md) that
set the fill of an OpenMC cell to a user-specified material or list of materials (for distributed cells).
```

lewisgross1296: I'm happy with that! I think you want "sets the fill" (typo)

doc/content/source/actions/AddModelModifiersAction.md
lewisgross1296: For this one, maybe "to set the materials filling each instance of an OpenMC cell"?

src/actions/AddModelModifiersAction.C
aprilnovak: This is not actually how this works - one action is created to construct each ModelModifier, not one action doing all of them. But I did fix the typo in the grammar

src/modelmodifiers/OpenMCCellMaterialFill.C
aprilnovak: Has been updated, added another test

src/modelmodifiers/OpenMCCellMaterialFill.C
```diff
@@ -0,0 +1,91 @@
...
+    if (err)
+      paramError("material_ids",
+                 "Material with ID " + std::to_string(m) + " was not found in the OpenMC model!");
+    _material_indices.push_back(material_index);
+  }
+
+  if (n_mats == 1)
+  {
+    // make _material_indices equal in length to the number of cell instances
+    for (int i = 1; i < n_mats; ++i)
```
gonuke: Shouldn't this be
```suggestion
    for (int i = 1; i < n_cell_indices; ++i)
```
?

src/modelmodifiers/OpenMCCellMaterialFill.C
```diff
@@ -0,0 +1,91 @@
...
+      _material_indices.push_back(material_index);
```
gonuke: Could we just use `assign` here:
```suggestion
    // make _material_indices equal in length to the number of cell instances
    _material_indices.assign(n_cell_indices, material_index);
```

gonuke (changes requested):
I think there was a bug in the single material id case, and one simplification in that same place.

src/modelmodifiers/OpenMCCellMaterialFill.C
aprilnovak: oh true - thanks for the catch! I guess the test I wrote still passes because perhaps OpenMC handles the logic of applying that one material to each instance if only one is passed in? In any case, I've fixed this line to make it more explicit in Cardinal.

lewisgross1296: Is this just for consistency so that we can always assume `_material_indices` is the same length as the number of cell instances?

aprilnovak: yes, once we get to the point in the modifier class that goes to apply those materials to OpenMC

gonuke: > Is this just for consistency so that we can always assume `_material_indices` is the same length as the number of cell instances?

This now matches what the OpenMC python method does, also.

test/tests/model_modifiers/cell_material_fill/tests
```diff
@@ -0,0 +1,44 @@
...
+    input = openmc.i
+    csvdiff = openmc_out.csv
+    requirement = "The system shall be able to modify the material(s) filling a cell prior to executing OpenMC. This is verified by creating a standalone OpenMC model with the final set of materials, compared to a different starting model whose cell's fill is modified via Cardinal."
+    capabilities = 'openmc'
+  []
+  [change_material_all_instances]
+    type = CSVDiff
+    input = openmc.i
+    csvdiff = openmc_out.csv
+    cli_args = 'Problem/ModelModifiers/one/material_ids="1"'
```
meltawila, line 40: should we add a test for non-uniform distributed cell: material_ids="1 2 1"? both CSVDiff tests here apply material 1 to all instances but the non-uniform fills would be the main use of this object. should the order in material_ids follow the cell instance order?

aprilnovak: yes, I can add a test for this!

The material IDs must follow the cell instance order - so it's up to the user to get that part correct.

meltawila:
What happens if a zero instance cell is used (e.g. lattice.outer)? for example we do this when applying temperatures in OpenMCProblemBase.C:
```
    // special error message if cell has zero instances
    if (cellHasZeroInstances(cell_info))
      mooseError("Failed to set the temperature for cell " + printCell(cell_info) +
                 " with zero instances.");
```
Does this need a similar mooseError? ~the above error might appear first anyway so this might not be important~ a user can still pass the cell_id for a cell that doesn't map to the mesh mirror and get to this error I think

src/modelmodifiers/OpenMCCellMaterialFill.C
```diff
@@ -0,0 +1,90 @@
...
+  int err = openmc_get_cell_index(_cell_id, &_cell_index);
+  if (err)
+    paramError("cell_id",
+               "Cell ID " + std::to_string(_cell_id) + " was not found in the OpenMC model!");
+
+  // check that the length of the materials matches the number of instances
+  const auto & material_ids = getParam<std::vector<int32_t>>("material_ids");
+  auto n_cell_instances = openmc::model::cells[_cell_index]->n_instances();
+  auto n_mats = std::size(material_ids);
+  if (n_mats != 1 && (n_cell_instances != n_mats))
```
meltawila, line 65:
```suggestion
  if (n_cell_instances == 0)
    paramError("cell_id",
             "Cell ID " + std::to_string(_cell_id) +
                 " exists in the OpenMC model but has zero instances so its material fill "
                 "cannot be modified by OpenMCCellMaterialFill");
  if (n_mats != 1 && (n_cell_instances != n_mats))
```

include/base/OpenMCProblemBase.h
```diff
@@ -593,4 +594,7 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
 
   /// Object to use for a criticality search
   CriticalitySearchBase * _criticality_search = nullptr;
+
+  /// Objects to use for model modifications prior to execution
+  std::vector<ModelModifiersBase *> _model_modifiers;
```
meltawila: I don't see this used anywhere? `mm_objs` is what is used in OpenMCProblemBase

src/modelmodifiers/OpenMCCellMaterialFill.C
aprilnovak: I do think we need an error like this, but FWIW regarding your earlier question/comment - this object in this PR should work find even if the cell does not map to the [Mesh]. It's just changing the underlying model in OpenMC

src/modelmodifiers/OpenMCCellMaterialFill.C
```diff
@@ -0,0 +1,90 @@
...
+registerMooseObject("CardinalApp", OpenMCCellMaterialFill);
+
+InputParameters
+OpenMCCellMaterialFill::validParams()
+{
+  auto params = ModelModifiersBase::validParams();
+  params.addRequiredParam<int32_t>("cell_id", "Cell ID to modify");
+  params.addRequiredParam<std::vector<int32_t>>(
+      "material_ids",
+      "IDs of the materials to fill into the cell; the length of this array must match the number "
```
meltawila:
```suggestion
      "IDs of the materials to fill into the cell; the length of this array can be one material ID that will"
      " be applied to all instances. Otherwise, it must match the number "
```

test/tests/model_modifiers/cell_material_fill/tests
aprilnovak: added another test

src/modelmodifiers/OpenMCCellMaterialFill.C
aprilnovak: Thanks for catching this! I forgot about zero-instance cells. I've added a test covering this.

meltawila (approved):
LGTM!

PR #1384, opened 2026-07-08
Modify trigger / cell relaxation tally tests to harden them against floating point issues / threading

### LINKED ISSUE #1383: Harden OpenMC tests against floating point / threading differences
## Reason
We occasionally run into issues where differences in math function implementations between compilers / architectures causes tests to start failing. Threading also occasionally causes problems with OpenMC where floating-point non-associativity yields different results. We should harden the OpenMC test suite against these problems, notably tests which run OpenMC multiple times (such as triggers or relaxation). 

## Design
Re-implement and re-gold OpenMC trigger tests to change what they test. In particular, we should save the result of a boolean comparison between the tally/k stochastic uncertainty quantities (standard deviation, relative error, variance) and the trigger threshold. This satisfies the testing requirements (ensuring quantities converge to the trigger) without needing to gold the computed quantities (thereby avoiding the issues mentioned).  

We should also loosen the relative difference criteria on the relaxation tests. The purpose of these tests is to ensure the different relaxation schemes are implemented correctly, which can be tested with a looser tolerance.

## Impact
Harden tests against future changes in testing hardware or dependencies (OpenMC and MOOSE). This also allows the `devel` -> `master` submodule update to proceed (which has been disabled the past few months due to failed tests).

nuclearkevin:
I've changed the trigger tests to gold against whether or not they meet the trigger threshold (instead of storing tallied quantities). I've also upped the relative error on the `CSVDiff` / `ExoDiff` tests for cell tally relaxation to ensure they pass (after checking to make sure relaxation is being applied correctly). Hopefully things start passing on `Test rocky9-mpich` now.

It's rather strange that only the cell tally tests run into this problem, and it was caused by changes in a MOOSE submodule update (not an OpenMC submodule update)...

Closes #1383 

I also took the opportunity to tag some tests while I was here to move the needle on meeting SQA requirements.


test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -42,6 +45,7 @@
                   "comparing the heat source computed via relaxation with the un-relaxed iterations from the "
                   "same input file, but with the relaxation part commented out in the source code (so that we can compare "
                   "directly against runs that only differ by changing the number of particles."
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -74,6 +80,7 @@
     requirement = "A unity relaxation factor shall be equivalent to an unrelaxed case with "
                   "globally-normalzed cell tallies on a model with imperfect alignment between OpenMC model and the "
                   "mesh mirror."
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -85,6 +92,7 @@
                   "with imperfect alignment between the OpenMC model and the mesh mirror. This test is verified by "
                   "comparing the heat source computed via relaxation with the un-relaxed iterations from the "
                   "openmc_nonaligned.i case (without additional command line parameters)"
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -52,6 +56,7 @@
     cli_args = "Problem/normalize_by_global_tally=false Problem/relaxation=constant Problem/relaxation_factor=1.0"
     requirement = "A unity relaxation factor shall be equivalent to an unrelaxed case with "
                   "locally-normalzed cell tallies on a model with perfect alignment between OpenMC model and the mesh mirror"
+    rel_err = '1e-3'
```
aprilnovak, line 58: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -63,6 +68,7 @@
                   "with perfect alignment between the OpenMC model and the mesh mirror. This test is verified by "
                   "comparing the heat source computed via relaxation with the un-relaxed iterations from the "
                   "openmc.i case with normalize_by_global_tally=false."
+    rel_err = '1e-3'
```
aprilnovak, line 70: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -96,6 +104,7 @@
     min_parallel = 2
     requirement = "A unity relaxation factor shall be equivalent to an unrelaxed case with "
                   "locally-normalzed cell tallies on a model with imperfect alignment between OpenMC model and the mesh mirror"
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -108,6 +117,7 @@
                   "with imperfect alignment between the OpenMC model and the mesh mirror. This test is verified by "
                   "comparing the heat source computed via relaxation with the un-relaxed iterations from the "
                   "openmc_nonaligned.i case with normalize_by_global_tally=false"
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -126,13 +137,15 @@
...
     requirement = "The wrapping shall allow output of the unrelaxed heat source for cell tallies"
+    rel_err = '1e-3'
     capabilities = 'openmc'
   []
   [multi_tally]
     type = CSVDiff
     input = multi_tally.i
     csvdiff = multi_tally_out.csv
     requirement = "The wrapping shall allow output of multiple tally scores, with relaxation applied independently to each score"
+    rel_err = '1e-3'
```
aprilnovak: I think this test actually was not failing before, so we can remove?

aprilnovak:
Thanks for fixing this issue @nuclearkevin!

Issue #1392, opened 2026-07-16
Treat density similar to temperature for feedback in OpenMC


aprilnovak:
## Reason
Historically, we could only modify density if we had a unique OpenMC material object; however, since the inclusion of density multipliers in OpenMC, we can relax a lot of constraints related to density feedback.

## Design
We should support:

- [x] Density feedback to cells which contain universes and/or lattices; these should set the density on all contained material cells
- [ ] `identical_cell_fills` should be usable for density
- [ ] avoid creation of duplicate materials when filling skinned regions for each density bin in `MoabSkinner`

## Impact
Improved user experience and code maintenance.

PR #1393, opened 2026-07-16
Nested density

aprilnovak:
This PR updates Cardinal to treat density in the same manner as temperature, from the perspective of feedback into OpenMC. Historically, OpenMC needed to have a unique material to apply a unique density at a given position; now, with density multipliers, this is no longer the case - but there remained artifacts of the older approach within how we treat coupling. This PR takes the first step towards fully treating density equivalently to temperature. This PR entails:

Allowing density feedback to be applied from Cardinal to OpenMC cells which are not simple material-fill cells. The approach works the same as temperature: if we apply a density to an OpenMC cell filled by a universe or lattice, we recurse down and apply that density to all the nested cells.

Tests are added that confirm correct behavior for OpenMC cells filled by lattices or universes.

Points to note:

- `CellTemperatureAux` returns the temperature of the first material-fill cell, if querying the temperature of an OpenMC cell not filled by a material directly. I adopted the same approach for density, with one caveat: for void cells, if that void cell happens to be the first "simple" contained cell, `CellDensityAux` could give the wrong impression that ALL cells have zero density. I decided to screen out void cells when reporting `CellDensityAux` if reporting the density of an OpenMC cell filled by multiple material cells. This is not an issue for temperature, because OpenMC is still perfectly happy to apply a non-zero temperature to a void cell (but cannot apply non-zero density to a void cell). Therefore, `CellTemperatureAux` will also still return the temperature of the first non-void cell, though it would also have been okay to just return the temperature of a void cell since it is allowed to be set to a nonzero value.
- For void cells, I modified `CellDensityAux` to return zero as the density, since this makes more sense than -1 and is compatible with the bullet point above.  
- Because with the density multiplier, we are applying density to cells (and no longer materials), we don't need to throw an error if Cardinal sets the density on a void cell (previously, we would ask OpenMC to set the density on a void material, which doesn't really exist, so this would cause a cryptic error message). We can remove that error message and associated test.


src/base/OpenMCCellAverageProblem.C
```diff
@@ -2490,7 +2456,29 @@ OpenMCCellAverageProblem::sendTemperatureToOpenMC() const
...
   const auto & contained_cells = containedMaterialCells(cell_info);
+  for (const auto & c : contained_cells)
+  {
+    const auto & cell = openmc::model::cells[c.first];
+    for (const auto & instance : c.second)
+    {
+      const auto mat_index = cell->material(instance);
+      if (mat_index != openmc::MATERIAL_VOID)
+      {
+        cellInfo first_cell = {c.first, instance};
```
meltawila, line 2475: This now changes the behavior when used in `OpenMCCellAverageProblem::cellTemperature()` as well since it searches for non-void fill first. The PR description implies that `CellTemperatureAux` won't do this non-void screening?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2536,7 +2525,34 @@ OpenMCCellAverageProblem::sendDensityToOpenMC() const
...
+      {
+        cellInfo ci = {contained.first, instance};
+        if (cells_already_set.count(ci))
+        {
+          double rho;
+          openmc_cell_get_density(ci.first, &ci.second, &rho);
+
+          mooseError("Cell " + std::to_string(cellID(contained.first)) + ", instance " +
+                     std::to_string(instance) + " has already had its density set by Cardinal to " +
+                     std::to_string(rho) +
```
meltawila:
```suggestion
                     std::to_string(rho) + "g/cm3"
```

src/base/OpenMCProblemBase.C
```diff
@@ -631,27 +634,10 @@ OpenMCProblemBase::setCellDensity(const Real & density, const cellInfo & cell_in
...
-
-  if (!is_material_cell)
-    mooseError(
-        "Density transfer does not currently support cells filled with universes or lattices!");
-
-  // throw a special error if the cell is void, because the OpenMC error isn't very
-  // clear what the mistake is
-  if (material_index == MATERIAL_VOID)
-  {
-    mooseWarning("Skipping setting density for cell " + printCell(cell_info) +
```
meltawila, line 645: it's not clear to me what happens now if a density was assigned to a void cell, does OpenMC just ignore it?

src/base/OpenMCCellAverageProblem.C
meltawila, line 441: should this error be removed now? I am not sure

meltawila (changes requested):
Just a few comments below

src/base/OpenMCProblemBase.C
aprilnovak: We now use a density multiplier - if a material is at a density of `rho` in the XML files, then Cardinal modifies this density by setting a multiplier, like `multiplier * rho`. So, if the initial density is zero it remains zero.

src/base/OpenMCCellAverageProblem.C
aprilnovak: true, I will make that documentation clearer - but the behavior isn't actually any different for temperature from how `CellTemperatureAux` worked before, because the void cell _can_ have a non-zero temperature, unlike density.

src/base/OpenMCCellAverageProblem.C
aprilnovak: I left it for now because I plan to add tests for that in a future PR - I think technically it might be doable right now with the changes in this PR, but I decided to keep it separate. I left it as a checkbox here: https://github.com/neams-th-coe/cardinal/issues/1392

meltawila (approved):
looks good to me

nuclearkevin:
Sorry about the wait - I'll get my review in EoD today.

nuclearkevin (approved):
LGTM, thanks @aprilnovak!

PR #1401, opened 2026-07-22
Fix duplicate cell-instance when using Skinner with non-root DAGMC Universe

### LINKED ISSUE #1400: Duplicate cell-instance when using Skinner with non-root DAGMC Universe
## Bug Description

When using MoabSkinner with a geometry where a CSG cell is filled with a DAGMCUniverse, the OpenMCVolumeCalculation is wrong. Specifically,
After N reskins, every DAGMC-derived cell reports `instance k (of N+1)`, and `Actual Vol ≈ Mapped Vol / (N+1)` .

## Steps to Reproduce

- OpenMC model must contain a CSG cell that is filled with a DAGMCUniverse.
- Use MoabSkinner and set `cell_level = 1` in openmc.i

Can reproduce in csg-to-cad tutorial (must use a Cardinal build without Double-Down ([possible issue](https://github.com/neams-th-coe/cardinal/discussions/1278#discussioncomment-15593251))).
Modifications: in openmc.i, add
```
  verbose = true
  volume_calculation = vol
```
to the Problem block, and
```
  [vol]
    type = OpenMCVolumeCalculation
    n_samples = 100000
  []
```
to the UserObjects.

Example log:
```
after first reskin:
|           Cell           |    T   |  rho  | T+rho | Other | Mapped Vol |       Actual Vol        |
----------------------------------------------------------------------------------------------------
| 627, instance  1 (of  2) | 134882 |     0 |     0 |     0 | 1.865e+03  | 9.266e+02 +/- 3.248e+00 |

after second reskin:
| 627, instance  2 (of  3) | 129273 |     0 |     0 |     0 | 1.470e+03  | 4.885e+02 +/- 2.731e+00 |
| 628, instance  2 (of  3) |   5607 |     0 |     0 |     0 | 3.942e+02  | 1.292e+02 +/- 1.834e+00 |
| 629, instance  2 (of  3) |      1 |     0 |     0 |     0 | 1.449e-01  | 9.973e-03 +/- 1.727e-02 |

after third reskin:
| 627, instance  3 (of  4) | 129311 |     0 |     0 |     0 | 1.471e+03  | 3.667e+02 +/- 2.365e+00 |
| 628, instance  3 (of  4) |   5566 |     0 |     0 |     0 | 3.929e+02  | 9.651e+01 +/- 1.586e+00 |
| 629, instance  3 (of  4) |      1 |     0 |     0 |     0 | 1.305e-01  | 1.496e-02 +/- 2.115e-02 |
```

## Impact

The volume calculation is just a diagnostic so it results in the same solution but this bug causes one to doubt the Skinner's results.

## Cause
In [OpenMCCellAverageProblem::reloadDAGMC()](https://github.com/neams-th-coe/cardinal/blob/devel/src/base/OpenMCCellAverageProblem.C#L2941), `openmc::populate_universes()` is called on every reskin. This works fine when the root is a DAGMCUniverse since they are rebuilt each time. However, when the root universe contains the CSG cell (which is filled by DAGMCUniverse), the root is never removed, so its `cells_` vector is never cleared before the next `populate_universes()` call. Each reskin appends another duplicate index of the DAGMC-filled CSG cell into the root universe's `cells_`.
Since the wrapper cell now appears N+1 times in the root universe's list, every DAGMC cell's `n_instances_` =N+1. Ultimately, this causes the observed bug of the Actual Volume calculation being reduced by a factor of N+1.

## A Fix
Clear the `cells_` on all the universes before re-running `populate_universes()` in `reloadDAGMC()`.
```
    for (auto & universe : openmc::model::universes)
      universe->cells_.clear();
```
This clears the `cells_` and then `populate_universes()` rebuilds the cell lists from scratch.
This fixes the error on my end and I don't believe it affects any other functionality.

ssgudala-bwxt:
Closes [#1400](https://github.com/neams-th-coe/cardinal/issues/1400)

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -2905,6 +2905,10 @@ OpenMCCellAverageProblem::reloadDAGMC()
   _console << "Re-generating OpenMC model with " << openmc::model::cells.size() << " cells... "
            << std::endl;
 
+  // Clear cells on all surviving universes
+  for (auto & universe : openmc::model::universes)
+    universe->cells_.clear();
+  
   // Add cells to universes
   openmc::populate_universes();
 
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -2908,7 +2908,7 @@ OpenMCCellAverageProblem::reloadDAGMC()
   // Clear cells on all surviving universes
   for (auto & universe : openmc::model::universes)
     universe->cells_.clear();
-  
+
   // Add cells to universes
   openmc::populate_universes();
 
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -2905,7 +2905,7 @@ OpenMCCellAverageProblem::reloadDAGMC()
   _console << "Re-generating OpenMC model with " << openmc::model::cells.size() << " cells... "
            << std::endl;
 
-  // Clear cells on all surviving universes
+  // Clear cells on all surviving universes.
   for (auto & universe : openmc::model::universes)
     universe->cells_.clear();
 
--- a/test/tests/neutronics/dagmc/with_csg/tests
+++ b/test/tests/neutronics/dagmc/with_csg/tests
@@ -11,3 +11,20 @@
     capabilities = 'dagmc'
   []
 []
+
+[Tests]
+  issues = '#1400'
+  design = 'MoabSkinner.md OpenMCCellAverageProblem.md'
+
+  [cell_instances_stable_across_reskins]
+    type = RunApp
+    input = openmc.i
+    expect_out = 'of 1\)'
+    absent_out = 'of [2-9]\)'
+    requirement = "The system shall not artificially inflate an OpenMC cell's instance count "
+                  "when the MOAB skinner reloads DAGMC geometry multiple times for a "
+                  "CSG-wrapped DAGMC universe."
+    mesh_mode = 'replicated'
+    capabilities = 'dagmc'
+  []
+[]


aprilnovak:
Thanks @ssgudala-bwxt! Can you do a force push after accepting the `cardinal-devs` invite? That will then launch the test suite.

nuclearkevin (changes requested):
Thanks for this fix @ssgudala-bwxt - It looks good to me! 

Would you be willing to add a test to Cardinal to make sure we don't hit this issue again? If you don't have the bandwidth that's fine, I can add the test if you're OK with me pushing to your branch (I'll then pull someone else in for a review).

ssgudala-bwxt:
@nuclearkevin Added a [test for cell accumulation](https://github.com/neams-th-coe/cardinal/pull/1401/commits/cc943933a7c0d5b17a2adcb9d7f4a6ac2f041acf). Let me know if this works.

nuclearkevin (approved):
The test looks great! It's a pretty clever way of looking for the failure - I had no idea `expect_out` and `absent_out` were possible parameters for `RunApp`. Once I convince the NekRS tests to pass (the failures look unrelated but its always good to check), I'll merge.

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2937,6 +2905,10 @@ OpenMCCellAverageProblem::reloadDAGMC()
   _console << "Re-generating OpenMC model with " << openmc::model::cells.size() << " cells... "
            << std::endl;
 
+  // Clear cells on all surviving universes
```
nuclearkevin:
```suggestion
  // Clear cells on all surviving universes.
```

nuclearkevin: Just a note. For our SQA process we need one commit in the PR to reference an associated issue. I tried to commit directly to your branch, but Github wouldn't let me do that so this was my workaround. Apologies for the spam!

ssgudala-bwxt: No worries!

PR #1404, opened 2026-07-23
Update OpenMC submodule

### LINKED ISSUE #1389: Update OpenMC submodule to include PeriodicBC fix
## Reason

In my GCMR model, coolant flows in the y-direction, and so I made a one sixth core model with periodic boundary conditions. While it was a gas coolant, I figured it would be better to allow the gravity vector to be perpendicular to flow for the THM model, since this is the actual design of the system. This exposed an issue with OpenMC which was just resolved in this [PR](https://github.com/openmc-dev/openmc/pull/4015). Essentially, if anyone tried to use a periodic BC for a non-z-aligned reactor, it could fail during transport especially in a multiphysics simulation with many OpenMC evaluations. This is big for anyone who wants to do a model with periodic boundary conditions and use a non-z-aligned flow direction as many reactor models would want to exploit symmetry. I imagine other reactors will have a much bigger impact from horizontal flow on the fluid solutions, and so this should help out a lot with those models.

## Design

Update submodule and regold any tests (likely none though since no OpenMC tests were not regolded and IIRC Cardinal does not have any periodic tests)

## Impact

Should not affect the API unless there's some other newer commits that require changing Cardinal. However, the last [update OpenMC PR](https://github.com/neams-th-coe/cardinal/pull/1382) was very recent.

lewisgross1296:
I'm mainly updating this to include the change to the periodic boundary condition, but this also includes any other newer commits merged from then until now.

Closes #1389.


lewisgross1296:
Looks like @nuclearkevin beat me to it while I was waiting for local tests :)

PR #1410, opened 2026-08-02
Reduce memory usage when applying temperature/density feedback

### LINKED ISSUE #1409: Decrease memory usage when applying temperature/density feedback
## Reason
Recently, I've been attempting to run the gas assembly multiphysics tutorial with delta tracking to ensure the majorant is robust under multiphysics feedback. I found that the tutorial transiently consumes upwards of 60 GB of memory per MPI rank, which makes it untenable to run for the average user without HPC resources. I've tracked this down to a few places in Cardinal which have to do with temperature/density feedback:

1. The identical cell fill optimization forces `containedMaterialCells(...)` to return a deep copy of a mapping data structure that contains every cell index + instance pair contained in a given cell+instance. In TRISO problems this yields very large transient increases in memory when applying feedback in either `sendTemperatureToOpenMC(...)` or `sendDensityToOpenMC(...)`.

2. An error was added to `sendTemperatureToOpenMC(...)` and `sendDensityToOpenMC(...)` to let the user know if a cell temperature or density has been set multiple times. This error message uses a temporary `std::unordered_set` of all cell index/instance permutations, which yields a large transient increase in memory when applying feedback in either `sendTemperatureToOpenMC(...)` or `sendDensityToOpenMC(...)` in TRISO problems. 

Reworking `sendTemperatureToOpenMC(...)` and `sendDensityToOpenMC(...)` to not use cached material cell information and commenting out the error message results drops the maximum memory consumption of the gas assembly tutorial to ~20 GB. The downside of the "simple fix" is that it takes 4x longer to send properties to OpenMC as material cells are being set with a geometry traversal. A more complicated but superior fix is described below.

## Design
1. To fix the transient increase in memory while retaining the identical cell fill and material caching optimzations, the indices in `_cell_to_contained_material_cells` should be perturbed in-place. This does increase code complexity but saves on a memory a decent amount.

2. To fix the transient increase in memory caused by the error check being applied in `sendTemperatureToOpenMC(...)` or `sendDensityToOpenMC(...)`  a few optimizations should be used. First, the check can be moved to `OpenMCCellAverageProblem::cacheContainedCells()` to avoid lining it up with auxsystem serialization (which is extremely memory intensive). This has the added bonus of decreasing code duplication and allowing for the removal of a test. Next, the check can be restricted to a single MPI rank to avoid duplicating the large unordered set required. Finally, the cells with identical fills do not need to be checked as they are already verified when caching contained cells.

Combining these optimizations decreases peak memory usage to ~22 GB, which is in-line with the memory usage when cells are set recursively!

## Impact
Large decrease in peak memory usage when running OpenMC problems with feedback. Performance enhancements as large data structures won't be copied.

nuclearkevin:
Closes #1409


src/base/OpenMCCellAverageProblem.C
```diff
@@ -1827,6 +1830,51 @@ OpenMCCellAverageProblem::cacheContainedCells()
...
+  // cell index + instance pairs, so we ensure its only performed on a single MPI rank
+  // and outside of any point where the auxsystem is serialized to reduce peak
+  // memory consumption.
+  if (_communicator.rank() == 0)
+  {
+    std::unordered_set<cellInfo> cells_already_set;
+    for (const auto & [cell_info, elements] : _cell_to_elem)
+    {
+      // Skip checking the identical cell fills outside of _first_identical_cell.
+      // These mapping errors are caught when verifying contained cells above.
```
```suggestion
      // These mapping errors are caught when verifying contained cells above (if 'check_identical_cell_fills' is true)
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1827,6 +1830,51 @@ OpenMCCellAverageProblem::cacheContainedCells()
...
+
+      const auto & contained_cells = _cell_to_contained_material_cells.at(cell_info);
+      for (auto & [cc_idx, cc_instances] : contained_cells)
+      {
+        for (auto cc_instance : cc_instances)
+        {
+          if (cells_already_set.count({cc_idx, cc_instance}))
+            mooseError("Cell " + std::to_string(cellID(cc_idx)) + ", instance " +
+                       std::to_string(cc_instance) +
+                       " has been mapped to multiple times by "
```
```suggestion
                       " has already had its properties (temperature and/or density) set by "
```

src/base/OpenMCCellAverageProblem.C
```diff
@@ -1827,6 +1830,51 @@ OpenMCCellAverageProblem::cacheContainedCells()
...
+                       "fill that UNIVERSE into multiple cells.\n\nFor more information, "
+                       "please consult https://github.com/neams-th-coe/cardinal/pull/918.");
+
+          cells_already_set.insert({cc_idx, cc_instance});
+        }
+      }
+    }
+  }
+  // All MPI ranks need to wait until rank zero has finished performing the mapping check.
+  _communicator.barrier();
```
aprilnovak, line 1878: would it be effective to do the search on each MPI rank for the elements it owns? Do we have a distributed memory version of this data?

nuclearkevin: I was originally thinking about doing something like that. However, contained cells can have temperatures set by elements on multiple partitions. To catch this case, we'd need to serialize the data structure (negating any memory savings).

test/tests/neutronics/feedback/universes/lattice_in_flat/tests
```diff
@@ -1,18 +1,11 @@
...
-    capabilities = 'openmc'
-  []
-  [triso_lattice_density]
+  issues = '#1392 #1409'
+  [triso_lattice_mapping]
     type = RunException
     input = density.i
-    expect_err = "Cell 10, instance 0 has already had its density set by Cardinal"
-    requirement = "The system shall error if Cardinal tries to change the density of a given OpenMC cell more than once, since this indicates a problem with model setup."
+    expect_err = "Cell 10, instance 0 has been mapped to multiple times by Cardinal"
```
nuclearkevin:
```suggestion
    expect_err = "Cell 10, instance 0 has already had its properties (temperature and/or density) set by Cardinal!"
```

test/tests/neutronics/feedback/universes/lattice_in_flat/tests
```diff
@@ -1,18 +1,11 @@
...
-    capabilities = 'openmc'
-  []
-  [triso_lattice_density]
+  issues = '#1392 #1409'
+  [triso_lattice_mapping]
     type = RunException
     input = density.i
-    expect_err = "Cell 10, instance 0 has already had its density set by Cardinal"
-    requirement = "The system shall error if Cardinal tries to change the density of a given OpenMC cell more than once, since this indicates a problem with model setup."
+    expect_err = "Cell 10, instance 0 has already had its properties (temperature and/or density) set by Cardinal!"
```
nuclearkevin:
```suggestion
    expect_err = "Cell 10, instance 0 has already had its properties \(temperature and/or density\) set by Cardinal!"
```

gonuke:
@lewisgross1296 did you follow this? this change is in the same places we were exploring other performance improvements - are they related?

lewisgross1296:
> @lewisgross1296 did you follow this? this change is in the same places we were exploring other performance improvements - are they related?

I'm now using Cardinal on INL HPC with this PR merged. If I understand correctly, this addresses a separate issue from ours with memory required during transfers (and Kevin was slick enough to do so without the original 4x runtime required). This should allow me to use more MPI ranks per node. I did some tests with more MPI, but MOOSE initialization seemed to like 3 MPI per node, 128 OpenMP per MPI best.

Between this and #1420, I'm now getting ~12 hours per step. First attempt was 34 hours per step, so big improvements!

PR #1416, opened 2026-08-13
[WIP] Add support for delta tracking in Cardinal

### LINKED ISSUE #1363: Add support for delta tracking in Cardinal
## Reason
OpenMC will support delta tracking in the near future (see https://github.com/openmc-dev/openmc/pull/3971), which greatly improves transport performance for certainly classes of problems compared to surface tracking. This includes problems with long mean free paths relative to region subdivisions (e.g. fast reactors), and problems with many small regions (e.g. TRISO-fueled reactors). We should add support for delta tracking in Cardinal once the PR is merged into OpenMC.

## Design
Users should be able to turn delta tracking on from the input block of an  `OpenMCCellAverageProblem`. Enabling delta tracking should respect the limitations of the method in OpenMC. Tallies added by Cardinal should be restricted to collision/analog estimators.

## Impact
Performance improvements for specific problems. The largest gains will be TRISO-fueled reactors, such as HTGRs.

nuclearkevin:
This draft PR adds initial [hybrid delta tracking](https://github.com/openmc-dev/openmc/pull/4044) support to Cardinal. Several hybrid delta tracking parameters are exposed in the input parameters for `OpenMCProblemBase` to allow for performance tuning without needing to regenerate `settings.xml`, and the appropriate tally restrictions are applied if the user enables hybrid delta tracking.

This PR points Cardinal to the branch in my OpenMC fork that implements hybrid delta tracking, and so it shouldn't be merged until the delta tracking PRs get merged into OpenMC. I just wanted to put this up for people to try out if interested.

Closes #1363

### DIFF
--- a/src/base/OpenMCProblemBase.C
+++ b/src/base/OpenMCProblemBase.C
@@ -94,6 +94,8 @@ OpenMCProblemBase::validParams()
 
   params.addParam<FileName>(
       "xml_directory", "./", "The directory in which to look for OpenMC XML files.");
+  params.addParam<bool>(
+      "delta_tracking", false, "Whether delta tracking should be used for transport or not.");
 
   // Kinetics parameters.
   params.addParam<bool>("calc_kinetics_params",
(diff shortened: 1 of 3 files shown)


src/base/OpenMCProblemBase.C
```diff
@@ -95,6 +97,43 @@ OpenMCProblemBase::validParams()
...
+  params.addParam<MooseEnum>(
+      "delta_hybrid_type",
+      hybrid_tracking_type,
+      "The type of hybrid delta tracking to use. 'cross_section' refers to the "
+      "hybrid-in-cross-section scheme, which is analogous to the approach implemented in "
+      "Serpent. 'energy' refers to the hybrid-in-energy scheme as proposed by Morgan et "
+      "al.");
+  params.addRangeCheckedParam<Real>(
+      "xs_threshold",
+      0.9,
```
aprilnovak, line 114: are these defaults perhaps global variables in OpenMC that we can point to instead of repeating the value here?

nuclearkevin: A few global variables are initialized with these default values. Like all OpenMC settings variables they get overridden if the user selects to use delta tracking with different hybrid parameters, so I don't think its safe to initialize these input parameters with those globals. 

Edit: Thinking about it some more, now that OpenMC is initialized before the `OpenMCProblemBase` the defaults will be overridden if the user sets up delta tracking in the xml files. It wouldn't be safe to point to the global variables here.

nuclearkevin: I could change things on the OpenMC side so they're initialized with defined `constexpr`s, but that ends up going against how things are set up in OpenMC at present.

PR #1418, opened 2026-08-15
Add NekRS volume norm postprocessor for nek_ci verification

### LINKED ISSUE #1043: Allow computing L2 norms on NekRS mesh

nandu90:
## Summary

This PR adds support for computing volume norms of NekRS solution fields directly on the native NekRS GLL mesh.

The new `NekVolumeNorm` postprocessor is needed by the `nek_ci` verification suite to compare computed fields against analytical MOOSE functions without implementing case-specific error calculations in each NekRS UDF.

For finite `N >= 1`, the postprocessor computes

$$
\left(\int_{\Omega} \left|f-\tilde{f}\right|^N,d\Omega\right)^{1/N},
$$

where `f` is the selected NekRS field and `\tilde{f}` is an optional analytical function.

Setting `N = infinity` computes the nodal maximum norm

$$
\max_{\boldsymbol{x}_i \in \Omega}
\left|f(\boldsymbol{x}_i)-\tilde{f}(\boldsymbol{x}_i)\right|.
$$

## Changes

* Added the `NekVolumeNorm` postprocessor.
* Added support for arbitrary real norm orders `N >= 1`.
* Added support for the L-infinity norm using `N = inf` or `N = infinity`.
* Added an optional MOOSE function for evaluating the difference between a NekRS field and an analytical solution.
* Evaluated finite norms using the NekRS GLL quadrature weights on the native computational mesh.
* Added optional function weighting to `NekVolumeIntegral`.
* Added documentation for `NekVolumeNorm`.

Function-based norms and integrals are currently limited to dimensional NekRS cases.

## Testing

Added a regression test that prescribes an analytical temperature field and evaluates its error relative to a MOOSE `ParsedFunction`.

The test verifies:

* The L1 norm against the volume integral of the absolute error.
* The L2 norm against the square root of the volume integral of the squared error.

Closes #1043


src/base/NekInterface.C
```diff
@@ -925,9 +926,32 @@ dimensionalizeSideIntegral(const field::NekFieldEnum & integrand,
...
+evaluateFunctionOnMesh(const Function * f, const Real time, const int id)
+{
+  double shift = 0.0;
+  if (f)
+  {
+    // the function is given in dimensional form from MOOSE, so we need to
+    // convert to non-dimensional form before we shift the field
+    Point p(x[id], y[id], z[id]);
+    p *= scales.L_ref;
+    auto t = time * scales.t_ref;
```
nandu90: @aprilnovak will appreciate if you can cross-check the multiplication of length and time scales here - whether this is appropriate for a dimensional case

nandu90:
The PR also includes changes to `NekRSProblem.C`:

- Updated host-device mesh synchronization to correctly handle both NekRS-controlled and MOOSE-controlled mesh motion.
- Reevaluated user-defined NekRS properties after transferring scalar data from MOOSE.

Without these changes the `mv_cyl` case in NekRS test suite fails

doc/content/source/postprocessors/NekVolumeNorm.md
```diff
@@ -0,0 +1,69 @@
...
+
+!alert note
+This postprocessor does not yet support nondimensional NekRS cases.
+
+## Example Input Syntax
+
+The following code snippet evaluates the L$^2$, L$^1$, and L$^\infty$ norms
+of the NekRS temperature field relative to a function provided in the input file.
+
+```ini
```
aprilnovak: instead of writing input file text, it's better to use the `!listing` feature (no need to come back to code snippets to update manually). You can check the other documentation pages for examples

test/tests/postprocessors/nek_function_norm/tests
```diff
@@ -0,0 +1,12 @@
+[Tests]
+  design = 'NekVolumeNorm.md'
+  issues = '#1043'
+
```
aprilnovak, line 4: can you also add test cases for the invalid user inputs? e.g.:

- values that are not integers
- misspellings of infinity/inf
- values less than 1

We should also add a test case for the value of the infinity norm being computed properly

src/interfaces/NekFieldInterface.C
```diff
@@ -60,4 +69,16 @@ NekFieldInterface::NekFieldInterface(const MooseObject * moose_object,
     checkUnusedParam(parameters, "velocity_component", "not using 'field = velocity_component'");
   }
 }
+
+Real
+NekFieldInterface::evaluateShiftFunction(const Real & time, const Point & point) const
```
aprilnovak: since this code is being added here, I think it's easy to make a test case in this PR that handles the non-dimensional form as well. If you like, I can push a test case

src/base/NekInterface.C
aprilnovak: I don't think this is right - for cases that are "dimensional" we consider the NekRS problem to already be in dimensional form, so no multiplication should be done. For non-dimensional cases, the coordinates coming from MOOSE are going to be dimensional, so we want to divide by them before we send the info into NekRS. That happens in the `evaluateShiftFunction`, so I think you're double-counting it here? 

The `dimensionalizeVolumeIntegral` function that is being called right now will then do the transformations needed to go from the non-dimensional thing that NekRS computes, back into a dimensional quantity in the postprocessor.

aprilnovak:
why is the `nek_ci` submodule hash being updated here? I think it should not be?

aprilnovak:
I opted to actually not do the non-dimensional feature here - it's not critical right now and I want to get to your other PRs too!

aprilnovak:
Closing, since this was reopened as #1425

PR #1419, opened 2026-08-15
remove dead cmake lines. Refs #1296

aprilnovak:
As pointed out by @jcphill, these lines in the config/nekrs.mk are overwritten anyways by NekRS. So, these lines do nothing and can be removed.


aprilnovak:
Test failure is unrelated and is being addressed in https://github.com/neams-th-coe/nek_ci/pull/11

nandu90:
Changes look good.

PR #1420, opened 2026-08-15
try optimization for large problems to avoid unecessary subdomain-material table print

aprilnovak:
The `_subdomain_to_material` variable is only _required_ when using the skinner - for problems without the skinner, this variable is used to print a helpful table showing the mapping of subdomains to OpenMC materials. However, this can be expensive for large CSG problems - to help @lewisgross1296 with his TRISO model, this makes the computation of this variable optional (it will only be computed if `verbose = true`).

Estimated this will save @lewisgross1296 25% runtime associated with Cardinal model setup


nuclearkevin (approved):
This looks good to me. NekRS test failures are unrelated.

PR #1421, opened 2026-08-16
Loosen tolerance for tests due to compiler optimization

loganharbour:
idaholab/moose#33466 introduces compiler optimization to the entire stack (petsc, libmesh, moose). This compiler optimization can lead to small differences in basic computation (for example, fuse multiply add which can do `a + b × c` in a single step. We don't want to regold these tests because cases like running in debug will disable these optimizations, leading then to diffs in debug.

This resolves the issues seen when adding optimization in https://civet.inl.gov/job/4068831/.

I tried to be as explicit as possible about what all of the diffs were when I changed the tests. This is on you guys to see what's appropriate. Unfortunately, this is the cost of potentially making things faster. There are also options to dig into the code to make certain cases more resilent to these kinds of optimizations, but there isn't time for that at the moment.

I don't understand the dagmc tests fully, but what I suspect is that with the optimization the meshes are no longer fully equivalent.

If anyone wants to actually test this configuration themselves, they can do it easily on linux by running in one of the container images from idaholab/moose#33466. Let me know if you'd like to.

The things that need to be looked at in detail most are for the kappa_fission tallies. In each test spec where I increased abs_zero, I also put in a comment in the test spec what we could set the relative tolerance to instead. I'm not sure if it's best here to filter out noise via abs_zero or rel_err.

### DIFF
--- a/test/tests/neutronics/dagmc/cell_tallies/skin.cmp
+++ b/test/tests/neutronics/dagmc/cell_tallies/skin.cmp
@@ -1,0 +1,5 @@
+GLOBAL VARIABLES (all)
+	k                   relative 2e-2
+
+ELEMENT VARIABLES (all)
+	kappa_fission       relative 5.5e-6 floor 3e-3
--- a/test/tests/neutronics/dagmc/cell_tallies/skin_null_density.cmp
+++ b/test/tests/neutronics/dagmc/cell_tallies/skin_null_density.cmp
@@ -1,0 +1,5 @@
+GLOBAL VARIABLES (all)
+	k                   relative 2e-2
+
+ELEMENT VARIABLES (all)
+	kappa_fission       relative 5.5e-6 floor 3e-3
--- a/test/tests/neutronics/dagmc/cell_tallies/tests
+++ b/test/tests/neutronics/dagmc/cell_tallies/tests
@@ -23,6 +23,10 @@
                   "directory."
     mesh_mode = 'replicated'
     capabilities = 'dagmc'
+    # diffs with -march=x86-64-v3; set:
+    # - k rel to 2e-2
+    # - kappa_fission elemental abs_zero to 3e-3 OR rel 2e-1
+    custom_cmp = skin.cmp
   []
   [skin_null_density]
     type = Exodiff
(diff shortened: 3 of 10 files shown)


aprilnovak:
@meltawila and @nuclearkevin can you take a look at this PR and see if increasing the number of particles in these specific tests can allow us to use a tighter relative error tolerance?

loganharbour:
It's important to note that these tests are already pretty long:

```
[4.158s] [ 404MB]     DIFF test:neutronics/dagmc/implicit_comp_material/without_graveyard.no_graveyard_implicit_complement FAILED (CSVDIFF)
[4.757s] [ 403MB]     DIFF test:neutronics/dagmc/implicit_comp_material.add_implicit_complement FAILED (CSVDIFF)
[5.150s] [ 223MB]     DIFF test:neutronics/dagmc/cell_tallies.skin FAILED (EXODIFF)
[5.146s] [ 221MB]     DIFF test:neutronics/dagmc/mesh_tallies.skin FAILED (EXODIFF)
[4.630s] [ 224MB]     DIFF test:neutronics/dagmc/cell_tallies.skin_null_density FAILED (EXODIFF)
[5.194s] [ 222MB]     DIFF test:neutronics/dagmc/mesh_tallies.skin_direct FAILED (EXODIFF)
[5.188s] [ 223MB]     DIFF test:neutronics/dagmc/mesh_tallies.disjoint_bins FAILED (EXODIFF)
[41.17s] [ 151MB]     DIFF test:neutronics/mg/rr_thermal_expansion.s2_thermal_exp FAILED (CSVDIFF)
```

aprilnovak:
what is considered a long test/what should we aim for? in the grand scheme of things, these are still much faster than what we deal with for NekRS's JIT times, I think increasing the number of particles would still be smaller contributed runtime than adding a handful of additional NekRS tests

loganharbour:
> these are still much faster than what we deal with for NekRS's JIT times

which is why we can't afford to test these on moose 😆 

In general we aim for < 2 s

aprilnovak:
My concern is that if we loosen the test tolerances so much (as required if we are constraining on runtime even more than the test currently sits at), then the test becomes unhelpful because any real differences in code behavior might not be caught. This is due to OpenMC's stochastic nature - minor/floating point differences in upstream MOOSE will result in statistically-equivalent results down in OpenMC, but what is "statistically equivalent" becomes a fairly large range if we run with too few particles (and hence a loose test tolerance would be on par with actual real code diffs of concern).

Would tagging the tests as `heavy` help? (they are not heavy from Cardinal's POV, but are from MOOSE's)

loganharbour:
> Would tagging the tests as heavy help? (they are not heavy from Cardinal's POV, but are from MOOSE's)

It depends on if they're tests that are necessary for code coverage. If these things are the only thing things that cover certain lines of code, it's not the best idea

aprilnovak:
Merging due to upstream need - still working on making the tests more robust but do not want to hold others up.

PR #1422, opened 2026-08-16
MOAB Skinner updated to take arbitrary fields rather than just temperature and density. 

makeclean:
This change allows for the MOAB skinner to skin on arbitrary fields rather than just temperature and density. I have a use case for which I would like to skin on some other quantity and this felt like the most appropriate change. 

Its quite a large PR and happy to take comments. I guess I consider this a draft for now, until I've done more testing.


aprilnovak:
Thanks @makeclean! This looks neat - I will take a look this week! For the test suite to launch, can you go accept the invite to join the `cardinal-devs` organization and then force-push to update?

meltawila:
I activated the tests on Civet. Precheck failed with
```
ERROR: The following files do not contain a newline character before EOF:
	test/tests/userobjects/moab_skinner/three_fields.i

Run the "delete_trailing_whitespace.sh" script in your $MOOSE_DIR/scripts directory.
##########################################################################

[14:43:12] ERROR: Exiting with code 1
```
Tests should activate automatically going forward after you accept the cardinal-devs invitation

test/tests/neutronics/filters/energy/#mesh_energy_array.i#
```diff
@@ -0,0 +1,104 @@
+[Mesh]
```
meltawila: I think this test might have been added by mistake? it doesn't seem to belong to this PR, no?

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +214,111 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+    mins.push_back(getParam<Real>("temperature_min"));
+    maxs.push_back(getParam<Real>("temperature_max"));
+    n_bins.push_back(getParam<unsigned int>("n_temperature_bins"));
+
+    if (isParamSetByUser("density"))
+    {
+      if (names.back() == getParam<std::string>("density"))
+        mooseError("The 'temperature' and 'density' variables cannot be the same!");
+
+      checkRequiredParam(parameters(), "density_max", "binning by density");
```
meltawila: `checkRequiredParam()` uses `isParamValid()` which will always be true because `addDeprecatedParam()` now has default values for these. can we drop these defaults for density max/min, temperature max/min and n_temperature_bins, n_density_bins? right now I believe the check in checkRequiredParam() is dead code and will always be true

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +214,111 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+               "'density' parameters.");
+
+  std::vector<std::string> names;
+  std::vector<Real> mins;
+  std::vector<Real> maxs;
+  std::vector<unsigned int> n_bins;
+
+  if (!has_fields && !has_legacy)
+    paramError("fields",
+               "You must specify either 'fields' (with 'fields_min', 'fields_max', and "
```
meltawila: I'd like to suggest making `fields` optional parameter where in that case the skinning will use subdomain only skinning. The motivation is being able to skin independently of `OpenMCCellAverageProblem` which is what creates the temperature/density `AuxVariables` in a coupled run. right now if I want to generate a DAGMC model directly from a MOOSE mesh (e.g. with a future `cardinal-opt -i input.i --generate-dagmc-only` option I'm working on) I have to add dummy AuxVariable blocks just to satisfy the skinner then strip them back out when running the actual coupled problem. this is what I did in the tests:
`test/tests/neutronics/dagmc/assign_materials/generate_dagmc.i` and `‎test/tests/neutronics/dagmc/assign_materials/openmc.i` for example, but I'd like to consolidate these into a single input file with the proposed command line option after making `fields` an optional parameter

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +214,111 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+               "'n_field_bins') or the deprecated 'temperature'/'density' parameters.");
+
+  if (has_legacy)
```
meltawila: now this is completely untested. do you plan to remove the deprecated syntax at some point in the future? should we keep some tests for these or maybe remove `has_legacy` completely now and make the error messages clear about this update and what needs to be changed from an old syntax to new

test/tests/neutronics/dagmc/density_skin/only_solid.i
```diff
@@ -32,17 +32,13 @@
 [UserObjects]
   [moab]
     type = MoabSkinner
+    fields = 'temp density'
+    fields_min = '0 0'
+    fields_max = '1000 100'
+    n_field_bins = '1 4'
 
     # just one temperature bin
```
meltawila: old comment. remove.

test/tests/neutronics/dagmc/density_skin/openmc.i
```diff
@@ -78,17 +78,13 @@ drho = 50.0
 [UserObjects]
   [moab]
     type = MoabSkinner
+    fields = 'temp density'
+    fields_min = '0 ${fparse rho0 - drho}'
+    fields_max = '1000 ${fparse rho0 + 2 * drho}'
+    n_field_bins = '1 4'
 
     # just one temperature bin
```
meltawila: old comment. remove

meltawila (changes requested):
Thank you for the PR. these are a few initial comments until the test suite runs

makeclean:
> Thanks @makeclean! This looks neat - I will take a look this week! For the test suite to launch, can you go accept the invite to join the `cardinal-devs` organization and then force-push to update?

Could you re-invite - I think it may've expired can't seem to find it

src/userobjects/MoabSkinner.C
makeclean: Sounds very useful, in which case what what exactly determines what the skinner operates on? just the block ids?

aprilnovak:
@makeclean I just re-sent the invite, let me know if it doesn't show up

src/userobjects/MoabSkinner.C
meltawila: Yes, just block IDs. We can already do this with a single bin spanning the entire range (e.g. temperature_min = 0, temperature_max = 10000, n_temperature_bins = 1). the dummy temperature variable exists only to satisfy the syntax. Making fields optional just removes the need for the placeholder. 
I think nothing else should change after making the fields an optional parameter, since with no fields each bin in `groupLocalElems(_elem_bins.at(iSortBin), regions)` is just one block where the existing logic already splits each block individually

src/userobjects/MoabSkinner.C
makeclean: Personally, I prefer a clean break and removing 'legacy' as an option all together.

meltawila: @aprilnovak what do you think?

aprilnovak: I'm ok to fully remove deprecated syntax - we have to do that all the time anyways with NekRS :) We'll just want to be sure all the documentation is also updated.

test/tests/neutronics/dagmc/density_skin/openmc.i
makeclean: done

src/userobjects/MoabSkinner.C
makeclean: done

test/tests/neutronics/filters/energy/#mesh_energy_array.i#
makeclean: removed file

makeclean:
Ok, this is looking better tests are passing, however I'm seeing a failure from something I've not touched, 
```Failed Tests:
--------------------------------------------------------------------------------------------------------------
[92.30s] [1240MB]     DIFF userobjects/gap/nondimensional.nondim FAILED (EXODIFF) [FINISHED,slots=2]```
Is this a known failure or should I start digging deeper?

**Comment by meltawila** [comment]
> Ok, this is looking better tests are passing, however I'm seeing a failure from something I've not touched,

Thank you. yes, NekRS tests failures are not related to this PR and are being addressed separately

File: src/userobjects/MoabSkinner.C (hunk 4 above)
Comment (makeclean) [reply src/userobjects/MoabSkinner.C]: ok, i think that is resolved, i'll go through the docs now

**Review by nuclearkevin — CHANGES_REQUESTED** [review]
Thanks for submitting this @makeclean! I'd like to review the changes before it gets merged. Haven't had the time to yet due to the upcoming M&C submission deadline; should be able to take a look over the next week or so.

File: include/userobjects/MoabSkinner.h [hunk 7]
```diff
@@ -170,6 +188,17 @@ class MoabSkinner : public GeneralUserObject
     Reflective
   };
 
+  /**
+   * Get the bin index for a given binned field
+   * @param[in] field binned field
+   * @param[in] elem element
+   * @return field bin index
+   */
+  virtual unsigned int getFieldBin(const BinnedField & field, const Elem * const elem) const;
```
aprilnovak, line 197: could this function be moved up earlier so it's near-by the other overloaded versions of `getFieldBin`? Will just help slightly with readability

src/base/OpenMCCellAverageProblem.C
```diff
@@ -693,12 +693,43 @@ OpenMCCellAverageProblem::initialSetup()
...
       paramError("skinner", "The 'skinner' user object must be of type MoabSkinner!");
 
-    if (_skinner->hasDensitySkinning() != _specified_density_feedback)
-      mooseError(
-          "Detected inconsistent settings for density skinning and 'density_blocks'. If applying "
-          "density feedback with 'density_blocks', then you must apply density skinning in the '",
-          name,
-          "' user object (and vice versa)");
+    for (const auto & [temp_var, blocks] : _temp_vars_to_blocks)
+      if (!_skinner->binsByField(temp_var))
```
aprilnovak, line 697: I'm not sure if we need this error any more (but I might need a refresher on why it was here in the first place). In principle, I think there's nothing wrong with someone who applies temperature feedback, but doesn't want to skin based on temperature - they won't capture the continuous field of temperature very well, but it's not an error per se. Maybe we change this to a warning?

Same comment below for the other two errors that are thrown

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +204,68 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
   return _fe_problem.getAuxiliarySystem().getFieldVariable<Real>(0, name).number();
 }
 
+void
+MoabSkinner::readFieldParameters()
+{
+  // used to skin using only block ids
+  if (!isParamSetByUser("fields"))
+  {
+    _console << "Skinning using block ID's only..." << std::endl;
```
aprilnovak, line 213:
```suggestion
    _console << "Skinning using block IDs only..." << std::endl;
```

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +204,68 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+  std::vector<std::string> vnames = {"fields", "fields_min", "fields_max", "n_field_bins"};
+  std::array<std::size_t, 4> sizes = {names.size(), mins.size(), maxs.size(), n_bins.size()};
+  for (const auto i : index_range(sizes))
+    if (sizes[i] != sizes[0])
+      paramError(vnames[i],
+                 "'" + vnames[i] + "' (length " + Moose::stringify(sizes[i]) +
+                     ") must be the same length as 'fields' (length " + Moose::stringify(sizes[0]) +
+                     ")");
+
+  if (names.empty())
```
aprilnovak, line 235: Does this prevent us from skinning by subdomain only? I think @meltawila has been using only subdomain skinning by declaring a dummy skinning field - if not in this PR, we can make an issue and update in a future PR since I don't believe this PR is introducing any kind of constriction of functionality, just modernizing what we have. 

Example of dummy variable for skinning (choose 1 bin with a huge range between min and max so that everything falls into 1 bin):

```
[UserObjects]
  [s]
    type = MoabSkinner
    fields = 'temp'
    fields_min = 0
    fields_max = 10000
    fields_bins = 1
  []
[]
```

Proposed for a later PR:

```
[UserObjects]
  [s]
    type = MoabSkinner
    # skins by subdomain since nothing else is present
  []
[]
```

What do you think @meltawila @makeclean ? I think the latter is nicer syntax, but only if we think we'd _always_ want to have subdomain skinning enabled. Are there use cases where we do want to be able to prevent subdomain skinning?

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +204,68 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+    seen_names.insert(names[i]);
+
+    if (maxs[i] <= mins[i])
+      paramError("fields_max",
+                 "The maximum bound for field '",
+                 names[i],
+                 "' (" + Moose::stringify(maxs[i]) + ") must be greater than its minimum bound (" +
+                     Moose::stringify(mins[i]) + ").");
+
+    if (n_bins[i] == 0)
```
aprilnovak, line 253: Since the input type is `unsigned int`, any negative values will get cast into `unsigned int`. I checked locally, and it seems like MOOSE will throw its own error if inputting a negative number (you used to have to be more careful with casting negatives into unsigned), so this check for zero is sufficient. Just noting here for posterity because I was initially unsure.

src/userobjects/MoabSkinner.C
```diff
@@ -738,13 +770,14 @@ MoabSkinner::sortElemsByResults()
...
-      n_temp_hits[iBin] += 1;
+      std::vector<unsigned int> fbins(_fields.size());
+      for (const auto i : index_range(_fields))
+      {
+        fbins[i] = getFieldBin(i, elem);
+        n_field_hits[i][fbins[i]] += 1;
+      }
 
-      _elem_bins.at(getBin(iBin, iDenBin, iMat)).insert(elem->id());
+      _elem_bins.at(total_bin(fbins, iMat)).insert(elem->id());
```
aprilnovak, line 780: MOOSE convention is use to camelcase; `total_bin` --> `totalBin`

test/tests/userobjects/moab_skinner/errors/tests
```diff
@@ -5,54 +5,54 @@
...
     input = skinner.i
-    cli_args = 'UserObjects/moab/temperature_max=450.0'
+    cli_args = 'UserObjects/moab/fields_max="450 2000"'
     mesh_mode = 'replicated'
-    expect_err = "Variable 'temp' has value above maximum range of bins. Please increase 'temperature_max'."
+    expect_err = "Variable 'temp' has value above maximum range of bins. Please increase the maximum bound for this field."
     requirement = "The system shall error if the temperature is above the maximum bin bound"
     capabilities = 'dagmc'
   []
   [too_low_density]
```
aprilnovak, line 41: we can probably delete the `too_low_density` and `too_high_density` tests since the same functionality is checked in the corresponding temperature tests (since now all fields are treated the same).

test/tests/userobjects/moab_skinner/errors/tests
```diff
@@ -68,16 +68,16 @@
...
     input = skinner.i
-    cli_args = 'UserObjects/moab/temperature=temperature'
-    expect_err = "Cannot find auxiliary variable 'temperature'!"
+    cli_args = 'UserObjects/moab/fields="nope rho"'
+    expect_err = "Cannot find auxiliary variable 'nope'!"
     mesh_mode = 'replicated'
     requirement = "The system shall error if the specified temperature auxiliary variable cannot be found"
     capabilities = 'dagmc'
   []
   [no_aux_density]
```
aprilnovak, line 77: can probably delete this test since the corresponding `no_aux_temp` test covers the same functionality

src/auxkernels/SkinnedBins.C
```diff
@@ -28,36 +31,44 @@ SkinnedBins::validParams()
...
-  _skin_by(getParam<MooseEnum>("skin_by"))
+SkinnedBins::SkinnedBins(const InputParameters & parameters)
+  : AuxKernel(parameters), _skin_by(getParam<std::string>("skin_by"))
 {
   const UserObjectBase & base = getUserObjectBase("skinner");
   _skinner = dynamic_cast<const MoabSkinner *>(&base);
   if (!_skinner)
     paramError("skinner", "This userobject must be of type MoabSkinner!");
+
+  if (_skin_by != "all" && _skin_by != "subdomain")
```
aprilnovak, line 50: I think this one might be missing a test?

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +204,68 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+
+  const auto & names = getParam<std::vector<std::string>>("fields");
+  const auto & mins = getParam<std::vector<Real>>("fields_min");
+  const auto & maxs = getParam<std::vector<Real>>("fields_max");
+  const auto & n_bins = getParam<std::vector<unsigned int>>("n_field_bins");
+
+  std::vector<std::string> vnames = {"fields", "fields_min", "fields_max", "n_field_bins"};
+  std::array<std::size_t, 4> sizes = {names.size(), mins.size(), maxs.size(), n_bins.size()};
+  for (const auto i : index_range(sizes))
+    if (sizes[i] != sizes[0])
```
aprilnovak, line 229: is there a test for this error?

src/userobjects/MoabSkinner.C
```diff
@@ -257,6 +204,68 @@ MoabSkinner::getAuxiliaryVariableNumber(const std::string & name,
...
+      paramError(vnames[i],
+                 "'" + vnames[i] + "' (length " + Moose::stringify(sizes[i]) +
+                     ") must be the same length as 'fields' (length " + Moose::stringify(sizes[0]) +
+                     ")");
+
+  if (names.empty())
+    paramError("fields", "You must specify at least one field to bin by.");
```
aprilnovak, line 236: is there a test for an empty `fields`, e.g. `fields = ""`?

src/userobjects/MoabSkinner.C
```diff
@@ -147,6 +135,8 @@ MoabSkinner::MoabSkinner(const InputParameters & parameters)
   if (!getMooseMesh().getMesh().is_serial())
     mooseError("MoabSkinner does not yet support distributed meshes!");
```
aprilnovak, line 137: In the `august2026.md` news, it says that using the old syntax will error, but I don't see where that would happen here. I think we'd keep the deprecated parameters but just immediately error here if anyone specifies any of `temperature`, `temperature_min/max`, `n_temperature_bins`, etc. for density as well.

aprilnovak:
Thanks @makeclean, this is cool and much-needed generalization of this. Everything looks good to me, just a few requests for additional RunException-type tests.

src/userobjects/MoabSkinner.C
meltawila: @makeclean removed the restriction you mention following a comment above in [f598fd](https://github.com/neams-th-coe/cardinal/pull/1422/commits/f598fd30fb4b9adcce1a921433560eb3adcfe533) which now allows skinning by subdomain only without specifying any fields. But if the skinner is coupled with an OpenMCCellAverageProblem you might run into the error:

```
Detected inconsistent settings for temperature feedback and the 'moab' skinner. Temperature feedback is applied with variable 'temp' (via 'temperature_blocks'), which requires the skinner to bin by that variable. 
Please add 'temp' to the skinner's 'fields' parameter.
```

For what I had in mind this is enough. since that subdomain only skinning is really needed when it's a standalone skinner

src/base/OpenMCCellAverageProblem.C
makeclean: If you don't skin by temperature, I guess you'll get the block averaged (i.e. cell wise temperature) gradients will be ignored, which is fine; I think I agree

src/auxkernels/SkinnedBins.C
makeclean: Im not actually sure this is possible, you can only skin by "all", "subdomain", and one of the field names, I've no idea therefore how you could be all but not subdomain?

src/userobjects/MoabSkinner.C
makeclean: just adding a test now

makeclean:
Is there a policy on squashing, I think this is all done now, should I go ahead and squash the commits into one?

nuclearkevin:
We normally only squash if a PR made multiple commits adding and removing large Exodus gold files, as those can start inflating the size of the repo. If it's necessary we'll squash and merge, no need to worry about it in the PR itself.

doc/content/news/august2026.md
```diff
@@ -0,0 +1,13 @@
+# August 2026 News
```
nuclearkevin, line 1:
```suggestion
# September 2026 News
```

The file will also be renamed. This one is on us for the prolonged review process, sorry for the bother

nuclearkevin (changes requested):
Thanks for generalizing the skinner @makeclean! You'll need to link the August/September news in [news.md](https://github.com/neams-th-coe/cardinal/blob/devel/doc/content/news.md) for the entry to show up in our documentation build. Otherwise this looks good to me!

doc/content/news/august2026.md
aprilnovak: meh, I honestly think it's fine to have listed as August to save some hassle

src/auxkernels/SkinnedBins.C
aprilnovak: Do you mean that the `paramError("skin_by"`... could never be reached?

meltawila (approved):
My comments have been addressed. thank you

nuclearkevin (approved):
LGTM! Thanks @makeclean

PR #1425, opened 2026-08-18
Nek function norm

aprilnovak:
Refs #1043 (does not close, because this PR does not add the non-dimensional support)

Same as #1418, but I found it much easier to just start fresh than do the big rebase.


nandu90:
This looks good. Merging this to allow subsequent PRs that build on this.

PR #1427, opened 2026-08-19
Modify NekVolumeIntegral to allow integration of arbitrary functions 

nandu90:
## Summary

This PR extends `NekVolumeIntegral` to integrate a MOOSE function evaluated directly on the native NekRS GLL mesh.

`NekVolumeIntegral` now requires exactly one of:

* `field`, which preserves the existing field-integration behavior; or
* `function`, which evaluates and integrates the supplied MOOSE function.

When a function is supplied, the internal NekRS field is set to `unity`. This allows the existing volume-integral dimensionalization to correctly apply the reference-volume scaling in nondimensional cases.

## Testing

* Added dimensional function-integration coverage.
* Added nondimensional function-integration coverage.
* Compared function integrals against equivalent fields transferred to the NekRS mesh.
* Added error tests for specifying both `field` and `function`.
* Added an error test for specifying neither `field` nor `function`.


aprilnovak:
What do you think about instead requiring `field` to be present, but making `function` optional? We could always interpret the integrand as `field - function`, and then add a field option of `zero`? I think this would be a bit more general and if we added it to the base class, then it would be an option within all of the postprocessors in Cardinal (so, not only would a user be able to integrate a function over a volume, but over a sideset, etc.). If you agree, I can add those commits on top of your branch here.

nandu90:
> What do you think about instead requiring `field` to be present, but making `function` optional? We could always interpret the integrand as `field - function`, and then add a field option of `zero`? I think this would be a bit more general and if we added it to the base class, then it would be an option within all of the postprocessors in Cardinal (so, not only would a user be able to integrate a function over a volume, but over a sideset, etc.). If you agree, I can add those commits on top of your branch here.

If a field option of `zero` can be added sure we can go that route. We will have to update the description in `.md` file to make the user cognizant of the operation `field - function`. Feel free to add relevant commits.

aprilnovak:
@nandu90, after #1431 is merged, this PR should be closed, and I think it will not be necessary to move anything over since the functionality in #1431 should fully replicate the behavior here. I confirmed that this is true by doing the same integrals in the tests in this PR; they match with the new implementation in #1431.

nandu90:
Addressed in #1431

PR #1428, opened 2026-08-19
Fix NekRS moving-mesh synchronization and restore deformation tests

nandu90:
## Summary

This PR fixes mesh synchronization for NekRS moving-mesh cases and updates the deformation regression tests for NekRS v26.

## Changes

* Copy NekRS-controlled mesh coordinates and geometric factors back to the host after each solve.
* Distinguish between the two mesh-deformation modes:

  * With the `user` geometry solver, MOOSE provides the full volume deformation and coordinates are copied from host to device.
  * With a blending geometry solver, MOOSE provides boundary velocities and NekRS moves the volume mesh, so coordinates and geometric factors are copied back to the host.

* Update and re-enable the following deformation tests for NekRS v26:

  * `deformation/nek_standalone.volume`
  * `deformation/nek_standalone.boundary`
  * `deformation/vol-areas.deformed_areas`
  * `deformation/mesh-velocity-areas.boundary`
  * `deformation/mesh-velocity-areas.volume`
* Update the moving-mesh UDF, OUDF, parameter files, transfer syntax, and gold values to the NekRS v26 interfaces.

The `mesh-velocity-areas` tests provide separate coverage for boundary and volume mesh mirrors using the NekRS blending solver. 

A direct check of the synchronized host coordinates is left as a follow-up pending function support in `NekVolumeIntegral` (#1427)

## Additional Changes

* Evaluate user-defined NekRS properties only when at least one field or scalar transfer sends data to NekRS. Prevents double call of `evaluateProperties` for cases without data transfer to NekRS.


test/tests/deformation/mesh-velocity-areas/gold/pipe_out_nek0.csv
```diff
@@ -1,2 +1,2 @@
 time,moose_diff,nek_diff
-0.1,0.0039906935058518,0.0037543045648576
+0.1,0.0039924353051823,0.0039124479784345
```
aprilnovak: If the test is being "restored", why is the value different here?

test/tests/deformation/vol-areas/nek.i
```diff
@@ -44,40 +44,40 @@
 
 [Postprocessors]
   [nekbdry_icar1]
-    type = AreaPostprocessor
```
aprilnovak, line 47: why were these types changed? I think the purpose of having them as `AreaPostprocessor` was to check that the mesh is actually seen to be moving on the Cardinal side (which is more robust than solely checking the NekRS internal mesh, because checking on the Cardinal side confirms both that NekRS internal is being updated _and_ that data has been passed back/updated on the Cardinal side as well).

Can these be reverted back to `AreaPostprocessor`s?

test/tests/deformation/mesh-velocity-areas/gold/pipe_out_nek0.csv
nandu90: `nek_diff` changes because the fix now copies the NekRS-controlled mesh coordinates and updated geometric factors from the device to the host after each solve ([`[NekRSProblem.C](https://github.com/neams-th-coe/cardinal/pull/1428/changes#diff-ab3c1d3093c1a56724c00cee8a04d8dda2cff5dfd779ddd1f912dc78576b6a66R540-R545)`](https://github.com/neams-th-coe/cardinal/pull/1428/changes#diff-ab3c1d3093c1a56724c00cee8a04d8dda2cff5dfd779ddd1f912dc78576b6a66R540-R545)). Previously, `NekSideIntegral` used host-side geometric factors that were not refreshed after NekRS moved the mesh. It now evaluates the area using the current deformed geometry.

test/tests/deformation/vol-areas/nek.i
nandu90: This is reverted

src/base/NekRSProblem.C
```diff
@@ -797,7 +834,23 @@ NekRSProblem::copyHostToDevice()
   for (const auto & slot : _usrwrk_slots)
     copyIndividualScratchSlot(slot);
 
-  if (nekrs::hasMovingMesh())
+  bool moose_controls_mesh = false;
```
aprilnovak: We can remove some code duplication if we make `moose_controls_mesh` a member variable so we don't need to recompute it. I will push a commit with this change

nandu90:
This is ready to be merged. 

Tested all tests in neams-th-coe/nek_ci#11 locally

src/base/NekRSProblem.C
```diff
@@ -521,6 +522,22 @@ NekRSProblem::externalSolve()
...
+  // blending solver, MOOSE only provides boundary velocities and NekRS moves the volume mesh.
+  if (nekrs::hasUserMeshSolver())
+    for (const auto & t : _field_transfers)
+      if (dynamic_cast<NekMeshDeformation *>(t))
+      {
+        _moose_controls_mesh = true;
+        break;
+      }
+
+  if (nekrs::hasMovingMesh() && !_moose_controls_mesh)
```
aprilnovak: I guess I am confused on this line - why would we want the `[Mesh]` to only be updated for certain mesh solvers? Wouldn't we always want to be updating the mesh on the Cardinal side if it is moving?

nandu90: The `mesh-velocity-areas` test was failing without this guard

aprilnovak: failing in what kind of way? wrong values or aborting?

nandu90: I think you might be right. I cannot reproduce the error. I think the second if condition can be safely removed. Re-running relevant tests to confirm

nandu90: I have removed moose_controls_mesh flag. It was unnecessary. All deformation tests and nek_ci tests pass locally for me with the update

aprilnovak: nice, thanks for checking that! I will merge once the test suite runs through. Failures are anticipated in several areas but are being addressed separately in #1434 #1433

aprilnovak:
Test failure is unrelated and is being addressed in #1433

Issue #1426, opened 2026-08-18
Harden MoabSkinner / DAGMC tests against differences in floating point optimizations


nuclearkevin:
## Reason
PETSc/libMesh/MOOSE recently added compiler optimizations to improve performance at the cost of changing floating point operations. This causes several `MoabSkinner`/`DAGMC` tests to fail. #1421 loosens the relative/absolute tolerances on these tests, which succeeds in getting tests to pass at the cost of potentially compromising testing goals due to OpenMC's stochastic nature. Regolds aren't an option as these compiler optimizations are removed on debug builds, so tests won't pass on those.

## Design
As an alternative to #1421, tests should be redesigned to harden them against differences in floating point math implementations / instruction sets (e.g. FMA and the like) without compromising on testing goals. As an example, `tests/neutronics/mg/rr_thermal_expansion` checks to make sure deformation works with the random ray solver. Instead of storing the value of `k_eff`, we can instead check to make sure that `k_eff` is decreasing over Picard iterations and store the result of the boolean comparison `k_eff < k_eff_initial` on the final Picard iteration. This accomplishes the testing goal, and is substantially more resilient to stochastic error from OpenMC and differences in compiler optimizations.

## Impact
Tests pass on both debug and optimized builds while maintaining testing goals.

loganharbour:
> As an alternative to [#1421](https://github.com/neams-th-coe/cardinal/pull/1421), tests should be redesigned to harden them against differences in floating point math implementations / instruction sets

Keep in mind that if this isn't a solution in the near-term, we'll need to either allow cardinal to fail on MOOSE or get something temporary in place.

aprilnovak:
We are working on it today! So, the fixes will be near-term

PR #1429, opened 2026-08-19
Harden various `MoabSkinner` tests using 3 sigma comparisons

nuclearkevin:
Ref. #1426


aprilnovak:
Thanks @nuclearkevin this looks great! And just to confirm, in the off chance there's a bug in the testing methodology itself - can you perturb/introduce a bug locally that invalidates the test? (for instance, these tests could erroneously pass if `in_three_sigma`'s logic accidentally assigned 1 to all points in space for some reason)

nuclearkevin:
To test this, I doubled the density in one half of the slab (for the CSG reference). `k_in_three_sigma` becomes zero as the values of `k_eff` no longer match to $3\sigma$, and some parts of the mesh tally start to mismatch as well resulting in values of `in_three_sigma` that are zero (though, some parts of the domain still match).

image

aprilnovak:
Thanks for confirming! LGTM!

The test failure is unrelated and is being addressed in https://github.com/neams-th-coe/nek_ci/pull/15

Issue #1046, opened 2025-02-04
Add `function` and `function_difference` as fields


aprilnovak:
## Reason
To support many of the error metrics in NekRS's test suite, we need to be able to take intergrals/max/min/norms of both functions and NekRS fields relative to functions.

## Design
Add `function` and `function_difference` as fields which are then supported by the postprocessors in Cardinal.

## Impact
New feature.

PR #1430, opened 2026-08-19
prevent function usage when not enabled.

aprilnovak:
Not exactly elegant, but I realized later that by adding `function` to the `NekFieldInterface`, that all the derived classes will show it on the documentation, but the parameter is not actually used yet. Planning to add it, so this is just a placeholder to ensure correct code behavior in the interim while those features are added.

 Refs #1046


nuclearkevin (approved):
LGTM as a stop-gap measure.

nuclearkevin:
The test failure is unrelated and is being addressed in https://github.com/neams-th-coe/nek_ci/pull/15

PR #1431, opened 2026-08-19
Add support for shifting by a function in the NekFieldPostprocessor-derived obejcts

aprilnovak:
Refs #1046

Adds function shift support for `NekPointValue`, `NekVolumeExtremeValue`, `NekVolumeIntegral`, and `NekVolumeAverage`. Test additions check by hand. To support @nandu90's #1427.


test/tests/postprocessors/nek_volume_integral/nek.i
```diff
@@ -102,6 +114,12 @@
     type = NekVolumeIntegral
     field = scalar01
   []
+  [s01_intf]
+    type = NekVolumeIntegral
+    field = scalar01
+    function = f
```
nandu90, line 125: What if i need to compute the integral of only the function? Can i specify field = zero?

aprilnovak: here is what I was thinking:

```
field = unity
function = '-<the_actual_function> + 1'
```

you could add 1 to the function you actually care about, so that the integrand becomes `1 - (-function + 1) = function`. But if this is too confusing, I could add a field of zero (the thought of just manipulating the function didn't occur to me until later).

aprilnovak:
Regarding integrals of time-dependent functions, I just wanted the test case to be as general as possible - the function does not need to include time if it's not needed.

nandu90:
> Regarding integrals of time-dependent functions, I just wanted the test case to be as general as possible - the function does not need to include time if it's not needed.

Yes, I deleted that comment realizing its good to have a general capability available. Sounds good!

test/tests/postprocessors/nek_volume_integral/nek.i
nandu90: > here is what I was thinking:
> 
> ```
> field = unity
> function = '-<the_actual_function> + 1'
> ```
> 
> you could add 1 to the function you actually care about, so that the integrand becomes `1 - (-function + 1) = function`. But if this is too confusing, I could add a field of zero (the thought of just manipulating the function didn't occur to me until later).
> 

I think adding a field = zero is a more elegant solution. We have a field = unity as well. This will add more flexibility.

aprilnovak: ok, I will re-open that PR and have you review it. It's not necessary for this one here, but we can combine them together for nek_ci needs

nandu90: Merged #1432. We should modify the test case to use:

field = zero
function = '-<the_actual_function>'

aprilnovak: is this referring to a test case in this PR or in nek_ci (or somewhere else)?

nandu90: A test using zero field is added in #1432 . So we can skip adding a similar test here. 

What i was suggesting was to change the test here: https://github.com/neams-th-coe/cardinal/pull/1431/changes#diff-dd73dc844e70fd26ac1c84177e8f2429201d9d0aae5089c5779bbd0109320a4fR106-R107

to :

field = zero
function = - 1.0

so that unity_intf == unity_int -> as a demonstration of computing volume using NekVolumeIntegral using function instead of field and also to test field = zero.

However, i leave this as a suggestion. The capability is already tested implicitly through other tests.

aprilnovak: sure, I can add that!

aprilnovak: done

aprilnovak:
All tests pass locally; the failures can be explained by:

- unit test failure is caused by changes to MOOSE's compiler optimizations; it is being addressed in #1433 
- initializing coverage for NekRS is being addressed on the INL side and is also due to MOOSE environment changes
- small mesh diffs in the `userobjects/gap/nondimensional` test have been fixed in #1435

PR #1432, opened 2026-08-20
add zero as a field

aprilnovak:
To support #1427; Refs #1046

When shifting a field by a function prior to evaluating with a postprocessor, we can use the same system to simply evaluate the function in that postprocessor directly. In other words, we are updating NekRS's postprocessors to take various operations of (field - function), instead of just (field). But, we can also allow someone to operate a function on a NekRS mesh if we allow the field to be zero.


aprilnovak:
Actually, I'm going to close this - it's not necessary, because you could always use `unity` and then just add a +1 to the function which is being subtracted off.

aprilnovak:
reopening on request by @nandu90

nandu90:
Looks good. Merging to enable subsequent merge.

aprilnovak:
@nandu90 you're not supposed to merge if tests are failing, we have to wait until they pass to do that

nandu90:
Sorry to jump to merging. Does this need to be reverted?

PR #1433, opened 2026-08-25
fix fp in gtest. Refs #1426

aprilnovak:


aprilnovak:
Test failures are unrelated and being addressed in other PRs (e.g. #1434)

unit/src/SymmetryPointGeneratorTest.C
```diff
@@ -54,7 +54,7 @@ TEST_F(SymmetryPointGeneratorTest, reflect_xy_plane)
   Point pt3(-1.0, 2.0, 2.0);
   Point pt3_r = _spg2->transformPoint(pt3);
   EXPECT_DOUBLE_EQ(pt3_r(0), 2.0);
-  EXPECT_DOUBLE_EQ(pt3_r(1), -1.0);
+  EXPECT_NEAR(pt3_r(1), -1.0, 1e-8);
```
nandu90: Does it make sense for this to be machine precision (1e-16, instead of 1e-8)?

loganharbour: The current failure is at 5e-15. I'd say something like 1e-12 is sufficient. 1e-8 is definitely too loose

aprilnovak: ok, will update to 1e-12

loganharbour (approved):

PR #1435, opened 2026-08-25
Make userobject test more robust against small mesh differences. 

aprilnovak:
This updates the `userobjects/gap/nondimensional` test to make more robust in light of compiler optimizations in MOOSE. FYI @loganharbour 

The mesh that was used in this test was the same one used in `userobjects/side/nondimensional`, which we found necessary to update in #1340. Basically, this mesh had a few nodes that fell very close to the subchannel boundaries, so small differences in the mesh generation could lead to noticeable changes in the test results. I am repeating below the explanation we use for updating it's companion test back in #1340 for context:

> re-golded the userobjects/side/nondimensional test: this test was showing failures on CIVET, but not locally. After further testing, I believe this was due to a small floating point indeterminism in which bin of the userobject an element falls into, since the elements in the original mesh were nearly perfectly straddling one of the boundaries of the userobject edges. I recreated the mesh to have better quality and re-golded, but still confirmed that the values look identical to the prior test and still match the dimensional version for which the purpose of this test exists.

I re-golded this test, using a better NekRS mesh, and again compared to the dimensional case (values match perfectly to my eye). I also removed the integral tests from the simulation, since we are integrating over very small areas those values were on order of 1e-8 and not a good test candidate. We still test for averages, which exercises the integral functionality in the numerator, so I don't find this to be a loss of coverage.

The test was modified to just check that the averaged UO values are sufficiently close to the original field - these are within 20% which is quite reasonable expecting that the average will not equal the pointwise value anyways.


aprilnovak:
Test failure is unrelated and being addressed separately

test/tests/userobjects/gap/nondimensional/subchannel.i
```diff
@@ -26,24 +26,94 @@
...
+    pp_names = 'max_T_diff max_T min_T'
+  []
+  [pass_T]
+    type = ParsedPostprocessor
+    expression = 'if (fraction_T < 0.2, 1, 0)'
+    pp_names = 'fraction_T'
+  []
+  [fraction_v]
+    type = ParsedPostprocessor
+    expression = 'max_v_diff/(max_v-min_v)'
```
MohamedElkamash: Why are we dividing the max relative error by the range?

aprilnovak: I was trying to just get some feel for how big the differences were, relative to the scale in the field. e.g. if the values differ by 3.0 (absolute) but the field ranges from 1 to 1000, that's pretty small. I'm open to changes here, this test just can't be a traditional re-gold due to the MOOSE compiler optimization changes which lead to small floating point differences in which elements get assigned to which bins of the postprocessor.

MohamedElkamash: I agree that we need something relative, but I thought it would be abs(velocity-avg_v)/abs(velocity) or abs(velocity-avg_v)/(max_v-min_v) but we currently have abs(velocity-avg_v)/[abs(velocity)(max_v-min_v)]

test/tests/userobjects/gap/nondimensional/subchannel.i
```diff
@@ -26,24 +26,94 @@
...
+  [diff_T]
+    type = ParsedAux
+    variable= diff_T
+    expression = 'abs(temp-avg_T)/abs(temp)'
+    coupled_variables = 'temp avg_T'
+  []
+  [diff_v]
+    type = ParsedAux
+    variable= diff_v
+    expression = 'abs(velocity-avg_v)/abs(velocity)'
```
MohamedElkamash, line 57: just to clarify the previous comment, you already divided by abs(velocity) here so diff_v is already relative not absolute

aprilnovak: Oh I see - yeah I will change that

aprilnovak: done

MohamedElkamash:
My only comment is that we compute and pass avg_p but it is not used in the test, for the sake of time we can merge as it doesn't affect the results of the test.

PR #1436, opened 2026-08-26
Adds mooseWarning for when a moving OpenMC problem is detected, but the moose mesh is fixed.

### LINKED ISSUE #1335: Add a warning when a fixed mesh is being coupled to a moving OpenMC problem
## Reason
We warn users when they try to use a displaced MOOSE problem with a stationary OpenMC geometry:
```
  if (_use_displaced && !_using_skinner && !hasCellTransform())
    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner' or an "
                 "OpenMCCellTransform user object (both of which move the OpenMC geometry). The "
                 "[Mesh] will move, but the underlying OpenMC geometry will remain unchanged. "
                 "Unexpected behavior may occur.");
```
We should also warn against the opposite case where a moving OpenMC problem is detected, but the moose mesh is fixed

## Design
Add a mooseWarning message

## Impact
Warn users against unexpected behavior

cvoorhees15:
closes #1335

### DIFF
--- a/src/base/OpenMCCellAverageProblem.C
+++ b/src/base/OpenMCCellAverageProblem.C
@@ -568,19 +568,23 @@ OpenMCCellAverageProblem::initialSetup()
   }
 
   getOpenMCUserObjects();
+  if (_needs_to_map_cells)
+  {
+    if (_use_displaced && !_using_skinner && !movesOpenMCGeometry())
+      mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner', an "
+                   "OpenMCCellTransform user object, or a geometry changing criticality search "
+                   "(all of which move the OpenMC geometry). The [Mesh] will move, but the "
+                   "underlying OpenMC geometry will remain unchanged. Unexpected behavior may "
+                   "occur.");
+    else if (!_use_displaced && movesOpenMCGeometry())
+      mooseWarning("The OpenMC geometry is being moved (via an OpenMCCellTransform user object "
+                   "or a geometry changing criticality search), but your problem has a fixed "
+                   "mesh. The underlying OpenMC geometry will change, but the [Mesh] will not "
+                   "move. Unexpected behavior may occur.");
+  }
 
-  if (_use_displaced && !_using_skinner && !hasCellTransform())
-    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner' or an "
-                 "OpenMCCellTransform user object (both of which move the OpenMC geometry). The "
-                 "[Mesh] will move, but the underlying OpenMC geometry will remain unchanged. "
-                 "Unexpected behavior may occur.");
-
-  // Coupling re-initialization should be triggered if we have cell transforms which can happen
-  // even if the mesh isn't moving or adaptive.
-  _need_to_reinit_coupling |= hasCellTransform();
-  // The criticality search may modify the geometry.
-  if (_criticality_search)
-    _need_to_reinit_coupling |= _criticality_search->changingGeometry();
+  // Coupling re-initialization should be triggered if the OpenMC geometry is being moved
+  _need_to_reinit_coupling |= movesOpenMCGeometry();
 
   if (!_needs_to_map_cells)
     checkUnusedParam(parameters(),
--- a/test/tests/userobjects/openmc_cell_transform/tests
+++ b/test/tests/userobjects/openmc_cell_transform/tests
@@ -68,4 +68,12 @@
     expect_err = "UserObjects/translate_cells/vector_value: Provide exactly 3 values/postprocessors: 'dx dy dz' in mesh units for translationtransform or 'φ, θ, ψ' in degrees for rotation transform."
     capabilities = 'openmc'
   []
+  [mismatching_geometry_and_mesh]
+    type = RunException
+    input = cells_shadow.i
+    cli_args = '--error'
+    expect_err = "The OpenMC geometry is being moved \(via an OpenMCCellTransform user object or a geometry changing criticality search\), but your problem has a fixed mesh. The underlying OpenMC geometry will change, but the \[Mesh\] will not move. Unexpected behavior may occur."
+    requirement = 'The system shall warn the user when the OpenMC geometry is moved by a cell transform while the mesh remains stationary.'
+    capabilities = 'openmc'
+  []
 []
--- a/include/base/OpenMCProblemBase.h
+++ b/include/base/OpenMCProblemBase.h
@@ -437,6 +437,12 @@ protected:
   /// Whether OpenMC cell transforms are being applied to the geometry
   bool hasCellTransform() const;
 
+  /**
+   * Whether the OpenMC geometry is being moved by a cell transform
+   * or a geometry changing criticality search
+   */
+  bool movesOpenMCGeometry() const;
+
   /// Ensure that the IDs of OpenMC objects in UserObjects don't clash
   void checkOpenMCUserObjectIDs() const;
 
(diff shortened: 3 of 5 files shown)


cvoorhees15:
I wasn't sure whether the `_using_skinner` flag was supposed to be included in this case... Mostly because I'm not familiar with what a "skinner" is. Feel free to explain or point me to relevant docs, thanks!

meltawila:
Thank you for the PR. the "skinner" refers to [MoabSkinner](https://cardinal.cels.anl.gov/source/userobjects/MoabSkinner.html) UserObject in Cardinal. it's another example where we can encounter a displaced problem (for example transferring displacements to Cardinal app following thermomechanics solve and update OpenMC geometry with these). so that's what I had in mind when I opened this issue. However thinking more about this now, I think leaving this out of the new condition is correct because the skinning operates on the [Mesh] block and I don't think there's a scenario where the skinner can make the updated OpenMC geometry mismatch the mapping with the [Mesh].

We have a few tests that should now show this new warning such as `test/tests/userobjects/openmc_cell_transform/cells_shadow.i`. However something I'm just noticing is that criticality search objects derive from `OpenMCCellTransformBase` not `OpenMCCellTransform` and might not be captured by `hasCellTransform()`. I'll want to make sure that rotation criticality search tests also show this warning when they modify the OpenMC geometry while the MOOSE mesh remains unchanged. I'll take a deeper look tomorrow.

NekRS tests failures are unrelated and are being handled separately

aprilnovak:
Hi @cvoorhees15, I've just sent you an invite to join `cardinal-devs` - once you join, the test suite here will automatically launch upon further updates (Mahmoud manually triggered it for the first run).

src/base/OpenMCCellAverageProblem.C
```diff
@@ -574,6 +574,11 @@ OpenMCCellAverageProblem::initialSetup()
                  "OpenMCCellTransform user object (both of which move the OpenMC geometry). The "
                  "[Mesh] will move, but the underlying OpenMC geometry will remain unchanged. "
                  "Unexpected behavior may occur.");
+  else if (!_use_displaced && hasCellTransform())
```
meltawila: I think this should be `else if (!_use_displaced && (hasCellTransform() || (_criticality_search && _criticality_search->changingGeometry())))` to show the warning when a criticality search object uses `OpenMCCellTransformBase` and changes the geometry. feel free to rewrite this if you think there's a better way to catch this scenario

src/base/OpenMCCellAverageProblem.C
meltawila: the corresponding opposite case should be added here

meltawila (changes requested):
the below is my main concern since hasCellTransform() is just:
```
bool
OpenMCProblemBase::hasCellTransform() const
{
  return !_cell_transform_uos.empty();
}
```
you could maybe add a new `bool OpenMCProblemBase::movesOpenMCGeometry() const` that combines this with the criticality search scenario I was writing inline below

src/base/OpenMCCellAverageProblem.C
cvoorhees15: Should this now be collapsed to just:

`_need_to_reinit_coupling |= movesOpenMCGeometry();`

src/base/OpenMCCellAverageProblem.C
```diff
@@ -569,12 +569,17 @@ OpenMCCellAverageProblem::initialSetup()
...
-                 "Unexpected behavior may occur.");
-
+  if (_use_displaced && !_using_skinner && !movesOpenMCGeometry())
+    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner', an "
+                 "OpenMCCellTransform user object, or a geometry changing criticality search "
+                 "(all of which move the OpenMC geometry). The [Mesh] will move, but the "
+                 "underlying OpenMC geometry will remain unchanged. Unexpected behavior may "
+                 "occur.");
+  else if (!_use_displaced && movesOpenMCGeometry())
+    mooseWarning("The OpenMC geometry is being moved (via an OpenMCCellTransform user object "
```
meltawila: please add a test that shows this warning. I think the most straightforward one should be in `test/tests/userobjects/openmc_cell_transform/tests` as a `type = RunException` test with `input = cells_shadow.i` and add this warning as `expect_err`.

Another one you can do would be in `test/tests/criticality/rotation/tests` which runs `input = coupling.i`

src/base/OpenMCCellAverageProblem.C
```diff
@@ -569,12 +569,17 @@ OpenMCCellAverageProblem::initialSetup()
 
   getOpenMCUserObjects();
 
-  if (_use_displaced && !_using_skinner && !hasCellTransform())
-    mooseWarning("Your problem has a moving mesh, but you have not provided a 'skinner' or an "
-                 "OpenMCCellTransform user object (both of which move the OpenMC geometry). The "
-                 "[Mesh] will move, but the underlying OpenMC geometry will remain unchanged. "
-                 "Unexpected behavior may occur.");
-
+  if (_use_displaced && !_using_skinner && !movesOpenMCGeometry())
```
meltawila: I think we should have both these checks inside `if (_needs_to_map_cells) 
{ 
...
}` to only show the warning when OpenMC geometry is mapped to the [Mesh].

src/base/OpenMCCellAverageProblem.C
meltawila: right, I'd replace:
```
 // Coupling re-initialization should be triggered if we have cell transforms which can happen
 // even if the mesh isn't moving or adaptive.
 _need_to_reinit_coupling |= hasCellTransform();
  // The criticality search may modify the geometry.
  if (_criticality_search)
    _need_to_reinit_coupling |= _criticality_search->changingGeometry();
```
with:
```
_need_to_reinit_coupling |= movesOpenMCGeometry();
```
and add a comment describing this

meltawila (changes requested):
Thank you @cvoorhees15! just final thoughts before merging. they should be quick additions. lmk if the tests I'm thinking of are not clear, I haven't checked these myself but I think both tests should work

test/tests/userobjects/openmc_cell_transform/tests
```diff
@@ -68,4 +68,11 @@
     expect_err = "UserObjects/translate_cells/vector_value: Provide exactly 3 values/postprocessors: 'dx dy dz' in mesh units for translationtransform or 'φ, θ, ψ' in degrees for rotation transform."
     capabilities = 'openmc'
   []
+  [mismatching_geometry_and_mesh]
+    type = RunException
+    input = cells_shadow.i
+    cli_args = '--error'
+    expect_err = "The OpenMC geometry is being moved \(via an OpenMCCellTransform user object or a geometry changing criticality search\), but your problem has a fixed mesh. The underlying OpenMC geometry will change, but the \[Mesh\] will not move. Unexpected behavior may occur."
+    capabilities = 'openmc'
+  []
```
meltawila, line 78:
```suggestion
  requirement = 'The system shall warn the user when the OpenMC geometry is moved by a cell transform while the mesh remains stationary.'
  []
```

meltawila (changes requested):
A test `requirement` parameter is necessary to meet SQA

test/tests/userobjects/openmc_cell_transform/tests
meltawila: please also add
```
  [rotation_mismatching_geometry]
    type = RunException
    input = coupling.i
    cli_args = '--error'
    expect_err = "The OpenMC geometry is being moved"
    requirement = 'The system shall warn the user when a criticality search rotates the OpenMC geometry while the mesh remains stationary.'
    capabilities = 'openmc'
  []
```
in `test/tests/criticality/rotation/tests`

cvoorhees15: Should be all set now. Not sure why there've been CI failures all along...

meltawila (approved):
LGTM!

Issue #1439, opened 2026-08-26
Update BitterRoot & WindRiver modules


gavin-standish:
## Bug Description
After the most recent INL HPC outage, current modules cause error in the building of Cardinal with NekRS enabled.

aprilnovak:
Can this be closed now?

PR #1440, opened 2026-08-26
Modified modules for BitterRoot and WindRiver

gavin-standish:
Closes Issue #1439

### DIFF
--- a/doc/content/hpc.md
+++ b/doc/content/hpc.md
@@ -387,7 +387,7 @@ fi
 if [[ "$SHORTHOST" ~= br[0-9]+ ]] || [[ "$SHORTHOST" =~ bitterroot* ]]
   echo "Loading Bitterroot modules..."
   module purge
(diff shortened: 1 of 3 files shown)


aprilnovak:
Did you test this with both OpenMC and NekRS enabled? The best way is to run the `pincell_multiphysics` tutorial. Can you also update the job script so that it is consistent? It should be in `scripts`

gavin-standish:
> Did you test this with both OpenMC and NekRS enabled? The best way is to run the `pincell_multiphysics` tutorial. Can you also update the job script so that it is consistent? It should be in `scripts`

Yes, both the OpenMC `lattice` and NekRS `sfr_pincell` install checks successfully ran without errors. I have also updated the job script.

aprilnovak:
LGTM! Thanks @gavin-standish

aprilnovak:
Failures are unrelated and are being addressed in #1435

PR #1450, opened 2026-09-01
Make build_nekrs/moab/embree/doubledown/dagmc/openmc order-only prereqs

jcphill:
`$(app_objects)` and `$(test_objects)` listed these six .PHONY guard targets as normal prerequisites. A .PHONY target is never considered up to date, so any target depending on one is unconditionally rebuilt -- meaning every app and test object file was recompiled (and everything relinked) on literally every invocation of make, whether or not anything had actually changed.

The comment directly above these lines states the actual intent: 'We need to build these first so we get include dirs' -- a pure ordering requirement, not a freshness/rebuild-on-change one. Order- only prerequisites (the '|' syntax) express exactly that: they still guarantee build_nekrs etc. run before any object compiles, without forcing perpetual staleness. Genuine rebuild-on-change tracking for an actual header/library edit is unaffected -- it's already handled separately, per file, via the compiler-generated .d dependency files app.mk/moose.mk already -include.

Verified with the CMake superbuild (github.com/jcphill/cardinal, cmake-superbuild branch): a full build proceeds identically (guard messages for the skipped MOAB/Embree/Double-Down/DagMC builds still print before any compilation begins, confirming ordering is preserved), and a second, immediately-following build with nothing changed now correctly does zero recompiling/relinking of Cardinal's own app and test code, versus recompiling and relinking everything every time beforehand.


aprilnovak:
I'm not sure why coverage is failing, can you force push to re-trigger?

jcphill:
It looks like make is being re-run with coverage=true in an already built tree, so nothing is being built:

> How coverage=true gets applied at all: it's a plain make command-line variable, not tracked by any file. MOOSE's own pattern rule for compiling app objects is:
> 
> ```
> # framework/build.mk:163
> %$(1).$(obj-suffix) : %.C $(ADDITIONAL_SRC_DEPS) | $$(prebuild)
> ```
> 
> The only real prerequisites are the .C file and its .d-tracked headers — $(coverage)/CXXFLAGS never appear in the dependency graph. This is universal, well-known Make behavior: flag values are invisible to staleness checks. So if coverage=true is added on a second make invocation against a tree whose objects are already up to date from an earlier invocation, Make sees "source unchanged" and does nothing — the objects silently stay uninstrumented (no -fprofile-arcs -ftest-coverage), and they'll never emit .gcda no matter how many tests you run against them.
> 
> What was masking this until now: before my fix, `$(app_objects)/$(test_objects)` listed the six .PHONY guard targets as normal (not order-only) prerequisites. A .PHONY target is never "up to date," so every app/test object was unconditionally rebuilt on every make invocation — which incidentally also re-applied whatever CXXFLAGS/coverage value was active on that particular invocation. That's exactly the "everything recompiles/relinks every time" bug I fixed. Converting those to order-only prereqs (correctly) stops that blanket rebuild — and in doing so exposes this pre-existing gap in MOOSE's coverage flag tracking.
> 
> This matches the failure signature exactly: unit/ still produced 5 .gcda files (that build step is presumably a separate, fresh invocation), while build/ and src/ — Cardinal's own app objects — produced zero, which is what you'd see if those objects were already compiled (without coverage flags) by an earlier step/stage and then just re-linked, rather than recompiled, when the coverage job added coverage=true.
> 
> I confirmed the rule structure directly (build.mk:163 has no flag-dependency) — that part is certain. What I can't confirm from here is whether CIVET's recipe actually does reuse/cache a build tree across steps for this job (I don't have the CIVET recipe repo checked out locally, and gh is blocked by the org's PAT policy). Could you check the fuller CI log for an earlier make (without coverage=true) against the same BUILD_ROOT/cardinal/build? If there is one, that confirms it, and the real fix belongs in the CI recipe (do a make clobberall — or start the coverage job from a clean tree — before building with coverage=true) rather than reverting my Makefile fix.

jcphill:
This will conflict with and need updating after #1456 merges but the real blocker is the CIVET coverage failure.

aprilnovak:
I'm not sure why coverage is failing; I've pinged @loganharbour who can perhaps comment. Other PRs are passing fine, so I'm inclined not to merge as-is with the failure before understanding it

jcphill:
I'm pretty sure the reason coverage is failing is that the changes in this PR are working as intended and the CIVET test script sequence needs to be modified to enable coverage in the "Build openmc" step (or just remove it and let "Build" compile everything with coverage enabled).

If you compare the "Build" step of https://civet.inl.gov/job/4136738/ ("Test" from this PR) with the passing https://civet.inl.gov/job/4139127/ you can see that the passing PR does re-compiles that this PR avoids: 
```
% diff results_4136738_Test/07_Build results_4139127_Test/07_Build | grep Compiling
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/tallies_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/parser_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/mesh_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/transfers_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/meshgenerators_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/modelmodifiers_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/markers_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/transforms_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/ics_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/timesteppers_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/utils_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/criticality_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/controls_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/postprocessors_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/filters_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/auxkernels_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/base_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/actions_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/interfaces_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/indicators_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/build/unity_src/userobjects_Unity.C...
> Compiling C++ (in opt mode) /tmp/build/cardinal/test/src/base/CardinalTestApp.C...
```

aprilnovak:
I checked with @loganharbour and he agreed to provide access to the `civet_recipes` repo - did that come through already?  https://github.com/idaholab/civet_recipes

jcphill:
> This will conflict with and need updating after #1456 merges but the real blocker is the CIVET coverage failure.

The build/test system issues are resolved and this does not conflict with #1456's replacement #1466.

PR #1451, opened 2026-09-02
Update nek inputs syntax

GiudGiud:
refs idaholab/virtual_test_bed#877


aprilnovak:
There are more changes requires, the `.udf` and `.oudf` also need to be updated for v26. @stetsonschott do you want to help out on this one?

aprilnovak:
huh, also I am confused why these were out of date in the first place since we did the v26 update. Possibly forgot to update these tutorials

GiudGiud:
I think that's what happened, these inputs might not be under CI.
I'll add the .udf and .oudf changes if I figure it out

GiudGiud:
Added more fixes. There is a missing cardinal_sub.i input still

aprilnovak:
The `other_apps` tutorial is broken if Cardinal is the sub-app and dynamically linked to a super simple MOOSE App ([Chickadee](https://github.com/aprilnovak/Chickadee.git), which only exists for the sake of this dynamic linking tutorial). However, the same inputs run just fine if Cardinal is the main app and Chickadee the sub app. This tutorial was quite likely broken a long time in the past, since we can't test it effectively since it requires cloning another repo.

The error from NekRS is something probably related to clobbering the cache

```
Aborting in /home/ajnovak2/cardinal/contrib/nekRS/src/core/nekInterface/nekInterfaceAdapter.cpp:L676
set_usr_handles:
/home/ajnovak2/cardinal/tutorials/other_apps/.cache/nek5000/libbrick.so: undefined symbol: nekf_registerptr_
```

GiudGiud:
thanks for the additional fixes. I independently approve them.

aprilnovak:
And I approve your changes - thanks!

PR #1455, opened 2026-09-04
Fix building unit tests before building cardinal

jcphill:
Copy changes from Makefile to unit/Makefile.


aprilnovak:
Since we're modifying things here - do you have an idea of how to reduce this technical debt in the future? e.g. is there a way to include the relevant parts of the `../Makefile` without needing to copy the file down into `unit`?

jcphill:
I would recommend making the unit tests a target in the main Makefile and replacing unit/Makefile with:
```
.DEFAULT_GOAL := cardinal-unit

%:
	$(MAKE) -C $(CURDIR)/.. $@
```

aprilnovak:
Let's try that! Do you want me to merge this PR first or do you think it'd make sense to have that included on this one?

jcphill:
Please merge this PR first (red -> green -> refactor).

PR #1456, opened 2026-09-04
Replace unit/Makefile's hand-copied build with a forward to Makefile

jcphill:
unit/Makefile duplicated most of the top-level Makefile's own logic (ENABLE_*/module list/HDF5/PETSc/MOOSE-core/third-party dependency setup) by hand, with no way to keep the two in sync -- e.g. RAY_TRACING was missing from unit/Makefile's own module list, silently building the unit test binary without it. Recent history here is exactly this cost: several commits doing nothing but resyncing one Makefile's copy against the other (a8b0d4bd, f0ff2243, and others in this file's own log).

Adds a "cardinal-unit" target to the main Makefile instead: a second $(FRAMEWORK_DIR)/app.mk inclusion, which app.mk explicitly supports doing more than once in a single process for exactly this purpose (a second app linking against the first one's library via app.mk's own app_LIBS accumulation) -- no separate "rebuild libcardinal" block needed the way unit/Makefile used to carry, since it's the same process/rule graph that already knows how to build lib/libcardinal-<method>.la. Guarded with `ifneq ($(filter cardinal-unit%,$(MAKECMDGOALS)),)` so a plain `make`/`make cardinal-opt` never parses any of it -- necessary because ADDITIONAL_INCLUDES/ADDITIONAL_LIBS are read at recipe execution time (confirmed for ADDITIONAL_LIBS specifically, referenced directly in app.mk's own $(app_EXEC) link recipe; ADDITIONAL_INCLUDES turns out to already be a no-op after the include either way, folded into app_INCLUDES at include-processing time, harmless since MOOSE's own build.mk already adds contrib/gtest to every app unconditionally), so without the guard the GTEST library would silently link into cardinal-opt too whenever "cardinal" and "cardinal-unit" are both requested in one invocation. Both variables are scoped to cardinal-unit's own targets specifically (matching the EXTERNAL_FLAGS pattern already used above for "cardinal" itself) so that can't happen regardless.

unit/Makefile itself becomes a two-line forward to that target -- the only thing it has ever actually been used for (checking_install.md's own instructions are exactly `cd unit; make -j8`; nothing in this repo invokes it any other way). A generic `%: ...` catch-all forwarder was tried and rejected: it needs a FORCE prerequisite to avoid silently no-op'ing once unit/cardinal-unit-opt or unit/run_tests already exist on disk (confirmed live: `make cardinal-unit-opt` after a first build produced "is up to date" and never reached the real Makefile at all), and that FORCE prerequisite in turn makes the pattern match the literal name "Makefile" too, tripping GNU Make's own "remake the makefile" logic and rerunning everything twice -- solvable (an explicit `Makefile: ;` wins over the pattern match) but for no real benefit, since even "clean" wouldn't forward to anything that actually cleans unit/'s own artifacts either way (cardinal-unit's own targets in the main Makefile only exist once cardinal-unit is itself among the requested goals).

Verified directly (CPU-only, ENABLE_NEK/OPENMC/DAGMC off, against the container's prebuilt PETSc/libMesh/WASP): a fresh `make cardinal-unit` and `cd unit && make` both build/link unit/cardinal-unit-opt correctly (linking cardinal's own library automatically via app_LIBS, plus GTEST), all 9 unit tests pass via ./unit/run_tests, `make -n` for a plain build shows zero cardinal-unit/GTEST references, `make -n all cardinal-unit` confirms GTEST's library links into cardinal-unit-opt and nowhere near cardinal-opt's own link command, and a second invocation after the first successful build correctly rebuilds/relinks rather than silently no-op'ing.


aprilnovak:
Thanks @jcphill! Are any other changes coming to this PR, or is this ready for review/merge?

jcphill:
> Thanks @jcphill! Are any other changes coming to this PR, or is this ready for review/merge?

I think this is as good as it gets for minimum redundancy.

nuclearkevin (approved):
Thanks for this @jcphill! I'm a huge fan of the -411 line diff in our buildsystem :)

jcphill:
Supplanted by the smaller, simpler, more robust, and more canonical #1466.

aprilnovak:
should we therefore close this PR if it's been replaced by #1466?

PR #1457, opened 2026-09-04
Allow users to skip re-applying feedback in a criticality search 

### LINKED ISSUE #1454: Add option to send temperature/density once per multiphysics iteration when using CriticalitySearch
## Reason

There are certain cases in which it would be sufficient to only send temperature / density at the beginning on a multi physics iteration while searching for the critical root. Currently, the data is sent before every Monte Carlo root evaluation, which catches all cases which might need the data to be resent. It would be a slight optimization (and an approximation of varying quality) to just send it once per Picard.

## Design

Probably add a param to turn this on/off and a warning of sorts if we can try to detect if it would be problematic to do so. Maybe adding a method argument here with some error checking so that [reinitCouplingAndApplyFeedback](https://github.com/neams-th-coe/cardinal/blob/b28fd869dfa04de76e318e1ef614dd6dce17ed4b/src/base/OpenMCCellAverageProblem.C#L2750) and thus [sendTemperaturetoOpenMC](https://github.com/neams-th-coe/cardinal/blob/b28fd869dfa04de76e318e1ef614dd6dce17ed4b/src/base/OpenMCCellAverageProblem.C#L2457) and [sendDensityToOpenMC](https://github.com/neams-th-coe/cardinal/blob/b28fd869dfa04de76e318e1ef614dd6dce17ed4b/src/base/OpenMCCellAverageProblem.C#L2552) are only called once per multi physics iteration during a critical search.

https://github.com/neams-th-coe/cardinal/blob/b28fd869dfa04de76e318e1ef614dd6dce17ed4b/src/base/OpenMCCellAverageProblem.C#L2839-L2850
## Impact

Speedup for some `CriticalitySearch`s

nuclearkevin:
Closes #1454


src/criticality/CriticalitySearchBase.C
```diff
@@ -61,6 +61,13 @@ CriticalitySearchBase::validParams()
       "Whether non-eigenvalue tallies should be disabled during the search process as a "
       "performance optimization. If set to 'false', `run_critical_state` must be set to 'true' "
       "to ensure tallies are computed on the last iteration.");
+  params.addParam<bool>(
+      "apply_feedback_in_search",
+      true,
+      "Whether temperature/density feedback should be re-applied when performing a criticality "
+      "search iteration. Setting this to 'false' will skip sending feedback to OpenMC on every "
+      "criticality search iteration, which improves performance at the cost of ignoring the "
+      "effect of remapping feedback on the final critical state.");
```
aprilnovak:
```suggestion
      "effect of remapping feedback on the final critical state. This optimization means that your critical state is based on the most recent temperatures/densities fed into OpenMC, which themselves could change as the OpenMC model is changed during the criticality search.");
```

test/tests/criticality/rotation/gold/disable_feedback_out.csv
```diff
@@ -0,0 +1,2 @@
+time,point_cell_temp
+1,408.92183627627
```
aprilnovak, line 2: I'm not sure how to understand this number - is this an initial temperature from the `model.xml` which simply is not getting changed? 

Can we add another test in the same directory that does not use the new parameter option to show that it functions as expected?

aprilnovak:
"Feedback" here refers to only temperature and density - but I'd argue that geometric feedback is also a form of feedback. The documentation here is clear and unambiguous, but I'm wondering if we'd want to handle geometry as part of this PR as well? 

This might intersect with #1436 

@meltawila

src/criticality/CriticalitySearchBase.C
nuclearkevin: I think I would prefer: 
`This optimization means that your critical state is based on the temperatures/densities fed into OpenMC at the beginning of the Picard iteration, which themselves could change as the OpenMC model is changed during the criticality search.`

Most recent is a little ambiguous on when feedback is sent into OpenMC relative to the search.

aprilnovak: works for me!

test/tests/criticality/rotation/gold/disable_feedback_out.csv
nuclearkevin: This is the temperature sent to a cell at the beginning of the Picard iteration, which corresponds to the new parameter being set to false. If the new parameter is set to true (default state), the cell temperatures are set on every search iteration and so this number changes to 600 K instead (the correct mapping at the critical angle). I can add a test if you still think it is necessary, or add a note to the test documentation to describe the failure state instead.

nuclearkevin:
I don't think we should handle geometry in this PR as disabling geometric feedback safely has implications on the cell-element mapping (and therefore cell tally results reported at the end of the search). At present, this PR is self-contained and addresses the issue Lewis raised.

test/tests/criticality/rotation/disable_feedback.i
```diff
@@ -0,0 +1,70 @@
+!include openmc.i
```
aprilnovak, line 1: I'm confused why this line is here if you are redefining the mesh, problem block, etc? Does anything get included from `openmc.i`?

nuclearkevin: In his case I'm only overriding a small subset of input parameters in [openmc.i](https://github.com/neams-th-coe/cardinal/blob/devel/test/tests/criticality/rotation/openmc.i). Most of the mesh and criticality search blocks are used verbatim from the include.

aprilnovak: Ohh, I realize i was misunderstanding how the includes work - the `[Mesh]` block in this file doesn't override that in the `openmc.i`, it just gets added to it. Makes sense!

include/base/OpenMCProblemBase.h
```diff
@@ -428,8 +428,12 @@ class OpenMCProblemBase : public CardinalProblem, public PostprocessorInterface
   const openmc::Tally & getMGBetaTally();
 
 protected:
-  /// A virtual function to allow for execution prior to each step in a criticality search.
-  virtual void critSearchStep() = 0;
+  /**
+   * A virtual function to allow for execution prior to each step in a criticality search.
+   * @param[in] apply_feedback whether temperatures/densities should be sent to OpenMC
+   * on a criticality search iteration or not
```
lewisgross1296:
```suggestion
   * A virtual function to allow for execution prior to each guess in a criticality search.
   * @param[in] apply_feedback whether or not to send temperatures/densities
   *  to OpenMC before it runs transport to evaluate each guess.
```

Just confirming but "a criticality search iteration" refers to one single Brent guess and transport to evaluate the guess, right? I think the term `guess` might be slightly more clear than `step`, but this is a minor suggestion. Generally, I see Brent related materials refer to each new root "step" as the newest guess

include/criticality/CriticalitySearchBase.h
```diff
@@ -42,9 +42,10 @@ class CriticalitySearchBase : public MooseObject, public OpenMCBase
   /**
    * Use Brent's method to search for criticality
    * @param[in] step_callback a callback to execute after the OpenMC model has been updated prior to
-   * running the OpenMC solve.
+   * running the OpenMC solve. This function takes a boolean parameter to determine if feedback
+   * should be sent to OpenMC on a given step or not
```
lewisgross1296:
```suggestion
   * should be sent to OpenMC before each guess or not
```

src/criticality/CriticalitySearchBase.C
```diff
@@ -61,6 +61,16 @@ CriticalitySearchBase::validParams()
       "Whether non-eigenvalue tallies should be disabled during the search process as a "
       "performance optimization. If set to 'false', `run_critical_state` must be set to 'true' "
       "to ensure tallies are computed on the last iteration.");
+  params.addParam<bool>(
+      "apply_feedback_in_search",
+      true,
+      "Whether temperature/density feedback should be re-applied when performing a criticality "
+      "search iteration. Setting this to 'false' will skip sending feedback to OpenMC on every "
+      "criticality search iteration, which improves performance at the cost of ignoring the "
```
lewisgross1296:
```suggestion
      "search iteration. Setting this to 'false' will skip sending feedback to OpenMC before"
      "each criticality search guess, which improves performance at the cost of ignoring the "
```

test/tests/criticality/rotation/tests
```diff
@@ -54,6 +54,18 @@
...
+    type = CSVDiff
+    input = disable_feedback.i
+    min_parallel = 6
+    csvdiff = disable_feedback_out.csv
+    # for computational expediency, the test is run with much fewer particles
+    requirement = 'The system shall allow users to skip the re-applying feedback in criticality searches. '
+                  'This test checks to see if the cell temperature is the same as the cell temperature '
+                  'sent to OpenMC when a criticality search is not in use. Setting apply_feedback_in_search '
+                  'to true (or a failure) will result in a mapped temperature of 600 K instead of the '
+                  'reported value of 408 K.'
```
lewisgross1296, line 69: Maybe I'm not totally understanding the test, but for the desired behavior, shouldn't the temperature 600 be sent once and then not again, as opposed to not at all?

lewisgross1296 (commented):
Small changes on language, but I prefer referring to each "step" in a critical search as evaluating a new guess to match the language of BrentsMethod I've seen used elsewhere.

Small question about the test too. Thanks for putting this together so quickly! If you'd like, I can run this on the GCMR for one iteration to see how it compares to a situation where we re-apply feedback (current behavior).

test/tests/criticality/rotation/tests
nuclearkevin: The drum angle is not known until the search is complete, so the temperature of 600 K (which is computed by volume averaging over elements mapped to cells) can't be set unless the drum has its temperature set on the final (and therefore every) iteration.

nuclearkevin: It should be noted that this PR does not reapply temperatures/densities on the critical iteration calculation. This keeps the eigenvalue / tally results consistent with the search, otherwise your "critical state" calculation is unlikely to actually be critical due to changes in feedback mapping.

test/tests/criticality/rotation/gold/disable_feedback_out.csv
nuclearkevin: I've added a test for the converse. Not a massive fan of including this as it feels unnecessary, and over-testing criticality searches results in large increases to our testing runtimes (which the MOOSE folk won't be happy about).

aprilnovak: But where does the 408 come from? My question is that I can't see anything in the input files that makes this number seem rational as a result you are seeing here - there's two ICs, one that makes temperature 300 and the other 600. My question is, why do you think that 408 is a reasonable temperature to see here? (so, my question is not about the disabling feedback being correct, necessarily, but what this test is showing).

PR #1460, opened 2026-09-06
Allow the user to control when relaxation is active with the MOOSE control system

### LINKED ISSUE #1459: Allow the MOOSE control system to enable/disable tally relaxation
## Reason
It may be desirable to enable/disable relaxation when running mesh tally adaptive mesh refinement even when running custom adaptive tally relaxation schemes (#1446). This would allow users to converge the initial problem (relaxation enabled, AMR disabled), which can be followed up with AMR in Picard iterations (with relaxation disabled) to improve the spatial fidelity of tallies in regions of interest.

## Design
A new  *controllable* boolean parameter (`allow_relaxation`) should be added to the `OpenMCCellAverageProblem` which allows relaxation to be toggled on/off as desired.

## Impact
More flexibility when applying AMR and relaxation in Picard iterations.

nuclearkevin:
Closes #1459


src/tallies/TallyBase.C
```diff
@@ -630,6 +630,11 @@ TallyBase::relaxAndNormalizeTally()
       mooseError("Unhandled RelaxationEnum in TallyBase!");
   }
 
+  // Override the above if OpenMCCellAverageProblem is disabling relaxation
+  // (e.g. due to controls)
+  if (!is_relaxation_allowed)
```
aprilnovak, line 635: Could the Controls system just set the relaxation type to `none` to temporarily turn off relaxation so we don't have to introduce this other parameter? I'm not sure if Controls work with enumerations

src/base/OpenMCCellAverageProblem.C
```diff
@@ -159,6 +159,13 @@ OpenMCCellAverageProblem::validParams()
       "Whether to check that your model does indeed have identical cell fills, allowing "
       "you to set 'identical_cell_fills' to speed up initialization");
 
+  params.addParam<bool>(
```
aprilnovak, line 162: How does this work if using Dufek-Gudowski or Robbins-Monro relaxation? For instance, in Robbins-Monro the alpha quantity is `1/(n+1)`, where `n` is the relaxation index. I think the correct thing to do, if we are turning this on/off, is to make sure that `n` is only incremented after relaxation has happened.

Similarly, for Dufek-Gudowski both the alpha and number of particles/batch are varied with relaxation index. 

Can we add 2 additional test cases for these use cases to be sure the right thing is being done?

test/tests/neutronics/relaxation/control_relax.i
```diff
@@ -0,0 +1,74 @@
...
+    type = BoolFunctionControl
+    function = 'turn_on_relaxation'
+    parameter = '*/*/allow_relaxation'
+    # Execute before we run OpenMC
+    execute_on = 'TIMESTEP_BEGIN'
+  []
+[]
+
+[Problem]
+  allow_relaxation = false
```
aprilnovak: does this need to be here, if the Controls system is changing it? I wonder if you could just omit it (I did not check locally). If it can be dropped, IMO that'd be easier to understand syntax

test/tests/neutronics/relaxation/tests
```diff
@@ -0,0 +1,17 @@
+[Tests]
+  issues = '#1459'
+  design = 'OpenMCCellAverageProblem.md'
+
+  [control_relaxation]
+    type = CSVDiff
+    input = control_relax.i
+    csvdiff = control_relax_out.csv
+    requirement = "The system shall allow the MOOSE control system to enable/disable relaxation. "
```
aprilnovak: I appreciate the consideration here for hardening against floating point - but for testing Robbins-Monro and Dufek-Gudowski as mentioned in another comment, it may not be enough to check that the tally is different. Could we perhaps modify this test so that the relaxed tally is correctly showing as `alpha * ... + (1-alpha) * ...` within statistics of some reference?

src/tallies/TallyBase.C
nuclearkevin: From my understanding the controls system does not work with MOOSE enums at present. The closest analog is the `WebServerControl`, which can handle/control strings. It could be extended to MOOSE enums, but that would require external coupling Python scripts (which are rather clunky, my preference would be to do this entirely within the MOOSE input file).

src/base/OpenMCCellAverageProblem.C
nuclearkevin: Good catch - I didn't think about the implications on the other relaxation schemes. I'll modify the iteration counter and add additional tests.

src/tallies/TallyBase.C
aprilnovak: agreed! What do you think of changing the parameter name from having "allowed" in the name to "relaxation_on" or something like that? "allowed" to me more-so conveys a compatibility thing, rather than an on/off flag.

src/base/OpenMCCellAverageProblem.C
aprilnovak: or possibly you want to restart those indices whenever relaxation is enabled. In the possible use case where it's on, then off, and then back on again? So, restarting those indices when relaxation is first turned on.

src/tallies/TallyBase.C
nuclearkevin: Works for me!

src/base/OpenMCCellAverageProblem.C
nuclearkevin: I've decided to add another series of counters: `_relaxed_fp_iterations` (for the number of fixed point iterations that apply relaxation) and `_relaxed_total_n_particles` (for the number of particles simulated when relaxation is applied). These are incremented only if `_is_relaxation_enabled_by_controls` is true. Dufek-Gudowski and Robbins-Monro relaxation use these parameters, which ensures that relaxation will start from where it left off in the sequence (in terms of the relaxation fraction and the particles per batch). 

Two new tests have been added to check to make sure enable/disable behavior functions as expected for Dufek-Gudowski and Robbins-Monro relaxation. These gold value of alpha (back computed from the tally fields, compared to hand-calculated values) and the number of particles run in a given iteration.

test/tests/neutronics/relaxation/tests
nuclearkevin: I elected to keep this test the same (as constant relaxation is quite simple). I've added separate tests for Robbins-Monro and Dufek-Gudowski which gold alpha and the number of particle histories simulated in a given fixed point iteration.

include/base/OpenMCCellAverageProblem.h
```diff
@@ -440,8 +440,24 @@ class OpenMCCellAverageProblem : public OpenMCProblemBase
...
+   * Get the number of fixed point iterations where relaxation has been applied.
+   * @return the number of relaxed fixed point iterations
+   */
+  int relaxedFixedPointIteration() const { return _relaxed_fp_iterations; }
+
+  /**
+   * Total number of particles run when relaxation is active (not multiplied by batches)
+   * @return total number of particles simulated during relaxation
+   */
+  int nRelaxedTotalParticles() const { return _relaxed_total_n_particles; }
```
aprilnovak, line 459: Do we need to add `_relaxed_total_n_particles`, or can we just use `_total_n_particles` (perhaps with some changes)? `_total_n_particles` was only used for the particle count update in Dufek-Gudowski, so I feel like these two separate parameters could be made one and the same.

nuclearkevin: My reasoning for having a second parameter is that `_total_n_particles` can also be fetched with a post-processor, where it's referred to as the total number of particle histories simulated. If we're OK with `OpenMCParticles`'s `total` parameter representing the number of particles simualted during relaxation (and adjusting the docs accordingly), I can remove the added variable.

nuclearkevin: (It's also a little more involved to modify `_total_n_particles` as this variable is owned by `OpenMCProblemBase`, so I'd need to move all of the accumulation logic over to `OpenMCCellAverageProblem`. This felt incorrect at the time from a design standpoint as any derived class may want to know the number of particles simulated over all iterations)

test/tests/neutronics/relaxation/control_const_relax.i
```diff
@@ -0,0 +1,73 @@
...
+
+  xml_directory = './cell_tallies'
+
+  [Tallies/Cell]
+    output = 'unrelaxed_tally'
+  []
+[]
+
+[Outputs]
+  exodus := false
```
aprilnovak: out of curiosity, what is the `:=` syntax?

test/tests/neutronics/relaxation/control_dufek_gudowski.i
```diff
@@ -0,0 +1,105 @@
...
+    expression = '(p2 - p2_prev) / (p2_raw - p2_prev)'
+    pp_names = 'p2 p2_prev p2_raw'
+  []
+  [p3_alpha]
+    type = ParsedPostprocessor
+    expression = '(p3 - p3_prev) / (p3_raw - p3_prev)'
+    pp_names = 'p3 p3_prev p3_raw'
+  []
+
+  [particles]
```
aprilnovak, line 97: Can we also add a `total` particles to the CSV file, just to be sure that option works properly?

test/tests/neutronics/relaxation/gold/control_dufek_gudowski_out.csv
```diff
@@ -0,0 +1,7 @@
+time,p1_alpha,p2_alpha,p3_alpha,particles
+1,1,1,1,7000
+2,1,1,1,7000
+3,0.61802905797959,0.6180290579796,0.61802905797957,11326
+4,1,1,1,15351
```
aprilnovak: wouldn't we expect the number of particles used in run 4 to be whatever was used in run 3? It doesn't seem like the behavior is the same for runs 1-2 (using the initial number of particles) vs. in run 4 (which jumps to use the number of particles in a future run, run 6).

test/tests/neutronics/relaxation/gold/control_const_relax_out.csv
```diff
@@ -0,0 +1,4 @@
+time,p1_equal,p2_equal,p3_equal
+1,1,1,1
+2,1,1,1
+3,0,0,0
```
aprilnovak: Are we positive we're not missing any test coverage by not making this test the same "style" as the Robbins-Monro and Dufek-Gudowski tests? Since you already set up that logic, it'd be easy to just copy over for this test.

include/base/OpenMCCellAverageProblem.h
aprilnovak: no worries, it's fine as-is! Thanks for the explanation

test/tests/neutronics/relaxation/control_const_relax.i
nuclearkevin: `:=` is known as a Walrus operator. In MOOSE syntax, it's equivalent to an assignment if the parameter/variable isn't set. If it has been set before (which is the case here as `exodus = true` in the included file), the value is overridden by the top-level include.

aprilnovak: yes, am familiar with it from Makefiles - but didn't know MOOSE supported it as well! neat

test/tests/neutronics/relaxation/gold/control_dufek_gudowski_out.csv
nuclearkevin: Good catch! I wasn't guarding the Dufek-Gudowski particle update with `_is_relaxation_enabled_by_controls`

test/tests/neutronics/relaxation/gold/control_const_relax_out.csv
nuclearkevin: I'm certain that we aren't, but for the sake of maintaining testing equivalent between the two schemes I've changed it to the same format as Robbins-Monro and Dufek-Gudowski.

aprilnovak:
Thanks @nuclearkevin! LGTM!

PR #1464, opened 2026-09-12
Adds mooseWarning for when properties are specified multiple ways

### LINKED ISSUE #1307: Warn user if properties specified multiple ways / report precedence of which will be accepted
With the addition of openmc-dev/openmc#3808, users can now specify `settings.properties_file = /path/to/properties.h5` in the XML in addition to general cell temperatures/densities that were previously only recorded in the `<geometry>` part of a model's XML (technically `<materials>` too if not using distributed densities). 

This will only affect the IC for transport, but a poor IC could lead to more iterations required for convergence. A well-posed problem should eventually converge to the correct solution, but I think it's important to alert the user that they might be defining the same thing differently in two separate places. Usually, to specify properties from XML a user would have `initial_properties = XML`. If the user is specifying cell properties as well as a properties file, I think we should default to using the properties file, but I'm open to being convinced of the converse with good reason.

If the user specifies `initial_properties = XML` and we detect that `settings.properties_file` is not `None`, the warning could be something
```
You have specified `initial_properites = XML` in addition to loading 
properties from `settings.properties_file` (->replace with actual path). 

The `settings.properties_file` will be used instead of properties set in the
 <geometry> section of the XML.
```

cvoorhees15:
Closes #1307


meltawila (approved):
Thank you for the PR. I'll wait before merging in case @lewisgross1296 has any comments

src/base/OpenMCCellAverageProblem.C
```diff
@@ -269,6 +269,12 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
   if (!_specified_temperature_feedback && !_specified_density_feedback)
     checkUnusedParam(
         params, "initial_properties", "'temperature_blocks' and 'density_blocks' are unused");
+  else if (_initial_condition == coupling::xml && !openmc::settings::properties_file.empty())
```
meltawila, line 272: We should also consider the case `_initial_condition == coupling::moose` where a properties file is specified in openmc settings.xml file

meltawila (changes requested):
Thinking more about this, I think everywhere else in Cardinal input file parameters would override what was specified in an OpenMC input, so I'd prefer if we keep that here and use `initial_properties=xml` instead of respecting a `openmc::settings::properties_file` parameter.

Also let's make sure we cover all cases that would specify properties with multiple ways. `initial_properties=moose` is the default and would contradict with a properties file specified in settings.xml similar to your new test. I imagine this would be the more common case since it's used by default so let's also show the moose warning for this as well

src/base/OpenMCCellAverageProblem.C
lewisgross1296: In this case, I agree that the `coupling::moose` should win because if `openmc.settings.properties_file` exists, it _should_ be able to be specified using `initial_properties = hdf5` and having the local `properties.h5`. However, I've experienced cases where using `initial_properties = hdf5` failed with a `properties.h5` file generated by Cardinal in my multiphysics + depletion cases. I think this is related to Cardinal using `import Properties()` as opposed to the C-API way that got merged into OpenMC, though.

Regardless, the MOOSE IC should win, because you can always remove it from your main app and use  `settings.properties_file`. Can likely move other ICs to subapps if really needed.

lewisgross1296: We might want to have a test where `initial_properties = xml` is specified and we confirm that a property in `settings.properties_file` is respected

test/tests/openmc_errors/input_params/tests
```diff
@@ -37,4 +37,13 @@
     requirement = "The system shall error if both or none of the cell level options have been prescribed."
     capabilities = 'dagmc'
   []
+  [xml_and_properties_file]
+    type = RunException
+    input = xml_and_properties.i
+    cli_args = '--error'
```
lewisgross1296:
```suggestion
```
Pretty sure that the MOOSE testing framework treats warnings as errors, so you might not need this

test/tests/openmc_errors/input_params/tests
```diff
@@ -37,4 +37,13 @@
...
+    expect_err = "You have specified 'initial_properties = xml' in addition to loading "
```
lewisgross1296: The `expect_err` leverages regular expressions so you can likely get away with a shorter message that matches the most important pattern

test/tests/openmc_errors/input_params/tests
```diff
@@ -37,4 +37,13 @@
...
+                 "properties from settings.properties_file \(xml_and_properties/properties.h5\)"
+    requirement = "The system shall warn if cell properties and a properties file are both specified."
+    capabilities = 'openmc'
+  []
```
lewisgross1296, line 48:
```suggestion
    issues = `#1307`
  []
```
We now add an issue ref for posterity

test/tests/openmc_errors/input_params/xml_and_properties.i
```diff
@@ -0,0 +1,21 @@
...
+    file = ../../neutronics/meshes/sphere.e
+  []
+[]
+
+[Problem]
+  type = OpenMCCellAverageProblem
+  temperature_blocks = '1'
+  cell_level = 0
+  initial_properties = xml
+  xml_directory = xml_and_properties
```
lewisgross1296:
```suggestion
  xml_directory = 'xml_and_properties'
```
For variables relating to paths, I prefer to wrap them in `''`

lewisgross1296 (changes requested):
Few small things, you can likely lump them into one commit by using "Add to batch".

Also, is `test/tests/openmc_errors/input_params/xml_and_properties/properties.h5` a new file? If not, maybe a sym link to wherever the copy exists so that there aren't duplicate files in the repo

src/base/OpenMCCellAverageProblem.C
cvoorhees15: So update logic to
```
else if ((_initial_condition == coupling::xml || _initial_condition == coupling::moose) && !openmc::settings::properties_file.empty())
```

Or am I misunderstanding?

test/tests/openmc_errors/input_params/tests
cvoorhees15: The test failed when this was removed, so I amended the batch of suggestions to add it back.

cvoorhees15:
> Few small things, you can likely lump them into one commit by using "Add to batch".
> 
> Also, is `test/tests/openmc_errors/input_params/xml_and_properties/properties.h5` a new file? If not, maybe a sym link to wherever the copy exists so that there aren't duplicate files in the repo

I generated this with a dumby run where `export_properties = true` was set so I believe it is technically new.

src/base/OpenMCCellAverageProblem.C
cvoorhees15: So add `_initial_condition == coupling::moose` to the else if logic and add the described test? Or is there something else I'm missing?

meltawila: yes correct, and correcting the warning to say that XML/moose properties are used instead of the properties file. `switch (_initial_condition)` in `OpenMCCellAverageProblem::reinitCouplingAndApplyFeedback()` is where these are handled and I don't think you need to do anything there. I think the current mooseWarning is not accurate (`settings.properties_file` is not used and the initial condition set in the switch case following `initial_properties` input parameter in Cardinal is what overrides that, which is what I think we agreed on in the comments above)

cvoorhees15: Let me know what you think of the latest commit, I think that should cover it. Also had some CI jobs failing and I'm not sure why...

lewisgross1296:
VTB  failure might be related to some kind of subchannel syntax change since they seem to have this error
```
 A 'SCMFrictionUpgradedChengTodreas' is not a registered object.
 ```
you can see the failed tests
```
Failed Tests:
--------------------------------------------------------------------------------------------------------------
[0.227s] [  17MB]     FAIL sfr/subchannel/ornl_19_pin.channel_syntax FAILED (EXIT CODE 1 != 0)
[0.242s] [  15MB]     FAIL sfr/subchannel/multiple_SCM_assemblies/7assemblies.SCM_syntax FAILED (EXIT CODE 1 != 0)
[0.227s] [  16MB]     FAIL sfr/subchannel/toshiba_37_pin.subchannel_syntax FAILED (EXIT CODE 1 != 0)
[0.256s] [  11MB]     FAIL sfr/subchannel/THORS.channel_syntax FAILED (EXIT CODE 1 != 0)
[0.813s] [ 149MB]     FAIL sfr/subchannel/multiple_SCM_assemblies/7assemblies.Master_app_syntax FAILED (EXIT CODE 1 != 0) [slots=5]
[2.123s] [1150MB]    ERROR sfr/subchannel/multiple_SCM_assemblies/7assemblies.SCM7 FAILED (NULL output in runner_run) [min_cpus=7]
```

Not sure why `Test` failed though, it looks like a compiler error but the other tests compiled so maybe some kind of intermittent failure
```
make[2]: Leaving directory '/tmp/build/cardinal/build/openmc'
Install the project...
/usr/bin/cmake -P cmake_install.cmake
-- Install configuration: "Release"
-- Installing: /tmp/build/cardinal/unit/../install/lib/cmake/pugixml/pugixml-targets.cmake
-- Installing: /tmp/build/cardinal/unit/../install/lib/cmake/pugixml/pugixml-targets-release.cmake
-- Installing: /tmp/build/cardinal/unit/../install/lib/pkgconfig/pugixml.pc
-- Installing: /tmp/build/cardinal/install/bin/openmc
-- Set non-toolchain portion of runtime path of "/tmp/build/cardinal/install/bin/openmc" to "/tmp/build/cardinal/unit/../install/lib:/opt/petsc/lib:/tmp/build/cardinal/install/lib:/opt/libmesh/lib:/opt/vtk/lib"
-- Installing: /tmp/build/cardinal/unit/../install/lib/libopenmc.so
-- Set non-toolchain portion of runtime path of "/tmp/build/cardinal/unit/../install/lib/libopenmc.so" to "/tmp/build/cardinal/unit/../install/lib:/opt/petsc/lib:/tmp/build/cardinal/install/lib:/opt/libmesh/lib:/opt/vtk/lib"
-- Installing: /tmp/build/cardinal/unit/../install/lib/cmake/OpenMC/OpenMCTargets.cmake
-- Installing: /tmp/build/cardinal/unit/../install/lib/cmake/OpenMC/OpenMCTargets-release.cmake
make[1]: Leaving directory '/tmp/build/cardinal/build/openmc'
Compiling C++ (in opt mode) /tmp/build/cardinal/unit/src/BinUtilityTest.C...
Compiling C++ (in opt mode) /tmp/build/cardinal/unit/src/CardinalUnitApp.C...
Compiling C++ (in opt mode) /tmp/build/cardinal/unit/src/NekUtilityTest.C...
Compiling C++ (in opt mode) /tmp/build/cardinal/unit/src/SymmetryPointGeneratorTest.C...
Linking Library /tmp/build/cardinal/unit/lib/libcardinal-unit-opt.la...
Linking Executable /tmp/build/cardinal/unit/cardinal-unit-opt...
/usr/bin/ld: /tmp/build/cardinal/unit/cardinal-unit-opt: hidden symbol `__gcov_exit' in /usr/lib/gcc/x86_64-linux-gnu/14/libgcov.a(_gcov.o) is referenced by DSO
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make: *** [/tmp/build/cardinal/contrib/moose/framework/app.mk:585: /tmp/build/cardinal/unit/cardinal-unit-opt] Error 1
[21:05:40] ERROR: Exiting with code 2

[21:05:40] Execution in moosebuild completed with return code 2
[21:05:40] Submitting step statistics
[21:05:40] Failed to submit step statistics
```

**Comment by lewisgross1296** [comment]
Maybe try another force push, sometimes CIVET fails when it shouldn't

**Comment by nuclearkevin** [comment]
The recent MOOSE submodule update merged in #1465 fixes both of these. If tests are triggered again, they will pass.

**Comment by lewisgross1296** [comment]
Both MOOSE and VTB seem to have updated to the `SCMFrictionUpgradedChengTodreas` syntax, not sure why it's failing still. Maybe need to rebase?

**Comment by aprilnovak** [comment]
The test failures are unrelated, they were fixed in #1465; they would go away with a rebase

File: src/base/OpenMCCellAverageProblem.C (hunk 1 above)
Comment (meltawila) [reply src/base/OpenMCCellAverageProblem.C]: Looks good thank you. if you'd rebase your branch on top of the latest upstream the failing tests should pass

File: test/tests/openmc_errors/input_params/tests [hunk 6]
```diff
@@ -37,4 +37,30 @@
...
+    expect_err = "You have specified 'initial_properties = moose' in addition to loading properties from settings.properties_file"
+    requirement = "The system shall warn if initial_properties = moose is used while an OpenMC properties file is also specified."
+    capabilities = 'openmc'
+    issues = '#1307'
+  []
+  [xml_properties_respected]
+    type = CSVDiff
+    input = xml_properties_respected.i
+    csvdiff = xml_properties_respected_out.csv
+    requirement = "The system shall apply temperatures from settings.properties_file when initial_properties = xml."
```
meltawila, line 62: is this an old comment? I think with your latest change temperatures will be applied from xml not hdf5. please upadte

test/tests/openmc_errors/input_params/xml_properties_respected.i
```diff
@@ -0,0 +1,50 @@
+# Confirm that settings.properties_file temperatures are used when
```
meltawila, line 1: same comment. settings.properties_file won't be used and XML properties are used instead. please update this comment

meltawila (changes requested):
Thank you. I think some old comments still exist that don't reflect the latest change. please make sure these are consistent with the warning message

PR #1465, opened 2026-09-15
Compare FEType family/order directly for CONSTANT MONOMIAL checks

lindsayad:
- Several CONSTANT MONOMIAL/MONOMIAL_VEC checks compared a variable's `FEType` for equality against a literal `FEType(CONSTANT, MONOMIAL)`.
- This breaks if a future MOOSE version's default for `FEType`'s `p_refinement` flag differs from the literal's default (see idaholab/moose#33248), since the flag is included in `FEType::operator==`.
- Compare `family`/`order` directly instead, which is unaffected by `p_refinement` either way.


lindsayad:
when it rains it pours

nuclearkevin (approved):
LGTM! Thanks @lindsayad!

PR #1467, opened 2026-09-17
Adds support for structured meshes

makeclean:
In order to better facilitate linkage with random ray, it would be convenient to be able to more easily interface with structured mesh tallies on the Cardinal side. I need to add some tests.


nuclearkevin:
Just an FYI @makeclean, once https://github.com/openmc-dev/openmc/pull/3889 is merged we'll be able to support native unstructured mesh tallies in the random ray solver with XDG. I have a branch implementing this here (which also uses cell-under-voxel decomposition to ensure source regions align with unstructured mesh elements): https://github.com/nuclearkevin/cardinal/tree/xdg_mesh_tally.

PR #1468, opened 2026-09-19
Enable support for relaxation when applying AMR to mesh tallies

### LINKED ISSUE #1446: Support relaxation when running with AMR
## Reason
Multiphysics problems with strong coupling between neutronics and thermal-fluids (e.g. LWRs) often require the use of relaxation schemes to ensure convergence. Even when a problem only exhibits weak coupling, relaxation schemes often speed up convergence. At present, Cardinal does not support relaxation when running with tally mesh refinement as tally bins may change between Picard iterations. This limitation should be removed with custom projection operators to support relaxation in multiphysics problems that use mesh tally AMR.

## Design
WIP.

## Impact
Support for relaxation when running multiphysics problems with mesh tally AMR.

nuclearkevin:
Closes #1446


include/tallies/MeshTally.h
```diff
@@ -82,6 +87,100 @@ class MeshTally : public TallyBase
    */
   void checkMeshTemplateAndTranslations();
 
+  /// An enum for the different cases when applying relaxation to an adaptive mesh tally.
+  enum class AMRRelaxation
+  {
+    CaseI = 0,
```
aprilnovak: maybe different names to help with readability? e.g. `ParentCoarsened`, `ParentRefined`, `DoNothing`?

test/tests/neutronics/relaxation/cell_tallies/tests
```diff
@@ -142,11 +142,4 @@
     requirement = "The wrapping shall allow output of multiple tally scores, with relaxation applied independently to each score"
     capabilities = 'openmc'
   []
-  [no_relax_with_fixed_mesh]
-    type = RunException
-    input = openmc_adapt.i
-    expect_err = "When adaptivity is requested or a displaced problem is used, the mapping from the OpenMC model to the \[Mesh\] may vary in time."
```
aprilnovak, line 148: Do we need to retain a test somewhere regarding the displace problem aspect of this error message?

aprilnovak: nvm, saw it above!

src/tallies/MeshTally.C
```diff
@@ -304,4 +319,211 @@ MeshTally::checkMeshTemplateAndTranslations()
...
+  }
+
+  const auto alpha = getRelaxationFactor();
+  for (unsigned int score = 0; score < _tally_score.size(); ++score)
+  {
+    // Extract raw results.
+    extractAndNormalizeRaw(score);
+
+    // Shortcut if relaxation isn't being applied.
+    if (_openmc_problem.fixedPointIteration() == 0 || alpha == 1.0)
```
aprilnovak: Do we actually get inside these parentheses? Or is there a check elsewhere around the call to `relaxAndNormalizeTally` that would make it such that these don't get called

src/tallies/MeshTally.C
```diff
@@ -304,4 +319,211 @@ MeshTally::checkMeshTemplateAndTranslations()
...
+
+  // Check for Case III.
+  std::vector<const Elem *> descendants;
+  current_element->total_family_tree(descendants, true);
+  for (const auto descendant : descendants)
+    if (previousSpatialBin(descendant) != INVALID_SPATIAL_BIN)
+      return AMRRelaxation::CaseIII;
+
+  mooseError("Internal error: MeshTally::classifyRelaxationCase failed to classify an element.");
+  return AMRRelaxation::CaseI;
```
aprilnovak: Do you need this return statement? I think it never gets to this line?

nuclearkevin: The compiler throws a warning if a return isn't provided (if this were removed and Cardinal were compiled with warnings as errors, we'd fail to build).

src/tallies/MeshTally.C
nuclearkevin: At the moment we hit that statement on the first iteration or if no relaxation is applied (which we flag with `alpha = 1.0`). I can short-cut this by changing:

```c++
if (!_is_adaptive)
{
  TallyBase::relaxAndNormalizeTally();
  return;
}
```
to
```c++
if (!_is_adaptive || _openmc_problem.fixedPointIteration() == 0 || alpha == 1.0)
{
  TallyBase::relaxAndNormalizeTally();
  return;
}
```

and removing the parentheses, as they also exist in `TallyBase::relaxAndNormalizeTally()`.

include/tallies/MeshTally.h
nuclearkevin: Went with `Unchanged`, `CoarseToFine`, and `FineToCoarse`.

test/tests/neutronics/relaxation/amr_mesh_tallies/tests
```diff
@@ -0,0 +1,31 @@
+[Tests]
+  issues = '#1446'
+  design = 'MeshTally.md'
+
+  [same_level]
+    type = Exodiff
+    input = openmc.i
+    exodiff = 'openmc_out.e'
+    requirement = "The system shall revert to standard relaxation when the mesh is unchanged."
```
aprilnovak: Could this test be accomplished without creating a new gold file, `openmc_out.e`? I wonder if you could use one of the existing relaxation tests in another directory, just add a new `test.i` that results in no-mesh-changes-with-adaptivity, and then just point to an existing gold file?

aprilnovak: Not a huge deal if you prefer to keep it here

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
```diff
@@ -0,0 +1,76 @@
+[Mesh]
+  [pebble]
```
aprilnovak: Could these three test cases be made in 2D for a simple rectangle? I'm having a hard time looking at these outputs in paraview and doing the manual confirmation that the values are as-expected, and it got me thinking that this sphere is probably not necessary for testing this?

nuclearkevin: We cannot do 2D meshes (restriction in OpenMC). I can do cube, if that's preferred.

aprilnovak: or I guess 2D is not supported for mesh tallies, but perhaps 3D mesh with just one element in the cross-plane direction? I think simpler can be done here

test/tests/neutronics/relaxation/amr_mesh_tallies/tests
nuclearkevin: The other relaxation tests in the repository use a more complex mesh (three spheres with a stoplight background), so I can't gold against them unless you're OK with increasing the complexity of the test.

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
nuclearkevin: I can do that. I think the 1->8 and 8->1 cases are definitely easier to inspect and check in Paraview (also doesn't cost us as much to gold in terms of repo size).

test/tests/neutronics/relaxation/amr_mesh_tallies/tests
aprilnovak: I'm not sure I follow - I was interpreting this test as basically a standard-procedure for relaxation, i.e. the mesh does not change between picard iterations so nothing differs from how we already do relaxation. So, instead of making a new gold file here, could you just add an adaptivity block (which based on its settings does nothing) to one of the existing relaxation tests?

(if I am not making sense, please tell me - I am still quite jet lagged!)

nuclearkevin: Ooooooh, that's what you meant. Yeah, I can definitely do that!

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
nuclearkevin: Also, something else to note. If the relaxation scheme fails (maps AMR bins between iterations incorrectly) the norm of the relaxed vector no longer integrates power to the value specified in the OpenMCCellAverageProblem. This is really easy to detect with with an integral post-processor, which is included in this test and gets hit in the Exodiff.

nuclearkevin: Ended up doing a 2 -> 16 / 16 -> 2 case instead, since that introduces tilts in the mesh tally which propagate to the next iteration when applying relaxation. Since the integral is preserved under this scheme, 1->8/8->1 don't show any differences between the unrelaxed and relaxed tally results.

test/tests/neutronics/relaxation/amr_mesh_tallies/tests
```diff
@@ -0,0 +1,32 @@
+[Tests]
+  issues = '#1446'
+  design = 'MeshTally.md'
+
+  [same_level]
+    type = Exodiff
+    input = same_level.i
+    exodiff = 'same_level_out.e'
+    requirement = "The system shall revert to standard relaxation when the mesh is unchanged. "
+                  "This test golds against an existing mesh tally relaxation test."
```
aprilnovak, line 10: I'm not sure I follow - I see you added a new gold file. What does this mean then?

nuclearkevin: OK, that's weird. I could've sworn I symbolic linked against an existing gold file

aprilnovak: it's showing up as a new 154 kB file - I think you could avoid the sym link and just use the relative path? (not 100% sure if that works on civet)

nuclearkevin: I don't know if the test harness works particularly well with relative paths for gold files. I think the sym link will work (iirc we have other sym linked tests in the repo).

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
```diff
@@ -0,0 +1,84 @@
...
+  power = 1500.0
+  temperature_blocks = '0'
+  output_cell_mapping = 'false'
+  xml_directory = '../mesh_tallies'
+
+  cell_level = 1
+  scaling = 100.0
+
+  relaxation = 'constant'
+  relaxation_factor = 0.5
```
aprilnovak, line 58: would there be anything to gain by using a factor which is not 0.5? just to confirm that the "before" and "after" are properly treated? (since 0.5 would not catch that)

aprilnovak: I changed alpha locally to 0.3 and re-ran the test cases, and confirmed the values of the relaxed tallies by hand and can confirm it works properly at that value as well. If you see any possible reason why 0.5 might let a bug slip through, then please change to 0.3 or some other value.

include/tallies/MeshTally.h
```diff
@@ -82,6 +87,100 @@ class MeshTally : public TallyBase
...
+  /**
+   * Get the tally bin associated with a current element.
+   * @param[in] current_elem an element that is active on the current iteration
+   * @param[in] ext_filter the index associated with an external filter
+   * @return the tally bin associated with current_elem. Returns INVALID_SPATIAL_BIN
+   * if the element is not active
+   */
+  int64_t currentTallyBin(const Elem * current_elem, unsigned int ext_filter) const;
+
   /**
```
aprilnovak, line 184: Stray thought: do we support AMR when using the `mesh_template` idea? Do our built-in error checks robustly check if different translated meshes in the [Mesh] refine the same (and therefore catch the most likely scenario where the different translated meshes refine differently), or do we just preclude it altogether?

include/tallies/MeshTally.h
```diff
@@ -111,13 +210,21 @@ class MeshTally : public TallyBase
   /// Whether we're using an indirection layer to map between the OpenMC mesh tally and the MOOSE mesh.
   const bool _use_dof_map;
 
-  /**
-   * For use with block restriction only. A copy of the mesh is made which only contains elements in
-   * the blocks the user wishes to tally on. This is necessary at the moment as the point locators
-   * used in OpenMC to find collision sites are not passed a set of block IDs to filter elements.
-   * TODO: Fix this in OpenMC
-   */
-  std::unique_ptr<libMesh::ReplicatedMesh> _libmesh_mesh_copy;
```
aprilnovak, line 120: was this fixed in some earlier PR with an OpenMC update?

src/base/OpenMCCellAverageProblem.C
aprilnovak, line 290: I think this comment also needs to be updated to reflect the code changes below?

src/base/OpenMCCellAverageProblem.C
```diff
@@ -293,12 +298,11 @@ OpenMCCellAverageProblem::OpenMCCellAverageProblem(const InputParameters & param
...
-  if ((_use_displaced || _has_adaptivity) && _relaxation != relaxation::none)
+  if (_use_displaced && _relaxation != relaxation::none)
     paramError(
         "relaxation",
-        "When adaptivity is requested or a displaced problem is used, the mapping from the "
-        "OpenMC model to the [Mesh] may vary in time. This means that we have no guarantee that "
-        "the "
+        "When a displaced problem is used, the mapping from the OpenMC model to the [Mesh] may "
+        "vary in time. This means that we have no guarantee that the "
         "number of tally bins (or even the regions of space corresponding to each bin) are fixed. "
```
aprilnovak, line 306: should the "number of tally bins" part be removed from the error message, since that refers only to AMR? Same question about the "different length" part later in the message

src/base/OpenMCCellAverageProblem.C
```diff
@@ -2772,6 +2776,17 @@ OpenMCCellAverageProblem::syncSolutions(ExternalProblem::Direction direction)
...
 
+// Mesh contraction (deletion of subactive elements that were coarsened) is disabled
+// when running adaptive mesh tallies with relaxation. This is necessary
+// as the relaxation approach implemented for adaptive mesh tallies uses restriction
+// and projection operators, which need the more refined elements to stay around
+// to determine the mapping from previous tally bins to current tally bins.
+bool
+OpenMCCellAverageProblem::allowMeshContractionAfterMeshChanged() const
+{
+  return !(_has_mesh_tallies && _relaxation != relaxation::none && _has_adaptivity);
```
aprilnovak, line 2787: Just confirming - no surprises if using AMR _and_ the controls system is turning relaxation on/off? Do we need to add a test to cover this new possibility now that #1460 was merged?

src/tallies/MeshTally.C
```diff
@@ -65,8 +65,7 @@ MeshTally::MeshTally(const InputParameters & parameters)
     _estimator = nu_scatter ? openmc::TallyEstimator::ANALOG : openmc::TallyEstimator::COLLISION;
 
   // Error check the mesh template.
-  if (_openmc_problem.getMooseMesh().getMesh().allow_renumbering() &&
-      !_openmc_problem.getMooseMesh().getMesh().is_replicated())
```
aprilnovak, line 69: Can you jog my memory - was this just a bad piece of stray code in the first place, or does something about this PR make it so we don't need to check if replicated or not?

src/tallies/MeshTally.C
```diff
@@ -304,4 +319,205 @@ MeshTally::checkMeshTemplateAndTranslations()
               "!\n\nThe copy transfer requires that the [Mesh] and 'mesh_template' be identical.");
   }
 }
+
+void
+MeshTally::relaxAndNormalizeTally(bool is_relaxation_allowed)
+{
+  // Set alpha to unity if OpenMCCellAverageProblem is disabling relaxation
+  // (e.g. due to controls)
+  const auto alpha = is_relaxation_allowed ? getRelaxationFactor() : 1.0;
```
aprilnovak, line 328: Do you need to add this here, or did the changes in #1460 effectively add this thing elsewhere in the code? (so, is this redundant with a similar expression somewhere else?)

src/tallies/TallyBase.C
```diff
@@ -600,81 +600,56 @@ TallyBase::renormalizeLinkedTallies()
...
-  }
+  const Real norm = tallyNormalization(score);
+
+  if (_check_tally_sum && _needs_global_tally)
+    checkTallySum(score);
+
+  auto mean_tally = _openmc_problem.tallySum(_local_tally, score);
+  /**
+   * If the value over the whole domain is zero, then the values in the individual bins must be
+   * zero. We need to avoid divide-by-zeros.
```
aprilnovak, line 613: add test case for this? can have one of the e.g. tritium production scores with an appropriately-defined model

src/tallies/MeshTally.C
aprilnovak: Are you sure? I think there's a bunch of places in the code base where we don't return anything, I thought that if it's a switch statement at least, that you're permitted to not return anything if the default holds an error. For instance,

```
openmc::TallyEstimator
OpenMCProblemBase::tallyEstimator(tally::TallyEstimatorEnum estimator) const
{
  switch (estimator)
  {
    case tally::tracklength:
      return openmc::TallyEstimator::TRACKLENGTH;
    case tally::collision:
      return openmc::TallyEstimator::COLLISION;
    case tally::analog:
      return openmc::TallyEstimator::ANALOG;
    default:
      mooseError("Unhandled TallyEstimatorEnum!");
  }
}
```

So, maybe that does not apply to this code we're commenting on here, but there's some other switch-case statements where there is a dummy return value elsewhere in the code base which, if my claim is true, that we could remove (can be a separate PR).

test/tests/neutronics/relaxation/mesh_tallies/openmc.i
```diff
@@ -47,10 +47,6 @@
   [Tallies]
     [Mesh]
       type = MeshTally
-      mesh_template = ../../meshes/sphere_in_m.e
```
aprilnovak, line 50: What is the reason for these changes? Are we losing some coverage by doing this and instead having the mesh be read directly from the [Mesh]?

include/tallies/MeshTally.h
nuclearkevin: We do not support AMR with a `_mesh_template`.

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
nuclearkevin: I can't think of any way a change in alpha could let a bug slip through, as the value of alpha is now fetched from a common function. If something goes wrong with how alpha is determined based on the relaxation schemes, all relaxation tests will start failing.

include/tallies/MeshTally.h
nuclearkevin: This was fixed last year with a PR into OpenMC, followed by a PR to Cardinal, that supports block restriction in mesh tallies. This pointer just stuck around, and is never used.

src/tallies/MeshTally.C
nuclearkevin: This was a poorly defined error check that allowed for the use of mesh tallies without setting `allow_renumbering = false`.

nuclearkevin: All AMR mesh tallies (regardless of if the new relaxation scheme is used or not) fail when `allow_renumbering = true`.

src/tallies/MeshTally.C
nuclearkevin: It's also included in `TallyBase::relaxAndNormalizeTally`, but we need a copy in the overload as well.

src/tallies/TallyBase.C
nuclearkevin: This code block wasn't added in this PR (I just moved a common chunk of `relaxAndNormalizeTally` into it's own function for reuse). However, it looks like this case was never tested in Cardinal. I'll add a test

test/tests/neutronics/relaxation/mesh_tallies/openmc.i
nuclearkevin: The reason for this change is to support the use of the this test for checking the AMR relaxation null case. I couldn't override the mesh template and translations parameters, as they would become empty vectors and trigger error messages.

We aren't losing any test coverage here as the same code is used to extract the tally vectors from the mesh tally, regardless of whether it is a file mesh or the mesh mirror.

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
aprilnovak: not in the value of alpha itself, but what if you were to accidentally flip where alpha is used vs. (1-alpha) when assembling the relaxed value?

src/tallies/MeshTally.C
aprilnovak: do the cases run, but the test suite fails? or does the simulation itself bomb out? we want to protect the user from bad results, even if the test suite catches it it does not guarantee the code prevents the user from making a mistake

test/tests/neutronics/relaxation/amr_mesh_tallies/openmc.i
nuclearkevin: Ahhh, that's fair. I'll swap the test to 0.3 to catch that case.

src/tallies/MeshTally.C
nuclearkevin: The simulation would run but the results would be garbled. We tested the error previously, but the error would only get thrown on a distributed mesh. I'll add a test to make sure we hit this on any mesh tally in the future

src/tallies/MeshTally.C
nuclearkevin: IIRC, this works with less compliant C++ compilers because exceptions return an error code (an integer) which can be cast to an enum. IMO, not providing a return value is poor form in C++ (even with an error state) and I would recommend against us doing it.

src/tallies/MeshTally.C
nuclearkevin: I removed the `--distributed-mesh` cli-arg to cover this.

PR #1470, opened 2026-09-20
Update OpenMC

### LINKED ISSUE #1469: Update OpenMC
## Reason
OpenMC recently merged a pull request that [implements domain decomposition for the random ray solver](https://github.com/openmc-dev/openmc/pull/4026), allowing for the use of MPI. We should update our OpenMC submodule to bring in this change.

## Design
Feature-wise, this is an easy drop-in inclusion. Unfortunately, OpenMC also merged a PR before domain decomposition that removed [git submodules for dependency management](https://github.com/openmc-dev/openmc/pull/4085)... This breaks our build system and CI, which will need to be updated.

## Impact
Superior parallel scalability of the random ray solver on distributed memory systems (HPCs).

nuclearkevin:
This requires an update to the CIVET recipe to initialize the new pugixml and fmt git submodules we need to maintain. Additionally, a regold was required for one test due to changes in the volume estimator used by the random ray solver (see https://github.com/openmc-dev/openmc/pull/4110).

Closes #1469


jcphill (changes requested):
I would advise not adding fmt and pugixml as sub-modules because someone would then have to manually mirror any FetchContent_Declare changes in https://github.com/paulromano/openmc/blob/develop/cmake/Dependencies.cmake when updating OpenMC. Even now I don't think the hashes you have the submodules set to match the versions set in OpenMC.

nuclearkevin:
Would there be another way to go about getting these dependencies for OpenMC? From my understanding, we can't run CMake on CIVET while the build boxes are able to access the internet. We also can't install dependencies with system package managers on CIVET either as far as I know.

For normal builds we can rely on FetchContent being able to pull the archived packages, but that isn't the case in our CI AFAIK.

jcphill:
Could you make -DFETCHCONTENT_FULLY_DISCONNECTED=on etc. optional and enable it for CIVET builds, pointing to an existing download that isn't a git submodule?

nuclearkevin:
I believe its possible to do that on CIVET if we download the source archives in the recipe prior to running CMake. It doesn't solve the problem of needing to manually update the pinned archives every time OpenMC changes them, as we can't run CMake while we have internet access (unless there's another way to extract that dependency).

jcphill:
That sounds much better. The manual updates are at worst only needed at the PR stage and are enforced by CIVET rather than relying on a developer updating OpenMC to check if its dependencies have changed. You should be able to reliably extract the required downloads by searching openmc/cmake/Dependencies.cmake for `\s*URL\s+https://github.com/`.

